PR #2, opened 2020-09-18
add strategy to vote with FARM token from the sharing profit pool

chrouzz:
Hey,
This voting strategy is based on MAKER_DS_CHIEF voting strategy.
Idea is people participating to FARM ecosystem don't keep their FARM token in their wallet but put it in the sharing profit pool.
So this strategy would give a voting power to people who sent their token to the sharing profit pool contract.
Fixes # .

- 
- 
-

### DIFF
--- a/src/strategies/farm-no-mint-reward-pool/index.ts
+++ b/src/strategies/farm-no-mint-reward-pool/index.ts
@@ -1,0 +1,25 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import { abi } from './NoMintRewardPool.json';
+
+const FARM_NO_MINT_REWARD_POOL_ADDRESS = '0x59258F4e15A5fC74A7284055A8094F58108dbD4f';
+
+export async function strategy(provider, addresses, options, snapshot) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const response = await multicall(
+    provider,
+    abi,
+    addresses.map((address: any) => [
+      FARM_NO_MINT_REWARD_POOL_ADDRESS,
+      'deposits',
+      [address]
+    ]),
+    { blockTag }
+  );
+  return Object.fromEntries(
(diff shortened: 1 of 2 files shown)


src/strategies/farm-no-mint-reward-pool/index.ts
```diff
@@ -0,0 +1,25 @@
...
+const FARM_NO_MINT_REWARD_POOL_ADDRESS = '0x59258F4e15A5fC74A7284055A8094F58108dbD4f';
+
+export async function strategy(provider, addresses, options, snapshot) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const response = await multicall(
+    provider,
+    abi,
+    addresses.map((address: any) => [
+      FARM_NO_MINT_REWARD_POOL_ADDRESS,
+      'deposits',
```
bonustrack, line 14: Does this smart contract have a function "deposits"? It doesn't look like. You will need to setup the function you want call here, like "balanceOf". If its an ERC20 and you use balanceOf you don't need this strategy. Strategy will be needed if you want a custom call diferent than erc20 balance of.

chrouzz: yes, I understand.
In my case the contract has a balanceOf method which tell you the amount of FARM you staked in.
So I only need to add a basic erc20 strategy with the address of the contract right?

`  "strategies": [
    ["erc20BalanceOf", {
      "address": "0x59258F4e15A5fC74A7284055A8094F58108dbD4f",
      "symbol": "FARM",
      "decimals": 18
    }]
]`

where the address here is not FARM token address but the contract where the FARM token are staked.
Will it work like that?

bonustrack: Yes

chrouzz: great so I am going to do the PR on snapshot-spaces.

PR #4, opened 2020-09-23
Add contract-call strategy

in19farkt:
Added contract-call strategy that allows any contract method to be used to calculate voter scores.


bonustrack (approved):
Looking great! Thank you for your contribution!

PR #6, opened 2020-09-28
FARM uniswap LP voting strategy

chrouzz:
This strategy allows people who provide liquidity in uniswap to vote for the evolution of FARM.

Fixes # .

- 
- 
-


src/strategies/farm-unilp/index.ts
```diff
@@ -0,0 +1,33 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import abi from './NoMintRewardPool.json';
+import { Interface } from '@ethersproject/abi';
+import { Contract } from '@ethersproject/contracts';
+import uniLpAbi from './UniLPABI.json';
+
+const UNISWAP_MINT_CONTRACT = '0x514906FC121c7878424a5C928cad1852CC545892';
```
bonustrack, line 8: Your strategy should be made generic enough so it could be used by others. You shouldn't hardcode an address here you can use the strategy options for that. There is another PR here that overlap with what you are doing https://github.com/bonustrack/snapshot.js/pull/7 Also i think a better approach would be to do it this way: https://github.com/bonustrack/snapshot.js/pull/7#issuecomment-703250682

bonustrack:
@chrouzz Hi! I'm closing this PR because we have now a strategy that can do the same, have a look here on the examples query if you want to use it: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/uniswap/examples.json or feel free to ask in Discord or open an issue if you need more informations.

PR #7, opened 2020-10-01
Added Uniswap LP

vfatouros:
- Added support for extracting the number of tokens locked in Uniswap liquidity provision, to include these in voting.

Unfortunately I don't have any way to test this, if you spot an issue I would be grateful if you could point me in the right direction.


bonustrack:
@vfatouros You can test using this file https://github.com/bonustrack/snapshot.js/blob/master/test/index.ts you will need to run build script first

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,41 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import { uni_abi } from './UniV2.json';
+import { token_abi } from './TestToken.json';
```
bonustrack, line 4: You can reuse the same json file than on `erc20-balance-of` folder

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,41 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import { uni_abi } from './UniV2.json';
```
bonustrack, line 3: Var names should be camelCased
```suggestion
import { uniAbi } from './UniV2.json';
```

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,41 @@
...
+      network,
+      provider,
+      token_abi,
+      [options.tokenAddress, 'balanceOf', [options.address]],
+      { blockTag }
+  )).map((value, _) => value)[0];
+
+  const tokensPerUni = (tokenBalanceInUni / 10 ** options.decimals) / (totalSupply / 1e18);
+
+  const response = await multicall(
```
bonustrack, line 27: It will be better if the 2 multicall request happen in parallel, you can use something like
```js
const res = await Promise.all([
  multicall(...)
  multicall(...)
];
```

vfatouros:
Thank you very much for your comments, incredibly helpful! I have amended accordingly and tested it.

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,46 @@
...
+      provider,
+      uniAbi,
+      [[options.address, 'totalSupply', []]],
+      { blockTag }
+    ),
+    multicall(
+      network,
+      provider,
+      tokenAbi,
+      [[options.tokenAddress, 'balanceOf', [options.address]]],
```
bonustrack, line 22: I'm not sure to understand why you request `balanceOf` for both the pool token and the token itself. Can you explain?

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    multicall(
+      network,
+      provider,
+      uniAbi,
+      addresses.map((address: any) => [options.address, 'balanceOf', [address]]),
+      { blockTag }
+    )
+  ]);
+
+  const totalSupply = res[0].map((value, _) => value)[0];
```
bonustrack, line 34: You don't need the map function here, this is the same as if you were doing this:
```suggestion
  const totalSupply = res[0][0];
```

bonustrack:
It seem that with your approach you need to manually list the Uniswap pool to be counted for. Have you considered an approach where you just need to define the base token and the strategy would take all the token that you hold across all Uniswap pools? This is what we do in the strategy "balancer". It may require the use of Uniswap subgraph.

src/strategies/uniswap/index.ts
vfatouros: For example in the BUILD-ETH uniswap pool, I need to know how many BUILD tokens there are in 1 UNI-V2 token. So I check BUILD's balanceOf the UNI-V2 token address, and get a figure e.g. 10,000 BUILD in total in the contract. I divide that by the UNI-V2 totalSupply to get e.g. 1.5 BUILD per UNI-V2. Then I get UNI-V2's balanceOf the user address, to see how many UNI-V2 tokens the user has. So, if a user has 10 UNI-V2 tokens in the above example, they have 15 BUILD. I don't think it's possible to do this with less than 3 method calls, but of course I could be missing something.

vfatouros:
Indeed I did see your approach with balancer, but then that becomes 2 things I am not experienced in (Javascript and subgraphs/GraphQL). The main counterpoint is that while tokens often use lots of different balancer pools, I would say in Uniswap's case 90% of the time it is only token/eth. I couldn't picture a scenario where a token's governance would need to use 2 different Uniswap pools. That said, I'm more than happy to look into it if you think it will be more performant or just more elegant.

vfatouros:
I've gone ahead and written the strategy using Uniswap's subgraph. I have left both strategies in because the `uniswap` strategy allows you to select one specific pool whereas the `unigraph` strategy checks all of them, but feel free to only keep one if you prefer. I am not sure my pull request is still working correctly as I don't see my latest 2 commits in the history here, I can submit a new one if that helps.

src/strategies/unigraph/index.ts
```diff
@@ -0,0 +1,66 @@
...
+          },
+          reserve1 : true,
+          totalSupply : true
+        }  
+      }
+    }
+  };
+  if (snapshot !== 'latest') {
+    // @ts-ignore
+    params.poolShares.__args.block = { number: snapshot };
```
bonustrack, line 42: The param `poolShares` doesn't exist here, it should be changed to `users` like this:
```suggestion
    params.users.__args.block = { number: snapshot };
```

src/strategies/unigraph/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+  const tokenAddress = options.address.toLowerCase();
+  const result = await subgraphRequest(UNISWAP_SUBGRAPH_URL[network], params);
+  const score = {};
+  if (result && result.users) {
+    result.users.forEach(u => {
+      u.liquidityPositions.filter(lp => 
+        lp.pair.token0.id == tokenAddress 
+        || lp.pair.token1.id == tokenAddress)
+                          .forEach(lp => {
```
bonustrack, line 52: I can see the space intend is wrong here, can you run the lint script? That should fix that. You can use `npm run lint`

src/strategies/index.ts
```diff
@@ -20,8 +22,10 @@ export default {
   'eth-balance': ethBalance,
   'maker-ds-chief': makerDsChief,
   uni,
+  uniswap,
   'yearn-vault': yearnVault,
   moloch,
+  unigraph,
```
bonustrack, line 28: Is it possible to remove the `uniswap` single pool strategy from this PR and rename `unigraph` strategy to `uniswap`? I think the multi pool strategy will be what projects will use, i would prefer if it take the uniswap name.

src/strategies/unigraph/index.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export async function strategy(network, provider, addresses, options, snapshot) {
+  const params = {
+    users: {
+      __args: {
+        where: {
+          id_in: addresses.map(address => address.toLowerCase())
+        },
+        first: 10,
```
bonustrack, line 15: Any reason to limit to 10 here? We can have 1000 instead
```suggestion
        first: 1000,
```

src/strategies/unigraph/index.ts
```diff
@@ -0,0 +1,66 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const UNISWAP_SUBGRAPH_URL = {
+  1: 'https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v2',
```
bonustrack, line 5: Can you map others Uniswap subgraphs here? Also some new fields are missing (version and github author) see example here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/balancer/index.ts#L4-L5

bonustrack (commented):
@vfatouros Sorry for the delay and thank you for your contribution! I've tested your PR and it seem to work well, i've suggested some changes. Please let me know if you can do the changes or if you prefer us to do it. You'll also need to merge master to your PR and fix conflicts.

src/strategies/unigraph/index.ts
vfatouros: I wasn't able to find an official testnet subgraph like balancer has, wasn't sure whether to use a community one.

src/strategies/uniswap/index.ts
```diff
@@ -0,0 +1,79 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const UNISWAP_SUBGRAPH_URL = {
+  1: 'https://api.thegraph.com/subgraphs/name/uniswap/uniswap-v2'
+};
+
+export const author = 'bonustrack';
```
bonustrack, line 8:
```suggestion
export const author = 'vfat';
```

src/strategies/balancer/index.ts
```diff
@@ -1,7 +1,7 @@
 import { getAddress } from '@ethersproject/address';
 import { subgraphRequest } from '../../utils';
 
-export const author = 'bonustrack';
+export const author = 'vfat';
```
bonustrack, line 4:
```suggestion
export const author = 'bonustrack';
```

src/strategies/uniswap/examples.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "uniswap",
+      "params": {
+        "address": "0x6b175474e89094c44da98b954eedeac495271d0f",
+        "symbol": "DAI",
```
bonustrack, line 8: You don't need decimals param here, it's abstracted with thegraph.

src/strategies/uniswap/examples.json
```diff
@@ -0,0 +1,18 @@
...
+        "decimals": 18
```
bonustrack, line 9: You don't need decimals param here, it's abstracted with thegraph.

bonustrack (approved):
It seem to be working well, well done! The PR is ready to go, sorry that it took that long.

PR #10, opened 2020-10-06
add pancake strategy

pancake-swap:


src/strategies/pancake/index.ts
```diff
@@ -0,0 +1,64 @@
+import { formatUnits } from "@ethersproject/units";
+import { multicall } from "../../utils";
+import { strategy as erc20BalanceOfStrategy } from "../erc20-balance-of";
+
+export const author = "pancake";
```
bonustrack, line 5: This must be a your GitHub username.

src/strategies/pancake/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    type: "function",
+  },
+];
+
+export async function strategy(network, provider, addresses, options, snapshot) {
+  const blockTag = typeof snapshot === "number" ? snapshot : "latest";
+  const score = await erc20BalanceOfStrategy(network, provider, addresses, options, snapshot);
+
+  const balances = await Promise.all(
+    options.chefAddresses.map((chefAddress) =>
```
bonustrack, line 35: I'm concerned that this will send too much request at the same time and overload user browsers, is there a way you can call the contract with only 1 to 3 requests?

bonustrack (commented):
Added some suggestions, would be nice to have a README.md file too similar than this strategy https://github.com/bonustrack/snapshot.js/tree/master/src/strategies/contract-call

src/strategies/pancake/index.ts
pancake-swap: chefAddresses will be only 2-4 length

PR #11, opened 2020-10-15
add synthetix strategy

andytcf:
Adds SNX strategy for vote weightings

- Quadratic voting
- Voters weight based on debt percentage at previous fee period

### DIFF
--- a/src/strategies/synthetix/index.ts
+++ b/src/strategies/synthetix/index.ts
@@ -1,36 +1,36 @@
-import { formatUnits } from "@ethersproject/units";
-import { multicall } from "../../utils";
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
 
-export const author = "andytcf";
-export const version = "0.0.1";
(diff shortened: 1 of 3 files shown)


src/strategies/synthetix/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      { name: "debtEntryIndex", type: "uint256" },
+    ],
+    payable: false,
+    stateMutability: "view",
+    type: "function",
+  },
+];
+
+const synthetixStateContractAddress =
+  "0x4b9Ca5607f1fF8019c1C6A3c2f0CC8de622D5B82";
```
bonustrack, line 23: Is there others same smart contract with different address, for example testnet contracts? If yes we may want to have this value as part of the strategy options.

andytcf: sure, i've seen the implementation elsewhere but I think for now we want to only include the main-net contract

bonustrack (approved):
Works well!

PR #16, opened 2020-10-20
Add new strategy

jxdv:


src/strategies/erc20-balance-of-coeff/index.ts
```diff
@@ -18,6 +18,6 @@ export async function strategy(
     snapshot
   );
   return Object.fromEntries(
-    Object.entries(score).map((address) => [address[0], options.coeff])
+    Object.entries(score).map((address:any) => [address[0], address[0] * options.coeff])
```
bonustrack, line 21:
```suggestion
    Object.entries(score).map((address: any) => [address[0], address[1] * options.coeff])
```

bonustrack:
You need to map the strategy here also https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/index.ts

PR #21, opened 2020-10-26
Add Tokenlon strategy

BenjaminLu:
- Added Tokenlon(https://tokenlon.im) strategy.


src/strategies/tokenlon/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [TFT_ADDRESS, 'balanceOf', [address]]),
+    { blockTag }
+  );
+  return Object.fromEntries(
+    response.map((value, i) => [
+      addresses[i],
+      (parseFloat(formatUnits(value.toString(), 0)) > 0 ? 1 : 0)
```
bonustrack, line 51: This strategy will give 1 score to any token holder regardless of their balance. Is this what you trying to do? Maybe you can change the name of the strategy to be more generic like "erc20-with-balance" so it can be used by others spaces. How are you dealing with sybil (someone could create 10000 wallets and add 0.0001 on each wallet and get also a 1 score)? Can you create a README.md like some others strategies to give more explanations, example: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/synthetix

BenjaminLu: You're right. TFT (TokenFans Token) currently is just a badge token. It can't be transfer between users unless your address is in the whitelist. imToken plans to use the badge token for governance only at the very beginning. We would issue a governance proposal for the next stage. And let the community decide the final governance mechanism.

You could find the code about transfer here:
https://etherscan.io/address/0xe534619defdbf0caf673b8abf7158714f5bd4bd9#code at line 541

Since the decimals property of the badge token is 0,  people can't divide their tokens into pieces. So we don't need to deal with sybil. Check the contract properties here:
https://etherscan.io/token/0xe534619defdbf0caf673b8abf7158714f5bd4bd9

src/strategies/tokenlon/index.ts
```diff
@@ -0,0 +1,54 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'TokenlonDev';
```
bonustrack, line 4: This must be a valid GitHub username

src/strategies/tokenlon/index.ts
```diff
@@ -0,0 +1,54 @@
...
+export const version = '0.1.0';
+
+const TFT_ADDRESS = '0xE534619dEFDBF0cAf673b8AbF7158714F5BD4bd9';
```
bonustrack, line 7: Can you make this a parameter in the strategy?

src/strategies/tokenlon/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        type: 'uint256'
+      }
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
+
+export async function strategy(
```
bonustrack, line 33: Instead of writing again the `erc20-balance-of` strategy you can import it like this: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of-coeff/index.ts that will simplify / reduce the code.

src/strategies/tokenlon/index.ts
bonustrack: Thanks for the details, then that make lot of sense, this is more like an NFT. If you can do the changes i pointed out that would be great, then we can go forward and merge this strategy in prod.

src/strategies/tokenlon/index.ts
BenjaminLu: Done. A big thx.

src/strategies/tokenlon/index.ts
BenjaminLu: Done.

src/strategies/tokenlon/index.ts
BenjaminLu: Done. A big thx again.

src/strategies/tokenlon/index.ts
BenjaminLu: Done.

bonustrack: Can you change the strategy name from `tokenlon` to `erc20-with-balance`? This strategy can be used by others i would prefer if there is a generic name for it.

BenjaminLu: All done. Need to update the space configuration after doing this.

bonustrack (approved):

PR #22, opened 2020-10-27
Add cream-long-term-pool

jeremyHD:
- Rename `cream-swap` strategy to `cream`
- Implement Cream Long Term Pool Strategy in order to reduce contract calls


bonustrack:
@jeremyHD If you want to add 2 strategies you need to create 2 PRs and have the strategies in a separated folders. You can not add 2 strategies in one.

jeremyHD:
@bonustrack  I've updated the PR as required. Please take a look Thanks!

src/strategies/cream/index.ts
```diff
@@ -0,0 +1,144 @@
...
+export async function strategy(
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+
+  const creamAddress = options.pools.find(pool => pool.name === 'CREAM').address;
+  let score = await creamSwapScore(network, addresses, creamAddress, snapshot);
```
bonustrack, line 46: Can you make the 2 requests happen in parallel with something like this:
```js
let [score, creamBalance] = await Promise.all([
  creamSwapScore(network, addresses, creamAddress, snapshot),
  creamBalanceOf(network, provider, addresses, options, snapshot)
]);
```
This will be faster to resolve than if we process request one by one

jeremyHD: Good call, will do that!

bonustrack (commented):
The PR looks good but i'm not sure how to test it, can you add a README.md file with some short description on what the strategy does and how to setup params for it, also include some test address that give positive score (along with a snapshot block number). Thank you!

jeremyHD:
README added, feel free to ask any questions @bonustrack

bonustrack (approved):
@jeremyHD Looking good! Can you just fix conflicts then we can merge it.

jeremyHD:
@bonustrack rebased. Thanks!

PR #27, opened 2020-11-03
Strategy merge - MANTRA DAO v1

spaceforce-dev:
Fixes # .

- MANTRA DAO voting strategy
- OM staking balance merged with OM-UNIV2 
- OM-UNIV2 token has a higher weight


src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,76 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'Spaceforce';
```
bonustrack, line 4: This should be your github username
```suggestion
export const author = 'spaceforce-dev';
```

src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const increaceWeighting = (number) => {
+    return number * options.omUniScale
+  }
+
+  const omStakingResponse = await multicall(
```
bonustrack, line 44: Can you merge the 2 multicall in one? That will avoid sending 2 requests from user browsers and make result faster to resolve.

bonustrack (commented):
Can you add a file examples.json in your strategy to make it easy to test on test/index.ts file, see example here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of/examples.json The example should give positive score.

src/strategies/mantra-dao/index.ts
spaceforce-dev: Ok, Can you give any ideas how to do that, i'm not familiar with multi-call. Thanks for the help!

src/strategies/mantra-dao/index.ts
spaceforce-dev: Ok will do

src/strategies/mantra-dao/index.ts
bonustrack: You will need to merge the 2 arrays of calls together:
- This one: https://github.com/snapshot-labs/snapshot.js/pull/27/files#diff-b5ff3ff9952840d980dad2da4ac1c03ae1bee0195913c96fe1f895937990f30fR48
- And this one https://github.com/snapshot-labs/snapshot.js/pull/27/files#diff-b5ff3ff9952840d980dad2da4ac1c03ae1bee0195913c96fe1f895937990f30fR55

You can use `array1.concat(array2)` for this. Then you would do a single multicall and separate the results you get in 2 using the array length. See example here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/ctoken/index.ts#L92-L93

spaceforce-dev:
Updated to come from MANTRA DAO git account 

https://github.com/snapshot-labs/snapshot.js/pull/29

PR #28, opened 2020-11-04
Add DIAData strategy

bitsandtea:
Adds the DIAData strategy. See src/strategies/diadata/README.md for details.


src/strategies/diadata/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    addresses,
+    options,
+    snapshot
+  ) {
+    const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+    const response = await multicall(
+      network,
+      provider,
+      abi,
+      addresses.map((address: any) => [options.address, 'balanceOf', [address]]),
```
bonustrack, line 44: Can you import the strategy `erc20-balance-of` here like this: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of-coeff/index.ts#L13
This help reduce duplicate code. You could also copy this part https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of-coeff/index.ts#L20-L25 and just change this line
```
address[1] * options.coeff
```
to this
```
address[1] > options.minTokenBalance ? 1 : 0
```
Also can you change the name of the strategy to be more generic, i recommend 'erc20-balance-of-minimum'

src/strategies/diadata/index.ts
```diff
@@ -0,0 +1,61 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'diadata';
```
bonustrack, line 4: This should be a valid github username

bitsandtea: Thanks, done!

src/strategies/diadata/index.ts
bitsandtea: Thanks, much cleaner now - done!

bonustrack:
@bitsandtea Sorry i just realized that actually there is a strategy that could do what you want with a little change. Have a look here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-with-balance/index.ts This strategy will give 1 score if you hold a specific token otherwise will give 0. I'm going to update that in the next few mn to support a minimum balance. Could you use this strategy instead of using a new strategy?

bitsandtea:
> @bitsandtea Sorry i just realized that actually there is a strategy that could do what you want with a little change. Have a look here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-with-balance/index.ts This strategy will give 1 score if you hold a specific token otherwise will give 0. I'm going to update that in the next few mn to support a minimum balance. Could you use this strategy instead of using a new strategy? Sorry for the hassle i should have realized earlier!

Yes thats fine by us

PR #29, opened 2020-11-05
MantraDAO strategy

spaceforce-dev:
MANTRA DAO voting strategy
OM staking balance merged with OM-UNIV2
OM-UNIV2 token has a optional higher weighting in scores


yarn.lock
```diff
@@ -0,0 +1,2061 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
bonustrack, line 1: This repo don't use Yarn only NPM, can you remove this file?

test/index.ts
```diff
@@ -1,22 +1,52 @@
-global['fetch'] = require('cross-fetch');
+// const { JsonRpcProvider } = require('@ethersproject/providers');
```
bonustrack, line 1: This file shouldn't be changed, and you should create a file "examples.json" using this format https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of/examples.json with a sample of request for your strategies that will give positive score.

src/utils.ts
```diff
@@ -1,3 +1,4 @@
+// import fetch from 'node-fetch';
```
bonustrack, line 1: This shouldn't be there aswell, can you merge latest changes from master?

src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const increaceWeighting = (number, increaceFactor) => {
+    return number * increaceFactor
+  }
+
+  addresses = [...addresses, ...addresses]
```
bonustrack, line 44: You shouldn't rewrite the "addresses" param here, can you use a different name for it?

src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  console.log(addresses)
```
bonustrack, line 46: There shouldn't be console.log in the strategy

src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      ]),
+      { blockTag }
+    );
+  } catch (error) {
+    console.log(error)
+  }
+
+  console.log(omStakingResponse)
+  
+  // const uniStakingResponse = await multicall(
```
bonustrack, line 65: Can you remove the comments here?

src/strategies/mantra-dao/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      addresses[i],
+      parseFloat(formatUnits((value).toString() , options.decimals))
+    ])
+  
+    addresses.forEach((address, i) => {
+      uni[i] = [address, 
+        (increaceWeighting(uni[i][1], 4) + om[i][1])]
+    });
+    let combined = uni
+    return {scores: Object.fromEntries(
```
bonustrack, line 89: The format you using for return is wrong, the strategy should return an object with a single level deep, like this:
```
{
  '0x123...': 100,
  '0x456...': 250,
  '0x789...': 1500
}
```

test/index.ts
```diff
@@ -1,22 +1,52 @@
...
-    console.error(e);
+
+const network = 1;
+const url = 'https://eth-mainnet.alchemyapi.io/v2/4bdDVB5QAaorY2UE-GBUbM2yQB3QJqzv';
+const provider = new JsonRpcProvider(url);
+
+const strategies = [
+  {
+    name: 'mantra-dao',
+    params: {
```
bonustrack, line 31: This should be part of an examples.json file

bonustrack (commented):
Added few suggestions, can you explain what the strategy does in a README.md file like others strategy example: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/cream/README.md also when you done the change you should run the script "lint" to fix lint issues.

spaceforce-dev:
Thank you for the thorough feedback.

I have merged all the changes you wanted.

src/strategies/mantra-dao/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Our tokens are the OM erc20 and the UNI V2 OM synthetic token found at:
+omStakingAddress and omUniStakingAddress
+
+Example:
+
+A address has a staked balance of 10 in contract-1 and a staked balance of 10 in contract-2.
+
+with a scaling factor of 1.0 their output score will be 20 (10 + 10)
+with a scaling factor of 2.0 their output score will be 30 (10 + 20)
```
bonustrack, line 16: Sorry i think this is something you can do already with another strategy called "erc20-balance-of-coeff". Here is an example of config you can use:
```json
{
  "strategies": [
    {
      "name": "erc20-balance-of",
      "params": {
        "address": "0x2bCD929283Ad0EE603E743412ddb214b91fbab88",
        "symbol": "OM",
        "decimals": 18
      }
    },
    {
      "name": "erc20-balance-of-coeff",
      "params": {
        "address": "0x99b1db3318aa3040f336fb65c55400e164ddcd7f",
        "symbol": "OMUNI",
        "decimals": 18,
        "coeff": 2
      }
    }
  ]
}
```
Can you check to see if you can use that strategy?

spaceforce-dev: Ok good suggestion.

Our problem with using that is, this only a first iteration, and were likely adding more complexity as time goes on, so we were hoping to have our own strategy that we can change.

spaceforce-dev: Ok perhaps we can use this one for now to get us going and propose our own when we alter it. Appreciate the advice!

gokaiorg:
Closing this PR for now.

PR #30, opened 2020-11-05
Add DeFiat Voting Power strategy

QuantSoldier:
Adds a voting power strategy specific to DeFiat and its staking system. The goal of this PR is to be able to fairly allocate votes among all participants of the network, regardless of token holding status. This proved to be too gas-intensive to manage ourselves, so to Snapshot we go!

- Add DeFiat Voting Power Strategy & README 
- Smart Contract code is published on Etherscan

https://etherscan.io/address/0x594aca0b33b041b8d66d482daccc7819fee45e0a#code

### DIFF
--- a/src/strategies/defiat/README.md
+++ b/src/strategies/defiat/README.md
@@ -1,0 +1,13 @@
+# DeFiat
+
+Snapshot strategy to calculate total DFT exposure at a given block and grant equivalent voting power via Smart Contract call.
+
+DeFiat_VotingPower Contract: https://etherscan.io/address/0x594aca0b33b041b8d66d482daccc7819fee45e0a#code
+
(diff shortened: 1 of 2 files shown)


bonustrack:
@QuantSoldier Hi! It seem like you could use the strategy "contract-call" to do the same thing, have a look here: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call

QuantSoldier:
> @QuantSoldier Hi! It seem like you could use the strategy "contract-call" to do the same thing, have a look here: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call

Thanks @bonustrack, I have updated my PR for in the `snapshot-spaces` project with these adjustments. Will close this one now.

PR #31, opened 2020-11-05
Aragon Plugin: Correct contract call, add call to get nonce

Evalir:
- Remove `id` property from the token on the contract call.
- Add call to get the nonce variable from the contract, rather than the account nonce itself.

### DIFF
--- a/src/plugins/aragon/index.ts
+++ b/src/plugins/aragon/index.ts
@@ -1,7 +1,7 @@
 import BN from 'bn.js';
 import { keccak256 } from '@ethersproject/keccak256';
 import { toUtf8Bytes } from '@ethersproject/strings';
-import { sendTransaction, subgraphRequest } from '../../utils';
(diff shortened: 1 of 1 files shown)


src/plugins/aragon/index.ts
```diff
@@ -197,7 +197,7 @@ async function scheduleAction(
   const config = result.registryEntry.queue.config;
 
   // Building the nonce for the next tx
-  const nonce = await web3.getTransactionCount(account, 'pending');
+  const nonce = await call(web3, abi, [result.registryEntry.queue.address, 'nonce', ''])
```
bonustrack, line 200: You'll probably need to update the ABI code here to include the nonce method, this part https://github.com/aragon/govern/blob/e98ce3613b1fb93f5ce103ddf5f4dea3a1846077/packages/govern-console/src/lib/abi/GovernQueue.json#L1281

PR #32, opened 2020-11-10
EC-712 primitives for frontend and backend

adridadou:
Fixes # .
This uses ERC-712 for signing votes and proposals as opposed to the current approach that uses JSON.stringify(...) and signs the raw string

- adds signVoteMessage to sign (or request signature) for a vote message
- adds validateMessage to validate that a signature is correct
- creates an interface for the signer so we can sign either with web3 or with ethSigUtil. This is particularly useful when you want to write tests and you don't have access to ERC-712 signing in your test environment


jaceppo (approved):

jaceppo (approved):
guanlong huang

bonustrack:
Closing this PR for now, the changes to support ERC-712 has been implemented already here https://github.com/snapshot-labs/snapshot.js/tree/master/src/sign

PR #37, opened 2020-11-17
Adds staked Uniswap LP tokens strategy

vfatouros:
This is for the common case where LP tokens are deposited into a staking contract.


src/strategies/staked-uniswap/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      { blockTag }
+    ),
+    multicall(
+      network,
+      provider,
+      tokenAbi,
+      [[options.tokenAddress, 'balanceOf', [options.uniswapAddress]]],
+      { blockTag }
+    ),
+    multicall(
```
bonustrack, line 55: It would be great if you can make a single multicall to gather all info you need at once. This way it will only have a single request to be sent instead of 3, it's more light for the users browsers.

src/strategies/staked-uniswap/UniV2.json
```diff
@@ -0,0 +1,568 @@
+{
```
bonustrack, line 1: We don't need the whole file here, it seem that only `balanceOf` and `totalSupply` inputs are needed.

src/strategies/uniswap/index.ts
```diff
@@ -9,9 +9,9 @@ export const author = 'vfatouros';
 export const version = '0.1.0';
 
 export async function strategy(
-  space,
+  _space,
   network,
-  provider,
+  _provider,
```
bonustrack, line 14: Any reason to do these change?

vfatouros: Just hides the "unused variable" hint

src/strategies/staked-uniswap/index.ts
vfatouros: Yes I should have looked at multicall a bit closer! That is resolved now.

src/strategies/staked-uniswap/UniV2.json
vfatouros: Good point, I've removed the file

bonustrack (approved):

PR #38, opened 2020-11-20
Add Gnosis Impact plugin

davidalbela:
A a new plugin to look for prediction markets.

The plugin includes two methods:
- getTokenInfo: retrieve ERC20 token name and symbol from a given token address.
- getSubgrapInfo: to query the Omen predition market subgraph by a given `conditionId`.


src/plugins/gnosis/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+const getTokenMethod = async (web3, tokenAddress, method) => {
+    return await call(web3, erc20Abi, [
+      tokenAddress,
+      method
+    ]);
+}
+
+export default class Plugin {
+  public author = 'David';
```
bonustrack, line 58: This param should be your GitHub username
```suggestion
  public author = 'davidalbela';
```

PR #39, opened 2020-11-20
Add strategy that includes unclaimed tokens and pooled LP tokens

ncitron:
- Add a strategy that includes unclaimed tokens and pooled LP tokens.
- This will be used for Indexcoop's voting scheme.


ncitron:
Looks like this can be done using the contract-call strategy. Closing the PR.

PR #43, opened 2020-11-25
Please add Meter.io mainet information

xiaohanzhu:
Need to add the following in web3 to interact with Meter mainnet
const meterify = require(“meterify”).meterify;
const Web3 = require(“web3");
const web3 = meterify(new Web3(), “https://rpc.meter.io”);

Fixes # .

adding meter main net network information and multicall contract address
The meterify library can be found in https://github.com/meterio/meterify
- 
-

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -82,6 +82,16 @@
     ],
     "explorer": "https://blockscout.com/etc/mainnet"
   },
+  "82": {
+    "key": "82",
+    "name": "Meter.io Mainnet",
+    "chainId": 82,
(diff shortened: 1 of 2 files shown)


bonustrack:
@xiaohanzhu How is `meterify` module different than using Ethers.js JsonRpcProvider? It will be easier to integrate if Meter.io is compatible with all the Ethereum tooling.

xiaohanzhu:
Yes, it should work as we have an emulation layer for the RPC interface.  Let me know if you run into any issues.

PR #44, opened 2020-11-25
feat: update pancake strategy

pancake-swap:
- add masterChef pool for strategy


bonustrack:
@pancake-swap Can you provide a file "examples.json" to make it easier to test? The query in examples file should return a positive score. See example here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of/examples.json

PR #45, opened 2020-11-26
StableXSwap Governance Params

stablexswapdev:
Fixes # .

- Adds stablexswap (STAX) token
- 
-

### DIFF
--- a/src/strategies/StableXSwap/README.md
+++ b/src/strategies/StableXSwap/README.md
@@ -1,0 +1,22 @@
+# Contract call strategy
+
+Allows the tokens staked in chef contracts to be used to calculate voter scores.
+
+## Examples
+
+Can be used instead of the erc20-balance-of strategy, the space config will look like this:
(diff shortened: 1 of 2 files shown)


src/strategies/StableXSwap/README.md
```diff
@@ -0,0 +1,22 @@
+# Contract call strategy
```
bonustrack, line 1: This should be changed to the strategy name

bonustrack (commented):
Some comments:
- The folder of the strategy should be lowercase. 
- The strategy should be added here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/index.ts
- Also can you add a `examples.json` file with a sample query to try with your strategy. The sample must return a positive number when tested. [See example](https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of-delegation/examples.json).

stablexswapdev:
will open a new PR with the fixes, thanks!

PR #46, opened 2020-11-26
Add ESD Strategy

l3wi:
- Adds `esd` strategy to read bonded ESD tokens
- Adds a `readme.md` for the strategy
- Adds an `example.json` file to test locally.


src/strategies/esd/example.json
```diff
@@ -0,0 +1,21 @@
...
+                "token": "0x36F3FD68E7325a35EB768F1AedaAe9EA0689d723",
+                "decimals": 18
+            }
+        },
+        "network": "1",
+        "addresses": [
+            "0x5b3d22cf16963eb57fea96482728d91e885cb71e",
+            "0xd1991b6fd521a3f357d96a15956702ce33342fec"
+        ],
+        "snapshot": "latest"
```
bonustrack, line 19: It's better to use a specific block number here, this way we will always get the same result.

src/strategies/esd/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  // Fetch DAO Bonded tokens
+  const daoBondedBalance = await multicall(
```
bonustrack, line 47: The 3 multicalls could be merged into a single one, if you can do this that would make the strategy resolve much faster.

src/strategies/esd/example.json
l3wi: Snapshot added

src/strategies/esd/index.ts
l3wi: Merged

l3wi:
Updated code to reflect requested changes and merged upstream changes

src/strategies/esd/index.ts
```diff
@@ -0,0 +1,97 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'lewi';
```
bonustrack, line 4: This must be your GitHub username
```suggestion
export const author = 'l3wi';
```

bonustrack (approved):
Works well! Great work on the PR.

PR #55, opened 2020-12-09
Feature/piedao strategy

Alexintosh:
Fixes # .

- Introduces a PieDAO strategy to calculate the vote weights across different representations of DOUGH


bonustrack (commented):
Looks great! Could you just remove the changes on the unrelated files? You can also remove `test/piedao.ts`, it's easy to use `test/index.ts` to test any strategy already. Ideally you should have only these files added/changed in your PR:
image

PR #56, opened 2020-12-13
Add Keeper Bonds strategy

akshaynexus:
Fixes # .

- Add strategy to pull bonded balance in a keeper token contract.


bonustrack:
@akshaynexus If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call
Here is an example of a space that is using "contract-call" strategy to get voting score https://hub.snapshot.page/api/spaces/badgerdao.eth

bonustrack:
Closing this PR for now, feel free to open it again if needed.

PR #60, opened 2020-12-17
🙌 ETH Philanthropy Strategy

mccallofthewild:
Fixes https://github.com/snapshot-labs/snapshot-bounties/issues/1

### 📝 Summary 
Allows addresses to increase their ETH balance-based score by donating to charity, adding 100 ETH to their score for every 1 ETH donated. This helps to level the playing field for less wealthy participants and incentivizes altruistic behavior from all participants. 

### 👨‍💻 Technologies 
* Uses ElasticSearch API from [AnyBlock Analytics](https://www.anyblockanalytics.com/) to query transactions from participants' addresses to charity addresses. 

### 🎁 Charities Currently Included:
[GiveDirectly](https://www.givedirectly.org/): 0xc7464dbcA260A8faF033460622B23467Df5AEA42
[Unsung.org](Unsung.org): 0x02a13ED1805624738Cc129370Fee358ea487B0C6
[Heifer.org](Heifer.org): 0xD3F81260a44A1df7A7269CF66Abd9c7e4f8CdcD1
[GraceAid.org.uk](GraceAid.org.uk): 0x236dAA98f115caa9991A3894ae387CDc13eaaD1B
[SENS.org](SENS.org): 0x542EFf118023cfF2821b24156a507a513Fe93539
[350.org](350.org): 0x50990F09d4f0cb864b8e046e7edC749dE410916b
[EFF.org](EFF.org): 0xb189f76323678E094D4996d182A792E52369c005
[WikiLeaks](WikiLeaks.org): 0xE96E2181F6166A37EA4C04F6E6E2bD672D72Acc1
[GiveWell.org](GiveWell.org): 0x7cF2eBb5Ca55A8bd671A020F8BDbAF07f60F26C1
[CoolEarth.org](CoolEarth.org): 0x3c8cB169281196737c493AfFA8F49a9d823bB9c5
[Run2Rescue.org](Run2Rescue.org): 0xd17bcbFa6De9E3741aa43Ed32e64696F6a9FA996
[Archive.org](Archive.org): 0xFA8E3920daF271daB92Be9B87d9998DDd94FEF08"


src/strategies/eth-philanthropy/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  ['Archive.org', '0xFA8E3920daF271daB92Be9B87d9998DDd94FEF08']
+];
+
+export async function strategy(
+  ...args: [
+    string,
+    string,
+    Web3Provider,
+    string[],
+    { erc20TokenAddress?: string },
```
bonustrack, line 29: This doesnt seem to be used

bonustrack:
That's a really cool idea! Some suggestions:
- Instead of hardcoding 100 ETH you could make it a strategy parameter example "coeff: 100" this would give more flexibility on how much power you want to give to altruistic voters.
- You don't need to include the "eth-balance" score in this strategy. If the space want the eth balance he can add that strategy separately, this will make it possible to use the philanthropy strategy with "erc20-balance-of" strategy for example, or use only the philanthropy strategy if you want to count only donations.
- I think this strategy could be made more generic to serve more use case if the list of addresses that receive donation is a parameter. We can imagine some projects that are looking for donors like Gitcoin or Snapshot using this strategy to give voting power on users who donated ETH to them. I would recommend make 2 strategies for this, one that you can call something like "eth-received" and where you can define some receiving addresses as params. And a strategy "eth-philanthropy" that leverage "eth-received" strategy and where the receiving addresses are hardcoded with the list you made.

mccallofthewild:
@bonustrack Thanks! That's phenomenal feedback. I'm on board. Donation-driven governance will be a great incentive for organizations to use Snapshot.

I'll try to get those changes pushed up later today. 👍

src/strategies/eth-philanthropy/index.ts
mccallofthewild: I was initially going to make it a dynamic strategy that worked with both ETH (default) and ERC20 tokens. But I decided to keep it simple for v1. This should be fairly straightforward to implement though. 

Would it be preferrable to have a couple generic strategies (e.g. "philanthropy", "amount-received") or several strategies specialized by currency type (e.g. "eth-philanthropy",  "eth-received",  "erc20-philanthropy",  "erc20-received") ?

bonustrack: It's better to have specialized ones, but i dont think we need an ec20-philanthropy strategy, its too specialized i dont think someone will use it.

mccallofthewild:
## Updates 

* Created `eth-received` strategy with parameters `receivingAddresses` and `coeff`
* Delegated `eth-philanthropy` logic to `eth-received`
* Created `erc20-received` strategy in the same form as `eth-received`, with additional parameters `contractAddress` and `decimals`

src/strategies/erc20-received/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  const requestDelay = 1000 / 5;
+  let iteration = 0;
+  let lastFetchTime = Date.now();
+
+  for (let fromAddress of addresses)
+    for (let toAddress of receivingAddresses) {
+      await new Promise((r) =>
+        setTimeout(r, requestDelay - Math.max(0, Date.now() - lastFetchTime))
+      );
+      const r = await fetch(
```
bonustrack, line 42: The issue is that this type of request don't scale well, there is some proposal with 1000 votes but we don't want user send 1000 request in their browser. Is there a different API you could use to achieve this?

mccallofthewild: It turns out that web3's `#getLogs` method is perfect for this. Just pushed the update 👍

bonustrack: Hum i think with getLogs there is a limit of 1000 results, can you try with an address with long history to see if that break or not?

bonustrack: Dfuse (https://docs.dfuse.io) could be an alternative solution

mccallofthewild: Infura has an increased log limit of 10,000 results, but Dfuse is definitely the way to go. It's more consistent and *extremely* scalable.  That was a great suggestion

mccallofthewild:
## Updates

* Implemented `erc20-received` with scalable Dfuse GraphQL subscription query
* Edited `test/index.ts` to enable testing specific strategies from the command line\*
  * Example: `npm run test --strategy=erc20-received`
  
\* may merit a separate pull request

src/strategies/erc20-received/index.ts
```diff
@@ -0,0 +1,154 @@
+import fetch from 'cross-fetch';
+import { InfuraProvider, Web3Provider } from '@ethersproject/providers';
+import { formatUnits } from '@ethersproject/units';
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber';
+import { createDfuseClient } from '@dfuse/client';
```
bonustrack, line 5: I'd prefer if we don't introduce a new dependency here, that would make the library heavier, i've seen dfuse being used just with fetch here https://github.com/snapshot-labs/snapshot.js/pull/66/files could you use the same approach?

mccallofthewild: No problem! Fixed in https://github.com/snapshot-labs/snapshot.js/pull/60/commits/4fcb966ae78c4cbab780118f5ed39294d945d6b0

package.json
```diff
@@ -20,7 +20,8 @@
     "ajv": "^7.0.0-beta.4",
     "bs58": "^4.0.1",
     "ethereumjs-util": "^7.0.7",
-    "json-to-graphql-query": "^2.0.0"
+    "json-to-graphql-query": "^2.0.0",
+    "ws": "^7.4.1"
```
bonustrack, line 24: This doesn't seem to be used anymore you can remove it

test/index.ts
```diff
@@ -2,12 +2,30 @@ global['fetch'] = require('cross-fetch');
...
+  `npm run test` // Tests default (erc20-balance-of)
+  `npm run test --strategy=erc20-received`
+  `npm run test --strategy=eth-balance`
+*/
+
+const strategyNameArgs = Object.keys(snapshot.strategies).map(
+  (name) => `--strategy=${name}`
+);
+const testStrategy =
+  process.argv
```
bonustrack, line 17: This is super cool! For some reason it always give me "erc20-balance-of" strategy i'm not able to see others strategies when i use `--strategy=erc20-received` argument. When i do a console log of process.env, i just get an array 2 file path but i dont see the argument.

mccallofthewild: Good catch. Yarn/NPM incompatibility. Fixed in https://github.com/snapshot-labs/snapshot.js/pull/60/commits/e81ad16fe8632d7eb752f18df5e7a4527c77aee0 👍

bonustrack (approved):
Works well! Thank you for your contribution, looks really good with the new test.

bonustrack:
@mccallofthewild Congrats you win the 2nd place of the Gitcoin competition! We've sent you 300 DAI here: https://etherscan.io/tx/0x62e3b81d3d20df6187223348e5a1351ee6680ca4c227edcf07d80cd9d81b42fb

mccallofthewild:
Thank you! This was a fun challenge. Snapshot is going to be big 💪

PR #63, opened 2020-12-18
Add xDai STAKE strategy

maxaleks:
- added a strategy that calculates the user's voting power based on the balance of STAKE tokens and the amount of tokens deposited on EasyStaking together with the accrued emission

### DIFF
--- a/src/strategies/xdai-stake/index.ts
+++ b/src/strategies/xdai-stake/index.ts
@@ -1,7 +1,6 @@
 import { formatUnits } from '@ethersproject/units';
 import { BigNumber } from '@ethersproject/bignumber';
 import { subgraphRequest, call } from '../../utils';
-import { getBlock } from '../../utils/web3';
 import { strategy as erc20BalanceOfStrategy } from '../erc20-balance-of';
 import { calculateEmission } from './utils';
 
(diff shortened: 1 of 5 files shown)


src/strategies/xdai-stake/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    easyStakingDeposits,
+    { sigmoidParameters, totalSupplyFactor, totalStaked },
+    erc20Score,
+    block,
+    totalSupply,
+  ] = await Promise.all([
+    getEasyStakingDeposits(network, addresses, snapshot),
+    getEasyStakingParams(network, snapshot),
+    erc20BalanceOfStrategy(space, network, provider, addresses, options, snapshot),
+    getBlock(provider, snapshot),
```
bonustrack, line 115: You can use `provider.getBlock(snapshot)` directly the new function `getBlock` is not necessary.

bonustrack (commented):
Looks good! Can you just do the suggestion and run the script "lint" with `npm run lin`? Thanks!

src/strategies/xdai-stake/index.ts
maxaleks: You're right, sorry 😅

maxaleks:
> Looks good! Can you just do the suggestion and run the script "lint" with `npm run lin`? Thanks!

Fixed!

bonustrack (approved):

PR #65, opened 2020-12-22
Aavegotchi strategy

vfatouros:
Aavegotchi has a staking contract where you can stake both GHST and GHST-ETH Uniswap LP tokens, this strategy calculates the sum of GHST in both.


src/strategies/staked-uniswap/UniV2.json
```diff
@@ -0,0 +1,568 @@
+{
```
bonustrack, line 1: This file doesnt seem to be used, can you remove it?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@snapshot-labs/snapshot.js",
-  "version": "0.1.0",
-  "repository": "snapshot-labs/snapshot.js",
+  "name": "@vfatouros/snapshot.js",
```
bonustrack, line 2: Can you remove the changes here, on package-lock and on test/index.ts? Thanks!

vfatouros: Sorry about those! With the test file I wasn't sure if you run the new strategy test too before deploying

PR #66, opened 2020-12-23
ETH Wallet Age strategy

ChaituVR:
# ETH Wallet Age strategy

Allows getting the first transaction date from etherscan, which are used to calculate voter scores with a number of days from the first transaction.

If this is successful we can add ERC20 token age very similarly


src/strategies/eth-wallet-age/index.ts
```diff
@@ -0,0 +1,45 @@
+import fetch from 'cross-fetch';
+
+export const author = 'chaituvr';
+export const version = '0.1.0';
+
+const delay = interval => new Promise(resolve => setTimeout(resolve, interval));
```
bonustrack, line 6: This doesnt seem to be used

src/strategies/eth-wallet-age/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+  ) {
+  const apiLimit = 300;
```
bonustrack, line 16: Why you need this?

ChaituVR: Etherscan has an API limit - https://info.etherscan.com/api-return-errors/

```5 calls per sec/IP with a valid API key```

so 1000/5 ms should be the gap between each API call.. even if we give 200 sometimes it is exceeding the limit. works great with 300

File: src/strategies/eth-wallet-age/index.ts (hunk 1 above)
Comment (ChaituVR) [reply src/strategies/eth-wallet-age/index.ts:6]: Fixed It 👍

File: src/strategies/eth-wallet-age/index.ts [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+  let data = []
+
+  data = await Promise.all(addresses.map((address, index) => {
+    return new Promise(resolve => setTimeout(resolve, apiLimit * index))
```
bonustrack, line 20: We can't send a request for each voter, some proposal have 1000 votes, we dont want the user browser to send 1000 requests. Not sure how you can bypass this.

bonustrack: Maybe using an API like the one used here can work https://github.com/snapshot-labs/snapshot.js/pull/60

ChaituVR:
@bonustrack Replace dfuse API instead of etherscan

src/strategies/eth-wallet-age/examples.json
```diff
@@ -0,0 +1,33 @@
+[
+  {
+    "name": "Age in Days",
+    "strategy": {
+      "name": "eth-wallet-age",
+      "params": {
+        "dfuseApiKey": "web_f527db575a38dd11c5b686d7da54d371"
```
bonustrack, line 7: I think this API Key should be hardcoded within the strategy. Is there some limitation with the API?

src/strategies/eth-wallet-age/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+      node: {
+        from: true,
+        to: true
+      }
+    }
+  }]));
+  
+  const dfuseJWT = await getJWT(options.dfuseApiKey || 'web_f527db575a38dd11c5b686d7da54d371');
+  data = await subgraphRequest('https://mainnet.eth.dfuse.io/graphql', query, {
```
bonustrack, line 54: Is this available on others network? Rinkeby, Kovan etc? If yes we may want to handle it with something like this: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/balancer/index.ts#L7-L10

src/strategies/eth-wallet-age/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    headers: {
+      Authorization: 'Bearer ' + dfuseJWT
```
bonustrack, line 56: We should always use this way
```suggestion
      Authorization: `Bearer ${dfuseJWT}`
```

src/strategies/eth-wallet-age/index.ts
ChaituVR: Only available for Mainnet and Ropsten... but Ropsten isn't available in https://github.com/snapshot-labs/snapshot.js/blob/master/src/networks.json  🤔

src/strategies/eth-wallet-age/index.ts
```diff
@@ -0,0 +1,71 @@
+import { EnumType } from 'json-to-graphql-query';
+import fetch from 'cross-fetch';
+import { subgraphRequest } from '../../utils';
+
+export const author = 'chaituvr';
```
bonustrack, line 5:
```suggestion
export const author = 'ChaituVR';
```

bonustrack (approved):
Well done!

PR #68, opened 2020-12-28
Add wse strategy

Norfolks:
Add a strategy for WSE token.

### DIFF
--- a/src/strategies/wse/index.ts
+++ b/src/strategies/wse/index.ts
@@ -1,0 +1,61 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'bonustrack';
+export const version = '0.1.0';
+
+const WSE_ADDRESS = {
+  '1': '0x77b8ae2E83c7d044d159878445841E2A9777Af38'
+};
+
+const abi = [
+  {
+    constant: true,
+    inputs: [
+      {
+        internalType: 'address',
+        name: 'account',
+        type: 'address'
+      }
+    ],
+    name: 'getCurrentVotes',
+    outputs: [
+      {
+        internalType: 'uint96',
+        name: '',
+        type: 'uint96'
+      }
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
+
+export async function strategy(
+  space,
+  network,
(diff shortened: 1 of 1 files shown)


bonustrack:
@Norfolks If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call
Here is an example of a space that is using "contract-call" strategy to get voting score https://hub.snapshot.page/api/spaces/badgerdao.eth

Norfolks:
@bonustrack
Looks like the example you provided should fit our requirements.
I'll check it.

Norfolks:
@bonustrack 
I am not sure we can use the method you provided
We have a uni's fork way of government calculation.
So, we need some fork of their here too.
https://github.com/WhiteSwap/snapshot.js/blob/master/src/strategies/uni/index.ts

Can we specify block we take votes from using this method, including the "latest" option?

bonustrack:
The UNI strategy could also be replaced by to "contract-call" strategy. It's much better if you use contract-call instead of making a strategy. The snapshot block is specified when you make a proposal, then in the strategy it's changed to `blockTag`, you don't need to change that.

Norfolks:
@bonustrack 
I've made a contract-call strategy with those parameters
https://hub.snapshot.page/api/spaces/wse.eth
But it shows that a valid address, that has tokens doesn't have any voting power.
What could be wrong?

bonustrack:
@Norfolks Perhaps the valid address didn't have token at the specific snapshot block number? If you cast a vote and point me to a proposal i could do some test.

Norfolks:
@bonustrack please, it would be great.
The proposal:
https://snapshot.page/#/wse.eth/proposal/QmV1bMVM44yfbHzcu3SdQvXZje5rFW7tiEexiZNmHUzdfd
Yes vote was made from the address: 
0x77b8ae2E83c7d044d159878445841E2A9777Af38

bonustrack:
@Norfolks Are you sure that's a correct address? On Etherscan it also return 0 for that address. 
image

Norfolks:
@bonustrack 
Seems I found an issue.
Thanks

PR #71, opened 2020-12-31
Fix for import Snapshot.js give empty object

### LINKED ISSUE #70: Import Snapshot.js give empty object
The package not found issue is definitely fixed, however in another rollup issue it doesn't seem to export anything?

e.g. 

```

import snapshot from "@snapshot-labs/snapshot.js";

function App() {
  console.log(snapshot) // returns {}
  console.log(snapshot.strategies) // expected: { balancer: {}, etc... }; actual: returns undefined
}
```

I also tried to `import snapshot from @snapshot-labs/snapshot.js/dist/index` and other variations but says `Module not found`

_Originally posted by @yjkimjunior in https://github.com/snapshot-labs/snapshot.js/issues/67#issuecomment-752095893_

ChaituVR:
Fixes #70

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -16,11 +16,10 @@ export default [
   {
     input,
     context: 'window',
-    output: {
-      name,
-      file: pkg.browser,
-      format: 'iife'
-    },
+    output: [
+      { name, file: 'dist/snapshot.min.js', format: 'iife' },
+      { name, file: pkg.browser, format: 'amd' }
+    ],
     plugins: [
       json(),
       builtins(),
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -17,8 +17,7 @@ export default [
     input,
     context: 'window',
     output: [
-      { name, file: 'dist/snapshot.min.js', format: 'iife' },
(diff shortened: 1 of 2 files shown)


rollup.config.js
```diff
@@ -16,11 +16,10 @@ export default [
...
     input,
     context: 'window',
-    output: {
-      name,
-      file: pkg.browser,
-      format: 'iife'
-    },
+    output: [
+      { name, file: 'dist/snapshot.min.js', format: 'iife' },
+      { name, file: pkg.browser, format: 'amd' }
```
bonustrack, line 21: These examples suggest we should use UMD
https://github.com/rollup/rollup-starter-lib/blob/master/rollup.config.js
https://medium.com/@adostes/authoring-a-javascript-library-that-works-everywhere-using-rollup-f1b4b527b2a9
Can you try if it work? Also we can remove the IIEF build and use min.js for the UMD version

ChaituVR: Replaced iife and amd with umd 👍

bonustrack (approved):
Looks good

PR #78, opened 2021-01-13
added stablexswap strategy

howydev:
- adding score strategy for the StableXSwap protocol


src/strategies/index.ts
```diff
@@ -64,6 +65,8 @@ export default {
   defidollar,
   aavegotchi,
   mithcash,
+  'balancer-unipool': balancerUnipool,
```
bonustrack, line 68:
```suggestion
```

src/strategies/stablexswap/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  // For erc20 token balances not staked in any contract
+  const score = await multicall(
```
bonustrack, line 98: There is too much multicall in your strategy, this will take too long to resolve on the browser, you should use a single call instead, i recommend you to use "Multicaller" instead of the function "multicall()`, it's the new way we use to do multicall, it's much easier to build query with it, have a look on a demo here: https://github.com/snapshot-labs/snapshot.js/blob/master/test/multicaller.ts and a strategy using it here https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/dittomoney/index.ts#L63-L82 to understand how it works.

howydev: Appreciate the time taken to review . It makes a lot of sense to batch the calls into a single multicall - I've upgraded the design to use the Multicall class instead

src/strategies/stablexswap/examples.json
```diff
@@ -0,0 +1,19 @@
+[{
+      "stakingDecimals": 18,
```
bonustrack, line 2: The format for this file is wrong can you use the same as others like this:
https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/defidollar/examples.json
https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of/examples.json
?

src/strategies/stablexswap/index.ts
bonustrack: Looks good now thanks!

src/strategies/stablexswap/examples.json
howydev: Hey @bonustrack - I didn't realize there was a standardized formatting, sorry! I've updated the formatting to match the examples specified and it should be good since calling "npm run test --strategy=stablexswap" works. On a side note when I run npm test locally i have to add an package to use Object.fromEntries. Should I add this package as a dev dependency?

bonustrack: Not sure how you get this error, Object.fromEntries should work with dependency

bonustrack (approved):
Looks good!

PR #80, opened 2021-01-14
Update cream strategy

jeremyHD:
- Removes Cream swap score
- Add Sushiswap LP score
- Add CREAM supplied in Cream Lending
- Implement time weighted average balance


bonustrack (approved):
Works well!

PR #83, opened 2021-01-16
Add Staked KEEP

corollari:
As the title says, this PR just adds a strategy that counts the number of KEEP tokens staked by an address.

I forked it from the `uniswap` strategy and tested it with `npm run test --strategy=staked-keep`.

### DIFF
--- a/src/strategies/staked-keep/index.ts
+++ b/src/strategies/staked-keep/index.ts
@@ -1,0 +1,45 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const UNISWAP_SUBGRAPH_URL = {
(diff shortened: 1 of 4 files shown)


src/strategies/staked-keep/index.ts
```diff
@@ -0,0 +1,45 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const UNISWAP_SUBGRAPH_URL = {
```
knarz, line 4: Maybe rename to `KEEP_SUBGRAPH_URL` :)

corollari: Will do, that's an artifact of the code I cloned

bonustrack (approved):
Looks great!

eightf0ld:
Sweet!

PR #89, opened 2021-01-29
Add balance-with-min strategy

samuveth:
- If balance is below minimum the voting power is replaces with zero
-

### DIFF
--- a/src/strategies/erc20-balance-of-with-min/examples.json
+++ b/src/strategies/erc20-balance-of-with-min/examples.json
@@ -1,0 +1,56 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "erc20-balance-of-with-min",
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
Works well!

samuveth:
@bonustrack How long does it usually take until it's live after merge.

samuveth:
Could the this strategy also notify the voter if the balance is below minimum?

bonustrack:
@samuveth When there is a new strategy, the frontend need to be redeployed, i'm doing this now it should be live in 5mn. When user will vote and have less than the minimum he will see 0 as voting power, so somehow he will get notified.

sinasabet81:
Does this affect both voting and proposing? Would it be possible to support a similar strategy that only blocks proposing for users below the threshold?

PR #90, opened 2021-02-01
Add Opium strategy

Alirun:
- $OPIUM balance
- $wOPIUM balance
- 1inch LP OPIUM-ETH balance (adjusted to $OPIUM held in LP tokens)
- Farming 1inch LP OPIUM-ETH balance (adjusted to $OPIUM held in LP tokens)


bonustrack (approved):
Works well!

PR #91, opened 2021-02-01
add ticket validity strategy

ethedev:


src/strategies/ticket-validity/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  const score = await erc20BalanceOfStrategy(
+    space,
+    network,
+    provider,
+    addresses,
+    options,
+    snapshot
+  );
+  return Object.fromEntries(
+    Object.entries(score).map((address: any) => [address[0], (address[1] > 0 ? 1 : 0)])
```
bonustrack, line 23: It's better if the minimum is a parameter like this
```suggestion
    Object.entries(score).map((address: any) => [address[0], (address[1] > options.min || 0 ? 1 : 0)])
```

PR #92, opened 2021-02-02
add example strategy

ethedev:


bonustrack:
What's the need of this strategy? We can use `erc20-balance-of` as example strategy

ethedev:
> What's the need of this strategy? We can use `erc20-balance-of` as example strategy

Yes we can, but i think more things will come on the top of that in the future, as an "example"
And we will have it here https://docs.snapshot.page/strategies/create

PR #96, opened 2021-02-07
Add Sushi Masterchef LP token strategy

0xKiwi:
This PR adds a strategy that allows users with their LP tokens staked in Sushi's masterchef to vote in snapshot proposals.

Note: this has not been tested and I'm not entirely familiar with The Graph and its inner workings. Please review and confirm it works however possible.

If there are any changes that need to be made, let me know! Thank you!

### DIFF
--- a/src/strategies/masterchef/index.ts
+++ b/src/strategies/masterchef/index.ts
@@ -1,0 +1,102 @@
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+
+const MASTERCHEF_SUBGRAPH_URL = {
+  '1': 'https://api.thegraph.com/subgraphs/name/sushiswap/master-chef'
+};
+
+const SUSHISWAP_SUBGRAPH_URL = {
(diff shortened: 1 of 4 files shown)


src/strategies/masterchef/index.ts
```diff
@@ -0,0 +1,102 @@
...
+};
+
+export const author = '0xKiwi';
+export const version = '0.1.0';
+
+export async function strategy(
+  _space,
+  network,
+  _provider,
+  address,
```
bonustrack, line 19: @0xKiwi This should not be `address` but `addresses` it's an array of voter addresses. The strategy must calculate score for each of these addresses not just a single one, the strategy should be refactored to reflect this now the code is made assuming this is just a single address.

0xKiwi:
Fixed! @bonustrack :smile: 

It assumes theres only 1 pool in the masterchef contract, which should work for most situations. If someone ever needs better support ping me!

bonustrack (approved):
Looks good now!

PR #97, opened 2021-02-09
Add Frax Finance Strategy to account for FXS in LP tokens

FortisFortuna:
This allows Frax Finance to calculate FXS balances from free FXS tokens, as well LP tokens, both free and staked. It gives more weight to LP tokens that are staked for longer periods of time.


bonustrack (approved):
Looking good!

FortisFortuna:
Thank you! 🙏🙏🙏

PR #103, opened 2021-02-17
Split the single strategy into 3 strategies for the-graph

davekaj:
Fixes a few bugs that were in the original strategy - wrote network (was set to 2 instead of 4) and misnamed params

Also splits the single strategy into three strategies. 

I also had a question for the maintainers:
- I have a high level folder `the-graph` that has common funcs used in my three strategies. The reason I did this, was because I was unsure if every strategy had to be it's own folder. If each strategy does not need to have it's own folder, I will combine all of these 4 folders into a single folder. Let me know when you have a chance, thank you 🙏


bonustrack (approved):
@davekaj It's fine having the folder "the-graph" for common functions. Sorry for late review, the strategies seem to work well for me.

PR #107, opened 2021-02-23
Robust grt testing

davekaj:
This adds testing for the strategies into the repo.

I delegated and indexed specific amounts in the network, so I knew what the results would have to be. Then added into the `example.json`s the correct values to expect. Caught a few bugs along the way, but now everything checks out.

### DIFF
--- a/src/strategies/the-graph-balance/balances.ts
+++ b/src/strategies/the-graph-balance/balances.ts
@@ -3,8 +3,8 @@ import { subgraphRequest } from '../../utils';
 import {
   GRAPH_NETWORK_SUBGRAPH_URL,
   GraphAccountScores,
-  WEI
-} from '../the-graph/utils';
+  bnWEI
+} from '../the-graph/graphUtils';
 
 export async function balanceStrategy(
   _space,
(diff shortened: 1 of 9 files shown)


bonustrack:
@davekaj Don't you want make the testing optional? If i deploy these changes i assume the strategy will fail because there isn't any expectedResults setup on the space here https://hub.snapshot.page/api/spaces/thegraphnetwork.eth

davekaj:
Ah, I see I for sure want to make the testing optional, I don't want to require this to run in the front end. 

Maybe I misunderstood how `examples.json` works. I thought it was strictly a test file that was used to test locally, and that it would no influence how the hub reads a strategy. I guess the hub uses this file as well?

Anyhow, if you could fill me in on how to make the testing optional, I could fix the PR and resubmit. Please let me know! And thanks for catching this :D

bonustrack:
@davekaj The examples.json file is just used locally not on Snapshot UI, it's like a mock of data. The data used on Snapshot UI come from the space settings and the proposal, votes. 

To make testing optional you could do something like this before calling `verifyResults` :

```js
if (options.expectedResults) {
  verifyResults(...);
}
```

davekaj:
Ah yeah, this is obvious now that you put it like this 🤣 . Easy fix , will push it today

davekaj:
Fixed, let me know if it looks o.k.

bonustrack (approved):
Works well!

PR #108, opened 2021-02-23
DAO module support

rmeissner:
This adds support for the [DAO module](https://github.com/gnosis/dao-module) to snaphot-js.

The plugin defined is required to support the DAO module in the snapshot interface.

TODOs:
- [x] Set proper plugin info (author, icon, name, version) 
- [x] Remove `console.log`s
- [x] Update abi to latest version after audit (e.g. `executedPropsalTransactions` was corrected`)


yarn.lock
```diff
@@ -0,0 +1,4039 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
bonustrack, line 1: This repo don't use Yarn but NPM, can you remove this file?

src/plugins/daoModule/index.ts
```diff
@@ -0,0 +1,214 @@
+import { getAddress, isAddress } from '@ethersproject/address';
+import { BigNumber, isBigNumberish } from '@ethersproject/bignumber/lib/bignumber';
+import { isHexString } from '@ethersproject/bytes';
+import { Contract } from '@ethersproject/contracts';
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { keccak256 as solidityKeccak256 } from '@ethersproject/solidity';
```
bonustrack, line 6: It would be better to import `keccak256` from the source like this:
```suggestion
import { keccak256 as solidityKeccak256 } from '@ethersproject/keccak256';
```

src/plugins/daoModule/index.ts
```diff
@@ -0,0 +1,214 @@
...
+const buildQuestion = async (proposalId: string, txHashes: string[]) => {
+  const hashesHash = solidityKeccak256(["bytes32[]"], [txHashes]).slice(2);
+  return `${proposalId}␟${hashesHash}`
+}
+
+const getNextTxIndex = async (provider: JsonRpcProvider, moduleAddress: string, proposalId: string, txHashes: string[]) => {
+  // Call syntax is [to, [params]]
+  const question = await buildQuestion(proposalId, txHashes)
+  const questionHash = solidityKeccak256(["string"], [question])
+  const executionStati = (await multicall(provider.network.chainId.toString(), provider, ModuleAbi,
```
bonustrack, line 90: Some network keys are not the same as their chainId, it would be better to pass the network string in the function param directly (same goes for others functions in the plugin).

```suggestion
  const executionStati = (await multicall(network, provider, ModuleAbi,
```

bonustrack (commented):
It looks great! I've added few comments, also can you run the "lint" script? That will fix some of code format.

src/plugins/daoModule/index.ts
rmeissner: they are slightly different. The solitidy version performs encoding on top of the hashing, that is especially important in the case where the hash of an array of bytes32 is generated.

bonustrack: Ho right, good catch!

PR #111, opened 2021-02-26
Add tracer-dao snapshot strategy

raymogg:
- Adds a strategy for the tracer-dao

### DIFF
--- a/src/strategies/index.ts
+++ b/src/strategies/index.ts
@@ -49,6 +49,7 @@ import { strategy as theGraphBalance } from './the-graph-balance';
 import { strategy as theGraphDelegation } from './the-graph-delegation';
 import { strategy as theGraphIndexing } from './the-graph-indexing';
 import { strategy as whitelist } from './whitelist';
+import { strategy as tracerDAO } from './tracer-dao';
 
 export default {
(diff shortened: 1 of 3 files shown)


samuveth (approved):
Looks good!

bonustrack:
@raymogg If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call
Here is an example of a space that is using "contract-call" strategy to get voting score https://hub.snapshot.page/api/spaces/badgerdao.eth Could use that instead?

raymogg:
> @raymogg If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call
> Here is an example of a space that is using "contract-call" strategy to get voting score https://hub.snapshot.page/api/spaces/badgerdao.eth Could use that instead?

Looks like this will work for our use case, thanks @bonustrack. Closing the PR.

PR #113, opened 2021-02-27
added strategy to give liquidity providers on the Ocean marketplace v…

w1kke:
…oting power

Fixes # .

- Add a strategy for Ocean Protocol to empower the liquidity providers on their marketplace to get voting power
- Added the strategy to index.ts
- Added a README.md that outlines what is being done in the strategy and possible ways to improve it in the future as well how to extend and test it


bonustrack:
@w1kke Hi! You must define the block number in your strategy like this:
https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/balancer/index.ts#L44-L47
Otherwise it will only return the data in live, regardless of the proposal snapshot.

w1kke:
It queries this GraphQL subgraph:
https://subgraph.mainnet.oceanprotocol.com/subgraphs/name/oceanprotocol/ocean-subgraph

And if I add the block parameter to the query it will not give any return values to the query.

w1kke:
@bonustrack I added the parameter - but I have to cast to an int to comply with the subgraph API - so it is +snapshot

src/strategies/index.ts
```diff
@@ -100,6 +101,7 @@ export default {
   work,
   'ticket-validity': ticketValidity,
   opium,
+  'ocean-marketplace': ocean
```
bonustrack, line 104:
```suggestion
  'ocean-marketplace': ocean,
```

bonustrack (approved):
Looks good!

PR #116, opened 2021-03-02
Fix the format of the returned score values

w1kke:
Fixes # .

- The voting works correctly with the last version of the strategy
- But the display of the voting power is not correct
- This has been corrected from my understanding but now it has to be proven in a test vote


bonustrack:
@w1kke Can you resolve the conflicts?

w1kke:
@bonustrack I resolved them

bonustrack (approved):
Looks good! Thanks and sorry for the delay

bonustrack:
The change on the strategy is live.

PR #121, opened 2021-03-08
Add Total Axion Strategy

JTravaux:
- Added a strategy to get the total [Axion](https://axion.network/) shares of an address.
- Amount can be verified from [this UI](https://axion.today/) after entering the address to check if necessary.


samuveth (approved):
Looks good!

```
Strategy: "total-axion"
Total Axion (Staked + Liquid)
[
  { '0xe8b283b606a212d82036f74f88177375125440f6': 29938042949.291603 }
]
getScores: 711.55ms
```

src/strategies/total-axion/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  return Object.fromEntries(
+    await Promise.all(addresses.map(
+      async (address: string) => [
+        address, 
+        await getAddressTotalAxion(address)
```
bonustrack, line 24: @JTravaux This will not scale well, if there is 1,000 voters there will be 1,000 requests. Is there a way you can get the amounts for multiple address in a single request?

src/strategies/total-axion/index.ts
```diff
@@ -0,0 +1,28 @@
+import fetch from 'cross-fetch';
+
+export const author = 'Jordan Travaux';
+export const version = '0.1.0';
+
+const getAddressTotalAxion = async (address: string) => {
+  const rawResponse = await fetch(`https://gateway.axntoday.com/v1/staker/${address}`);
```
bonustrack, line 7: One thing i see missing is the "snapshot", if you don't have a way to get staker amount at a specific block number the results of the proposals will change in live and never be fixed, this can be problematic.

src/strategies/total-axion/index.ts
JTravaux: I see. Looking into this!

src/strategies/total-axion/index.ts
JTravaux: Ah, good point. Will implement a fix!

JTravaux:
Need to change 1 thing.

JTravaux:
image

bonustrack (approved):
Looks good! Sorry for late review

PR #124, opened 2021-03-10
Add Avalanche mainnet configuration

mrFawkes:
- add avalanche network mainnet configuration in networks.json


bonustrack:
@mrFawkes Thanks for your PR! Something is missing, you need to have a multicall contract and reference it on the src/utils.ts file here https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L19-L37

mrFawkes:
> @mrFawkes Thanks for your PR! Something is missing, you need to have a multicall contract and reference it on the src/utils.ts file here https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L19-L37

Thank you for the heads up @bonustrack - multicall contract reference has been added to utils.ts

bonustrack (approved):

PR #125, opened 2021-03-10
Chunk strategy calls

bonustrack:
This change paginate strategy calls with 250 items. If there is 600 voters on a proposal, there will be 3 calls instead of one. This should fix the issue on this proposal https://snapshot.page/#/badgerdao.eth/proposal/QmW3xiKANJ7ed5vjgpDGdhWN2ri4g8CHPn5VxAe4WZUfLL (haven't tested yet)


ChaituVR:
@bonustrack proposals are working  even without this change currently, if it is useful in the future we can merge it

bonustrack:
Just realized this change may not work well with delegation. Delegation must be aware of the full list of voters.

PR #126, opened 2021-03-12
ADD POB hash strategy

dave4506:
Gets ERC1155 balances via the graph subgraph and apply a `Math.sqrt`

One thing: 
there is a limit via the graph with `first:1000` limiting to 1000 addresses, is there a workaround?


src/strategies/pob-hash-cv/index.ts
```diff
@@ -0,0 +1,55 @@
+import { subgraphRequest } from '../../utils';
+
+export const author = 'bonustrack';
```
bonustrack, line 3: This should be your GitHub username
```suggestion
export const author = 'dave4506';
```

src/strategies/pob-hash-cv/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  options,
+  snapshot
+) {
+  const hashOwnersParams = {
+    hashOwners: {
+      __args: {
+        where: {
+          id_in: addresses.map(a => a.toLowerCase()),
+        },
+        first: 1000, // IS THIS ENOUGH?
```
bonustrack, line 25: This is enough if there is less than 1000 voters (that's often the case), you would need to paginate request to support more than 1000 voters using `first: 1000, skip: 1000`

dave4506:
@bonustrack can I get another review? thanks!

bonustrack (approved):
Looks good, and fast! Can you just fix the conflict?

dave4506:
Conflict fixed

PR #132, opened 2021-03-16
Aavegotchi strat update for matic

vfatouros:
Aavegotchi has moved to Matic and has some changes to their staking contract, the strategy needed updating to work for this


bonustrack (approved):
Looks good!

PR #133, opened 2021-03-16
Add the Quorum plugin

lbeder:
In the PR, we're introducing the logic of the Quorum plug-in which can be used to present the quorum on an ongoing proposal. 

Currently, the plug-in supports the `balance` strategy which signals that the quorum result should be normalized by the specified `decimals` attribute. Please see `examples.json` from a sample config.

For example:

photo_2021-03-16 22 02 59


bonustrack (approved):
This is awesome! I really like that you make it generic so others can use it too :))

lbeder:
It was my pleasure to contribute 😌

PR #134, opened 2021-03-17
Refactor synthetix strategy to use subgraph

andytcf:
Fixes # .

- 
- 
-


src/strategies/synthetix/index.ts
```diff
@@ -1,62 +1,53 @@
...
+    // @ts-ignore
+    params.snxholders.__args.block = { number: snapshot };
+  }
+
+  const result = await subgraphRequest(SYNTHETIX_SUBGRAPH_URL, params);
+  const score = {};
+
+  if (result && result.snxholders) {
+    result.snxholders.forEach((holder) => {
+      score[holder.id] = quadraticWeighting(
```
bonustrack, line 46: Can you checksum the voter address here? You can use `getAddress(holder.id)` from ethers.js

andytcf: fixed!

PR #135, opened 2021-03-18
Add sportx-strategy

mrwillis:
Fixes # .

- 
- 
-


bonustrack:
@mrwillis If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call see example https://hub.snapshot.page/api/spaces/nice

mrwillis:
Ah got it thanks. Will try that.

bonustrack:
Closing this PR for now, feel free to open it again if you still want go forward with it.

PR #136, opened 2021-03-18
Add ruler-staked-lp strategy

vfatouros:
Ruler and Cover have a unique staking contract so the staked-uniswap strategy did not apply, this is slightly modified to work for those contracts.


src/strategies/ruler-staked-lp/index.ts
```diff
@@ -0,0 +1,116 @@
...
+        options.stakingAddress,
+        'getUser',
+        [options.sushiPoolAddress, address]
+      ])
+    ),
+    { blockTag }
+  );
+
+  const totalSupply = res[0];
+  const tokenBalanceInUni = res[1];
```
crypto-pumpkin, line 104:
```suggestion
  const tokenBalanceInPool = res[1];
```

Same for the var name below.

bonustrack (approved):
Looks good!

PR #138, opened 2021-03-19
Update "xdai-stake" strategy and add "xdai-posdao-staking" strategy

maxaleks:
- remove the usage of "erc20-balance-of" strategy from "xdai-stake" strategy because we can set both strategies for our space and the result will be the same.
- rename "xdai-stake" strategy to "xdai-easy-staking"
- add "xdai-posdao-staking" strategy


bonustrack (approved):
Looks good!

PR #142, opened 2021-03-23
Add ren-nodes strategy

noiach:
Thanks for creating/mainting Snapshot. We are hoping to set up a [Ren](https://renproject.io) snapshot (https://snapshot.org/#/ren-project.eth) using this strategy.

- Add ren-nodes strategy


bonustrack (approved):
Looks good! Can you just fix the conflicts and remove the change on `src/plugins/daoModule/index.ts`?

noiach:
Ah, didn't mean to commit that, it got mixed in with a merge. However I did need that change to be able to run tests locally, to avoid the error `semantic error TS2783: 'nonce' is specified more than once, so this usage will be overwritten`. It might be an issue with my local set-up though.

bonustrack:
@noiach Thanks looks good now, we will get the strategy deployed shortly

MaximilianR:
> @noiach Thanks looks good now, we will get the strategy deployed shortly

Hey just wondering what the approximate timeline is to get that added? 🤓 We have excited community members ready to start voting on proposals, just want to know roughly if that is something that can be added this week and we should prepare already, or if it's a next week kind of thing!

PR #149, opened 2021-03-25
Implement GRT's expectedResults to validate Ocean-Marketplace strategy.

idiom-bytes:
Changes proposed in this PR:
- Implement GRT's expectedResults model into ocean-marketplace strategy.
- Validate expectedResults to show that strategy is returning correct scores for blockHeight: 12097120 

Notice the following proposal:
- Implements this strategy.
- Has OceanMarketplace LP vote.
- Does not acquire the expected score for the target Address (0x655efe6eb2021b8cefe22794d90293aec37bb325)

[Proposal Link](https://snapshot.org/#/officialoceandao.eth/proposal/QmSbhBKUnJ6tPeqoHZB5aS7z72JfLAKCu5W6TndHSFqnmC)

[Expected Address 0x655efe6eb2021b8cefe22794d90293aec37bb325](https://etherscan.io/address/0x655efe6eb2021b8cefe22794d90293aec37bb325)

[Ocean Marketplace LP Holdings](https://market.oceanprotocol.com/asset/did:op:7Bce67697eD2858d0683c631DdE7Af823b7eea38)

[Expected Votes at blockHeight 12097120 => 234872.06400571144]
(https://github.com/idiom-bytes/snapshot.js/commit/320e6ca5c4a6776ea3f637f8a14649530e230df2)


bonustrack:
@idiom-bytes The issue is that the address are lowercased, the strategy should return only checksumed addresses, you can use the function "getAddress" from Ethers.js to do that, example: https://github.com/snapshot-labs/snapshot.js/blob/a17ac1f522693facd18443271b384921b5a61389/src/strategies/synthetix/index.ts#L47

idiom-bytes:
Huge, thank you @bonustrack, I think this is all setup now!

src/strategies/ocean-marketplace/examples.json
```diff
@@ -1,22 +1,41 @@
...
+      "0x5e22b34F2708EA7e25918B8FCF816F9CfF27a331",
+      "0x5266dc406d688bcb152bfCf2C412113b821c5A43",
+      "0x620bDD183Ec570E4de54001A6e3CbF4215d6d326",
+      "0xB9b14C4d0cbc18FC5f54701D578e85968cf19FD5",
       "0x12BD31628075C20919BA838b89F414241b8c4869",
       "0xf88bD9c3F81f7148E4076bBB8b2e2B0951A8cE38",
       "0xAAB9EaBa1AA2653c1Dda9846334700b9F5e14E44"
     ],
-    "snapshot": "11635000"
+    "snapshot": "12097120"
```
bonustrack, line 39: This should be a number
```suggestion
    "snapshot": 12097120
```

src/strategies/ocean-marketplace/index.ts
```diff
@@ -89,26 +84,42 @@ export async function strategy(
...
+    });
+
+    verifyResults(
+      JSON.stringify(results),
+      JSON.stringify(options.expectedResults.scores),
+      'Scores'
+    );
+  }
+
   return return_score || {};
```
bonustrack, line 124: It seem the strategy return all the token holders when i test, something is wrong, you should only return score from voters, i think you need to change here
```suggestion
  return results || {};
```

idiom-bytes: The strategy looks inside tokens inside the 1000 largest pools, and everything was being returned.

I have updated the strategy() function to only return the scores from the addresses provided/expected.

idiom-bytes:
Hi @bonustrack can I get a review? Thank you!

bonustrack (approved):
Looks good now!

PR #152, opened 2021-03-26
Add a strategy to record 1 vote for a multisig owner/member

davekaj:
- Add a strategy to record a single vote as an owner in a multisig wallet

I tried to implement it using `contract-call` but it didn't work because `isOwner()` in the gnosis multisig returns `true`, and not a value. 

I kind of have a feeling some one else might have made a similar strategy to this one. If so , please point me in that direction, and I will just reuse that strategy, and we can close this PR. 

Thank you!


bonustrack (approved):
Looks great! We dont have this specific strategy yet no.

PR #156, opened 2021-03-28
Add support for ERC1155 all-balances-of

fragosti:
Adds a new strategy that returns the total balances you have for a specific ERC1155 address using this subgraph https://thegraph.com/explorer/subgraph/alexvorobiov/eip1155subgraph

So this is your total balance across all IDs in the ERC1155.

Ran the linter. Really only adding `erc1155-all-balances-of` strategy


src/strategies/erc1155-all-balances-of/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export async function strategy(
+  _space,
+  network,
+  _provider,
+  addresses,
+  options,
+  _snapshot
+) {
+  const eip1155OwnersParams = {
```
bonustrack, line 18: The subgraph query should include the snapshot block number too, here is an example https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/ens-domains-owned/index.ts#L45-L48

src/strategies/erc1155-all-balances-of/examples.json
```diff
@@ -0,0 +1,19 @@
...
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0x4baf2ad9e92d547f428895e12ab8a6f195c6f058",
+      "0x34a745008a643eebc58920eaa29fb1165b4a288e",
+      "0xcc5Ddc8CCD5B1E90Bc42F998ec864Ead0090A12B",
+      "0x000000000000000000000000000000000000dead"
+    ],
+    "snapshot": "latest"
```
bonustrack, line 17: On the example file the snapshot should be a number, not latest

src/strategies/erc1155-all-balances-of/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      SUBGRAPH_URL[network],
+      eip1155OwnersParams
+    );
+    return result.accounts.reduce((acc, val) => {
+      const relevantTokenBalances = val.balances.filter((balance) => {
+        const isRightAddress =
+          balance.token.registry.id === options.address.toLowerCase();
+        return isRightAddress;
+      });
+      acc[val.id] = relevantTokenBalances.reduce(
```
bonustrack, line 51: The voters addresses must be checksumed (not lowercase), you can use the function "getAddress(address)" from ethers.js to do that example: https://github.com/snapshot-labs/snapshot.js/blob/aa5804cbd9f8745b3c227f6be1786a418142f8ab/src/strategies/staked-keep/index.ts#L39

bonustrack (approved):
Looks good! Can you just fix the conflict or authorize maintainers to edit your PR so we can do it for you?

image

PR #158, opened 2021-03-31
Fix "xdai-posdao-staking" strategy and add "xdai-stake-holders" strategy

maxaleks:
Fixes wrong block number in `xdai-posdao-staking` strategy because this strategy uses data from xDai chain but the snapshot is from Ethereum Mainnet.

- add `xdai-stake-holders` strategy

There are 2 files of `ocean-marketplace` strategy because they didn't run `lint`


bonustrack (approved):
Looks good!

PR #162, opened 2021-04-02
Saffron Finance Strategy

turpintinz:
-  Add saffron-finance strategy
-  Sum up balances from several LP Tokens
-  Apply multiplier to score as configured


src/strategies/saffron-finance/index.ts
```diff
@@ -0,0 +1,268 @@
...
+        boostResponses: new Array(),
+        votingScores: new Array()
+      }
+      let queries = addresses.map((address: any) => [
+        token.tokenAddress,
+        "balanceOf",
+        [address]
+      ]);
+      tokenRecord.queries.push(queries);
+      let response = await multicall(network, provider, abi, queries, {blockTag});
```
bonustrack, line 171: The way the strategy is designed is not going to scale well, the result of a strategy is calculated on user browser i recommend to send a maximum of 3 multicall request in a single strategy but this strategy seem to be sending more than 30 multicall.

turpintinz: Okay, thank you. I can create an implementation with that restriction.

src/strategies/saffron-finance/index.ts
```diff
@@ -0,0 +1,273 @@
...
+    });
+    votingScores.push(...batchScores);
+  });
+
+  // ================ Sum up everything =================
+  let addressVotingScore = addresses.map((address: any, index: number) => {
+    let total = BigNumber.from(0);
+    holdersQueryBatches.forEach((batch: Batch, index: number) => {
+      let votingScore = votingScores[batch.qIdxStart+index];
+      if (votingScore.address === address) {
```
bonustrack, line 254: @turpintinz Just tried the strategy, it fail for me at this line with `TypeError: Cannot read property 'address' of undefined`.

turpintinz: Yep, me too. I'll be repairing that today, April 9. Thanks for asking.

turpintinz: > @turpintinz Just tried the strategy, it fail for me at this line with `TypeError: Cannot read property 'address' of undefined`.

This is now repaired.

bonustrack (approved):
Works well now

PR #165, opened 2021-04-06
Change Total Axion Shares Strategy

JTravaux:
Change the total-axion-shares strategy to use the number of days in the params to exclude shares under a certain length.


bonustrack (approved):
Looks good!

PR #168, opened 2021-04-08
fix: Allow dashes in social usernames

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/293

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -33,13 +33,13 @@
         "twitter": {
           "type": "string",
           "title": "twitter",
-          "pattern": "^[A-Za-z0-9_]*$",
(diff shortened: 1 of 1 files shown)


src/schemas/space.json
```diff
@@ -33,13 +33,13 @@
         "twitter": {
           "type": "string",
           "title": "twitter",
-          "pattern": "^[A-Za-z0-9_]*$",
+          "pattern": "^[A-Za-z0-9_-]*$",
```
bonustrack, line 36: Twitter dont allow dash
```suggestion
          "pattern": "^[A-Za-z0-9_]*$",
```

bonustrack (approved):

PR #171, opened 2021-04-11
Refactor Total Axion Shares Strategy

JTravaux:
Refactor strategy to reduce chances of hitting rate limits. Now moved all checks to 1 multicall.


bonustrack (approved):
Works well now

PR #174, opened 2021-04-15
Tranche Staking LP strategy

ayush-jibrel:
Fixes # .

- 
- 
-


src/strategies/tranche-staking-lp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({ query: queryParam })
+  });
+
+  let { data } = await result.json();
+
+  if (data && data.pairs) {
+      data.pairs.map(async (object) => {
```
bonustrack, line 72: Can you remove `async` here?
```suggestion
      data.pairs.map((object) => {
```

src/strategies/tranche-staking-lp/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const lpTokenAddress = options.lpTokenAddress.toLowerCase();
+
+  let queryParam = `query {pairs (where: { id: "${lpTokenAddress}" }) {id totalSupply token0{id} token1{id} reserve0 reserve1 } }`;
+
+  const tokenAddress = options.tokenAddress.toLowerCase();
+
+  var rate;
+  
+  const result = await fetch(UNISWAP_SUBGRAPH_URL[network], {
```
bonustrack, line 60: Can you use the method `subgraphRequest` instead of fetch here? Here is an example: https://github.com/snapshot-labs/snapshot.js/blob/a0adc547aa0922aa6abd35708a4a292048bca6a2/src/strategies/staked-keep/index.ts#L18-L35

ChaituVR (commented):
LGTM 👍

PR #175, opened 2021-04-16
Change graph

novaknole:
Changing back the subgraph url to the original one that Aragon is going to use for the rinkeby environment.


bonustrack (approved):
Thanks!

PR #177, opened 2021-04-17
Create index.ts

75900:
Added erc20-shareTokenBalance strategy to use shareTokenBalance function instead of balanceof, please review.


bonustrack:
@75900 If you just need to call a specific function of a smart contract you dont need to create your own strategy you can use the strategy "contract-all" which allow you to do a custom call see: https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/contract-call see example https://hub.snapshot.page/api/spaces/nice

PR #179, opened 2021-04-21
Added masterchef-pool-balance strategy

bowsernet:
Generic strategy for voting with tokens staked in a masterchef type contract. Happy to discuss the details if not clear from the source code.

### DIFF
--- a/src/strategies/masterchef-pool-balance/examples.json
+++ b/src/strategies/masterchef-pool-balance/examples.json
@@ -1,0 +1,38 @@
+[
+  {
+    "name": "Example query",
+    "strategy": {
+      "name": "masterchef-pool-balance",
+      "params": {
+        "chefAddress": "0xD38abbAeC03a9FF287eFc9a5F0d0580E07335D1D",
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
LGTM @bonustrack

src/strategies/masterchef-pool-balance/examples.json
```diff
@@ -0,0 +1,38 @@
...
+        "pid": "0",
+        "weight": 1
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0xfCA5a27d4cfF104FC276897CA3f32cFeDc6f50BA",
+      "0x577bfa0898187c10bbbbb3d001c94aafb5cfc0e4"
+    ],
+    "snapshot": "latest"
```
bonustrack, line 18: This should use a specific block number. The result should always be the same, but using latest it may change
```suggestion
    "snapshot": 12288851
```

bowsernet: Changed, thanks

PR #180, opened 2021-04-22
Api strategy

ganzai-san:
Fixes # .

After reviewing someone's conversation in the Snapshot strategies discord, and seeing Sohkai's response, I decided to make a strategy that outsources the scoring to a backend. https://discord.com/channels/707079246388133940/759631271998652466/829043092500906064
I understand this sacrifices the verifiability of score calculation - the tradeoff is that the backend service can more flexibly handle strategies, without having to resubmit PRs and bring overhead to the Snapshot team.


bonustrack (commented):
Hey @ganzai-san thank you for your contribution, i really like that you made it generic enough so others can use it too, that's really cool! I just see 2 small things missing in the request, the snapshot and the network, this should be sent to the API so there is way manage score for specific point of time. Could you add these 2 as param? Thank you!

ganzai-san:
Hey @bonustrack 
Good call. I have added the network and snapshot to the API request.

bonustrack (approved):

PR #186, opened 2021-04-26
Cream updated

jeremyHD:
- Calculate voting power with smart contract on-chain


bonustrack (approved):
Looks good!

PR #187, opened 2021-04-27
fix: update heco mainet rpc url

crayfish007:
Fixes # .
Since HECO only support ```https://mainnet.heco.vote``` to get archived data, we have to use the vote rpc as snapshot rpc url
- update heco testnet rpc
- update heco mainnet explorer
- update heco mainnet rpc
- update heco mainnet ws
- update heco mainnet explorer


bonustrack (approved):
Looks good!

PR #188, opened 2021-04-27
Add uri format validation for new "terms" field

samuveth:


src/schemas/space.json
```diff
@@ -20,6 +20,11 @@
           "title": "about",
           "maxLength": 160
         },
+        "terms": {
+          "type": "string",
+          "title": "terms",
+          "format": "uri"
```
bonustrack, line 26: Can you add this format on "website" field too?

samuveth: `uri` will not be invalid without`*://` at the beginning. Any what these other ones mean?
- uri-reference: URI reference, including full and relative URIs.
- uri-template: URI template according to RFC6570
Source: https://github.com/ajv-validator/ajv-formats

samuveth: I don't think this will work because domain field isn't really a `uri`.

bonustrack: I mean `website` field, it's another one

samuveth: Oh, I see

bonustrack (approved):

PR #189, opened 2021-04-28
xSEEN Strategy for seen.haus

JayWelsh:

Creates strategy to derive the SEEN value from the amount of xSEEN tokens held by an address at the time of the Snapshot

The strategy will be relevant to the https://snapshot.org/#/seen workspace


ChaituVR (approved):
Lgmt @bonustrack 


@JayWelsh Please add more addresses to example.json if possible, Thanks

PR #192, opened 2021-04-30
Flashsake strategy

anassohail99:
Fixes #. There are not any fixes

- Strategy added for liquidity providers in flash protocol
- Strategy added for stakes in flash protocol
-


src/strategies/flashstake/index.ts
```diff
@@ -0,0 +1,99 @@
+import { getAddress } from '@ethersproject/address';
+import { formatUnits } from '@ethersproject/units';
+import { multicall, subgraphRequest } from '../../utils';
+
+const FLASHSTAKE_SUBGRAPH_URL = {
+  '1':
+    'https://api.thegraph.com/subgraphs/name/blockzerohello/flash-stake-stats-v2-subgraph'
+};
+
+export const author = 'BlockzeroLabs';
```
bonustrack, line 10: This must be your GitHub username
```suggestion
export const author = 'anassohail99';
```

src/strategies/flashstake/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  network,
+  _provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const params = {
+    users: {
+      __args: {
+        where: {
```
bonustrack, line 24: You may want to use `first: 1000` to get up to 1000 results

src/strategies/flashstake/index.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+      }
+    }
+  };
+
+  const params2 = {
+    stakes: {
+      __args: {
+        where: {
+          user: addresses[0].toLowerCase()
```
bonustrack, line 54: Here it seem that you are only query stake for a single address, the first address, you should instead query data for all the addresses.

bonustrack:
Also can you add map the strategy in this file: https://github.com/snapshot-labs/snapshot.js/blob/0295866d4aa4ef460a563b4127c9ce0e29620eac/src/strategies/index.ts
Once you do that you can test the strategy using `npm run test --strategy=flashstake`

PR #194, opened 2021-05-01
Add schema for admins field

samuveth:
Add schema for admins field and add string length and pattern requirements.

Related to https://github.com/snapshot-labs/snapshot/pull/339


bonustrack (approved):
Nice!

PR #196, opened 2021-05-03
fix: update heco mainet rpc url

Eason123:
Since the rpc archive node: mainnet.heco.vote has stopped providing services, it is requested to change back to the previous node.

Fixes # .

- update heco mainnet rpc
- 
-


bonustrack (approved):
Looks good!

bonustrack:
@Eason123 We are running into issue with the current Heco node. It seem that the node is not a full archive node. Are you able to assist?

PR #199, opened 2021-05-04
feat: Setup adventureland strategy

AdventurelandDeFi:
Adding adventureland strategy to call api


src/strategies/adventureland/examples.json
```diff
@@ -0,0 +1,28 @@
...
+    "network": "56",
+    "addresses": [
+      "0xa478c2975ab1ea89e8196811f51a7b7ade33eb11",
+      "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
+      "0x1AeE7eBabdAa9afe11B2abD2748a548316E6c065"
+    ],
+    "snapshot": 7120529
+  },
+  {
+    "name": "Synthetix council",
```
ChaituVR, line 16: Hi @AdventurelandDeFi can u plz remove this second if it is not required, Thanks 


LGTM @bonustrack

AdventurelandDeFi: done

src/strategies/adventureland/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  snapshot
+) {
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+
+  const promisesBalances = addresses.map((address) =>
+    fetch(
+      `${apiUrl}/api/governance/vote/${address}/${blockTag}`
+    ).then((response) => response.json())
+  );
+  const balances: ApiReturn[] = await Promise.all(promisesBalances);
```
bonustrack, line 25: Hi, sorry i've merged your PR but then reverted it. The reason is that this line will not scale well, if there is 1000 votes there will be 1000 requests being sent, the strategy are process on browser so this wont work well. I recommend to use the existing strategy called "api" this strategy will take multiple address into account. Have a look how is work here: https://github.com/snapshot-labs/snapshot.js/pull/180/files

AdventurelandDeFi: Oh, didn't even look into that strategy. I have got that setup. Thanks for the suggestion!

PR #201, opened 2021-05-06
Cross-chain HOPR token strategy

QYuQianchen:
This strategy calculates the voting weights by summing up the following balance at a given snapshot: 
- HOPR token (ERC-777) on Ethereum Mainnet with `erc20-balance-of`
- Bridged HOPR, aka "xHOPR"(ERC-677), on xDAI chain
- Wrapped xHOPR, aka "wxHOPR" (ERC-777), on xDAI chain
- Nominal HOPR token from circulating Uniswap V2 HOPR-DAI liquidity provider token. This allows LPs to vote without removing their liquidity.
- Nominal HOPR token from HOPR-DAI UNI-V2 LP liquidity provider token that are staked in the liquidity mining program - HoprFarm. This allows farmers to vote without withdraw their stake.


bonustrack:
The strategy take too long to resolve ~10sec to 20sec for me, is there a way you can reduce the number of queries? Can you use the strategy `erc20-balance-of` as a separate strategy within your space instead of having `hoprScore` inside this strategy?

src/strategies/hopr-uni-lp-farm/index.ts
```diff
@@ -0,0 +1,178 @@
...
+      params: {
+        address: options.hoprAddress,
+        symbol: "HOPR",
+        decimals: 18
+      }
+    }], "1", provider, addresses, snapshot);
+
+  return Object.fromEntries(
+    response.map((value, i) => [
+      addresses[i], // LP token amount * pool's HOPR token balnce / total pool supply
```
bonustrack, line 170: The score returned must use checksumed addresses . I think it will break with the subgraph requests where address get lowercased. You can use `getAddress` function from `@ethersproject/address` for that.

src/strategies/hopr-uni-lp-farm/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  const hoprBalanceOfPool = res[0];
+  const poolTotalSupply = res[1];
+  const response: BigNumber[] = blockTag >= options.farmDeployBlock || blockTag === 'latest'
+    ? res.slice(2, 2 + addresses.length).map((r, i) => (r[0] as BigNumber).add(res[2 + i + addresses.length][1] as BigNumber))
+    : res.slice(2).map(r => r[0] as BigNumber);
+
+  // get block timestamp to search on xDai subgraph
+  const snapshotTimestamp = (await provider.getBlock(blockTag)).timestamp;
+  const snapshotXdaiBlock = await getXdaiBlockNumber(snapshotTimestamp);
```
bonustrack, line 153: You could use Promise.all here to send request in same time and make strategy a bit faster to resolve, like this:
```js
const { snapshotTimestamp, snapshotXdaiBlock } = await Promise.all([
  provider.getBlock(blockTag),
  getXdaiBlockNumber(snapshotTimestamp)
]);
```

src/strategies/hopr-uni-lp-farm/examples.json
```diff
@@ -0,0 +1,26 @@
...
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0x2d8e358487feda42629274ce041f98629bf65cf3",
+      "0x52def6c112bffc66fc8c8258539d4d8c30c5000c",
+      "0x000000359d837dbc86ffcb8be6057062bf1fc1ff",
+      "0x0000000000000000000000000000000000000000"
+    ],
+    "snapshot": "latest"
```
bonustrack, line 24: This should use a specific block number
```suggestion
    "snapshot": 12401620
```

QYuQianchen:
> The strategy take too long to resolve ~10sec to 20sec for me, is there a way you can reduce the number of queries? Can you use the strategy `erc20-balance-of` as a separate strategy within your space instead of having `hoprScore` inside this strategy?

Yes. Mainnet HOPR token balances can be fetched with a separate strategy `erc20-balance-of`.

Will look into all your suggestions!

src/strategies/hopr-uni-lp-farm/index.ts
QYuQianchen: Actually L.152, L.153 and L.155 cannot be merged within a `Promise.all`, because each async call needs value returned from precedent async call.

QYuQianchen:
> The strategy take too long to resolve ~10sec to 20sec for me, is there a way you can reduce the number of queries? Can you use the strategy `erc20-balance-of` as a separate strategy within your space instead of having `hoprScore` inside this strategy?

Removed `erc20-balance-of`. When running `test`, it takes 2.3 seconds less than before:
```
getScores: 7.432s
```

QYuQianchen:
Also, another thing I observed is that it takes a bit longer for running test for the first time (`getScores: 18.344s`) and significantly faster for the second time (`getScores: 5.817s`).

src/strategies/hopr-uni-lp-farm/examples.json
```diff
@@ -0,0 +1,26 @@
...
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0x2d8e358487feda42629274ce041f98629bf65cf3",
+      "0x52def6c112bffc66fc8c8258539d4d8c30c5000c",
+      "0x000000359d837dbc86ffcb8be6057062bf1fc1ff",
+      "0x0000000000000000000000000000000000000000"
+    ],
+    "snapshot": "12401620"
```
bonustrack, line 24: This should be a number not a string
```suggestion
    "snapshot": 12401620
```

bonustrack (approved):
Looks much better now!

PR #206, opened 2021-05-09
Add Celo mainnet

critesjosh:
- Add Celo mainnet

Creating a draft PR until May 19th, when there will be a network upgrade to make ETH transactions compatible with Celo.


bonustrack:
@critesjosh Sound great! You'll need to add an image here too https://github.com/snapshot-labs/snapshot.js/tree/master/src/networks

bonustrack:
@critesjosh If you can just remove the file `src/networks/Celo Rings Large Light BG 500w.png`, your PR will be ready.

critesjosh:
Everything seems to work for reading Celo chain state. I noticed there is a `sendTransaction` method in the utils. This won't work until the Celo Donut hardfork upgrade scheduled for May 19th.

bonustrack:
The function `sendTransaction` is not necessary, it's only used if the space want the delegation voting strategy or settlement plugins.

jaceppo (approved):

critesjosh:
@bonustrack @jaceppo do I need to do anything else to get this merged?

bonustrack:
@critesjosh Hey sorry for the delay, the PR looks good now it will be live in ~10mn

bonustrack (approved):

ChaituVR:
Hi @critesjosh The node https://celo-mainnet--rpc.datahub.figment.io/apikey/e892a66dc36e4d2d98a5d6406d609796/ doesn't work anymore, throws a CORS error, can you point us to a new node or a reliable node provider for this? Thanks

critesjosh:
@ChaituVR You can use https://forno.celo.org

ChaituVR:
Hi, @critesjosh Looks like the new node https://forno.celo.org/ has a limit and doesn't work when we request voting power for more than 2 addresses. we need to find a better node.

ChaituVR:
The actual issue is it can get data for recent blocks only, doesn't work with somewhat older blocks (try 1093764 block from 13 hours ago), maybe it is not a full archive node

critesjosh:
Yea that's not an archive node. Quicknode might be the way to go. These API endpoints are public right? Is there a way to keep the api keys secret since there are rate limits?

ericnakagawa:
hi @ChaituVR @bonustrack would like to revisit a deploy of snapshot on Celo network.

We are working with Infura now. I can provide you an endpoint for Celo proposals. Would that work?

PR #208, opened 2021-05-10
added erc20-quadratic-balance-of for quadratic voting for behodler

gititGoro:
Fixes # .

-  Added quadratic voting strategy based on erc20-balance-of
For reference: https://en.wikipedia.org/wiki/Quadratic_voting

### DIFF
--- a/src/strategies/erc20-quadratic-balance-of/index.ts
+++ b/src/strategies/erc20-quadratic-balance-of/index.ts
@@ -1,0 +1,53 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'goro';
+export const version = '0.1.0';
+
+const abi = [
+  {
+    constant: true,
+    inputs: [
+      {
+        internalType: 'address',
+        name: 'account',
+        type: 'address'
+      }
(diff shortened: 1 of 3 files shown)


src/strategies/erc20-quadratic-balance-of/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [options.address, 'balanceOf', [address]]),
+    { blockTag }
+  );
+  return Object.fromEntries(
+    response.map((value, i) => [
+      addresses[i],
+      Math.sqrt(parseFloat(formatUnits(value.toString(), options.decimals)))
```
bonustrack, line 50: There is a strategy that does exactly the same called "erc20-balance-of-cv", see https://github.com/snapshot-labs/snapshot.js/blob/a0adc547aa0922aa6abd35708a4a292048bca6a2/src/strategies/erc20-balance-of-cv/index.ts

gititGoro: Thank you, I'll make use of that instead.

PR #210, opened 2021-05-11
Add support for tokens in Balancer V2 pools

gerrrg:
- Add queries to the Balancer V2 subgraph
- Updates V1 subgraph queries to target the "balancer" subgraph instead of the "balancer-beta" one
- Add examples.json to validate balances; the final address in addresses[] has an account with Balancer V2 pool tokens holding BAL


bonustrack (approved):
Works well! Thank you for the contribution @gerrrg 🙏

PR #216, opened 2021-05-13
[decentraland-estate-size] Add erc721-with-multiplier and decentraland-estate-size strategies

2fd:
## erc721-with-multiplier 

Allow using an arbitrary multiplier to increase the score

## decentraland-estate-size

Calculate the number of lands that an address own in estates

### DIFF
--- a/src/strategies/decentraland-estate-size/index.ts
+++ b/src/strategies/decentraland-estate-size/index.ts
@@ -1,12 +1,14 @@
-import { formatUnits } from '@ethersproject/units';
-import { multicall } from '../../utils';
+import { getAddress } from '@ethersproject/address';
+import { subgraphRequest } from '../../utils';
+import { getBlockNumber } from '../../utils/web3';
 
 export const author = '2fd';
 export const version = '0.1.0';
 
-const abi = [
-  'function getLANDsSize(address _owner) public view returns (uint256)'
-];
+const DECENTRALAND_MARKETPLACE_SUBGRAPH_URL = {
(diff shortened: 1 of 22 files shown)


src/strategies/decentraland-estate-size/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  options,
+  snapshot
+) {
+  const multipler = options.multiplier || 1
+  const blockTag = typeof snapshot === 'number' ? snapshot : 'latest';
+  const response = await multicall(
+    network,
+    provider,
+    abi,
+    addresses.map((address: any) => [options.address, 'getLANDsSize', [address]]),
```
bonustrack, line 25: Do you have this data on a subgraph? Somehow this call is taking lot of resources on our node, does this call have complex logic on the smart contract? With this implementation the multicall request may fail if there is too much addresses (~350).

jaceppo (approved):
guanlong huang

jaceppo (approved):

src/strategies/decentraland-estate-size/index.ts
2fd: Yes, it has a complex logic. I'm going to reimplement it using our graph 😄

2fd:
@bonustrack I just pushed the changes to use the graph, what do you think?

bonustrack (approved):
It's much faster now, great work! I'll merge the PR it should be live in ~5mn

PR #218, opened 2021-05-15
Add getUrl function for ipfs and ipns urls

samuveth:


src/utils.ts
```diff
@@ -97,6 +97,15 @@ export async function subgraphRequest(url: string, query, options: any = {}) {
   return data || {};
 }
 
+export function getUrl(uri) {
+  const uriSub = uri.substring(0, 7);
```
bonustrack, line 101: I think it's better to use `split('://')` or `startWith('ipfs://')` so it work regardless of the uri scheme lenght (later we would have `swarm://`). Also it should not use an hardcoded gateway. It should use the first gateway from gateways.json list instead.

jaceppo (approved):

bonustrack (approved):

Issue #212, opened 2021-05-12
Checklist for adding a new strategy


bonustrack:
Here is a simple checklist to look when reviewing a PR for a new strategy:

**Overview**
- The strategy must be unique.
- If the strategy does only a single call with an address as input, it's preferable to use the strategy "contract-call" instead of creating a new one.

**Code**
- There should be a maximum of 5 requests, a request can use "fetch" a "subgraphRequest" or "multicall".
- The strategy should not send a request for each voters, this doesn't scale.
- The strategy PR should not add any dependency in Snapshot.js.
- The score returned by the strategy should use the same casing for address than on the input, or should return checksummed addresses.

**Example**
- Example must include at least 1 address with a positive score.
- Example must use a snapshot block number in the past.

**Test**
- The strategy should take less than 10sec to resolve.
- The strategy should work with 500 addresses. [Here is a list of addresses](https://github.com/snapshot-labs/snapshot-strategies/blob/master/test/addresses.json).

**Recommended**
- Add a README.md file that describe the strategy and provide an example of parameters.
- Use string ABI instead of object.

ChaituVR:
> Test
> - The strategy should take less than 10sec to resolve.
> - The strategy should work with 500 addresses. Here is a list of addresses.

@bonustrack less than 10 sec to resolve 500 address? or 1 address? is there any time limit for 500 addresses?

bonustrack:
Less than 10sec for 500 addresses

rude69:
qqq

ChaituVR:
We now have these checks here https://github.com/snapshot-labs/snapshot-strategies#code

PR #220, opened 2021-05-17
Test strategy with jest

ChaituVR:
Tests few scenarios including those mentioned in https://github.com/snapshot-labs/snapshot.js/issues/212

Tests the following cases: (also prints scores and times in console)

image

If this is good, can try to add other scenarios as well, so that user creating strategy will know exactly what we need from a strategy before raising a PR

See in action: 
https://github.com/ChaituVR/snapshot.js/runs/2603281800?check_suite_focus=true


ChaituVR:
@bonustrack  we can use this if we remove 500 addresses test cases? or what other test cases can we remove
image

test/index.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      // wait for all logs to be printed (bug: printed after results)
+      await new Promise((r) => setTimeout(r, 2000));
+    }, 20000);
+
+    it(`getScores ${moreArg || 500} should return an array`, () => {
+      expect(scoresMore).toBeTruthy();
+      expect(Array.isArray(scoresMore)).toBe(true);
+    });
+
+    it(`The strategy for ${moreArg || 500} addresses should take less than 10 seconds to resolve`, () => {
```
bonustrack, line 104: Can you use 15sec here?

test/index.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    example.addresses,
+    example.snapshot
+  );
+}
+
+describe(`Snapshot Strategy Test:\n\nStrategy: "${strategy}"\nQuery: "${example.name}"\n\n`, () => {
+  let scores = null;
+  let getScoresTime = null;
+
+  it('Should run startegy without any errors', async () => {
```
bonustrack, line 50:
```suggestion
  it('Should run strategy without any errors', async () => {
```

ChaituVR:
@bonustrack fixed those suggestions

image

bonustrack (approved):

PR #221, opened 2021-05-20
Add OKExChain node

Frank2333333:
Fixes # .

- 
- 
-


bonustrack:
Heyi @Frank2333333 is the RPC node an archive node? This is mandatory. You will also need to add a multicall contract here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L22 And also add the images here: https://github.com/snapshot-labs/snapshot.js/tree/master/src/networks (you can use images from Twitter).

zccst:
@bonustrack already submit, please review, thanks

zccst:
> Hey @Frank2333333 is the RPC node an archive node? This is mandatory. You will also need to add a multicall contract here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L22 And also add the images here: https://github.com/snapshot-labs/snapshot.js/tree/master/src/networks (you can use images from Twitter).

already submit, please review, thanks

bonustrack (approved):
Looks good!

PR #223, opened 2021-05-21
Feat: DAO Module interaction with Realitio improved

cbrzn:

- Added `submitAnswer` support for `daoModule` plugin
- Added a read function to fetch the information of the question when **it has not finalized yet**


src/plugins/daoModule/index.ts
```diff
@@ -312,4 +374,38 @@ export default class Plugin {
...
+        questionId,
+        `0x000000000000000000000000000000000000000000000000000000000000000${answer}`,
+        bond
+      ],
+      {
+        value: bond.toString()
+      }
+    );
+    const receipt = await tx.wait();
+    console.log('[DAO module] executed vote on oracle:', receipt);
```
bonustrack, line 409: Here you could return the tx directly so it's possible to get it on the UI and do thing like tx.wait(1) instead of sleep().

cbrzn: Is this a pattern that we want to apply? What you are saying makes total sense but I just wanted to keep the same pattern used in other `sendTransaction` methods @bonustrack

bonustrack: Sorry didn't noticed others functions use the same pattern, this is just a suggestion, you can keep it the way it work, it can be changed later for all function if that become necessary.

bonustrack (approved):
Looks good!

src/plugins/daoModule/index.ts
```diff
@@ -161,6 +182,39 @@ const checkPossibleExecution = async (
   };
 };
 
+const retrieveInfoFromOracle = async (
+  provider: JsonRpcProvider,
```
bonustrack, line 186:
```suggestion
  provider: StaticJsonRpcProvider,
```

PR #231, opened 2021-05-25
Update HOPR-Farm strategy

QYuQianchen:
Fixes # .

- Force conversion to lowercase for results coming from subgraph.


src/strategies/hopr-uni-lp-farm/index.ts
```diff
@@ -176,7 +176,7 @@ export async function strategy(
     snapshotXdaiBlock
   );
   const hoprOnXdaiScore = addresses.map(
-    (address) => hoprOnXdaiBalance[address] ?? 0
+    (address) => hoprOnXdaiBalance[address.toLowerCase()] ?? 0
```
bonustrack, line 179: This will not work on the UI. The score returned by the strategy should use checksumed addresses.

QYuQianchen: `addresses` is checksumed because it's from the UI, whilst `hoprOnXdaiBalance` is returned from subgraph where addresses get lowercased. This line essentially converts the `hoprOnXdaiBalance` associated the lower-cased addresses and save it to the position of the  checksumed address.

bonustrack: That's right, my bad, looks good!

PR #233, opened 2021-05-26
[hopr-uni-lp-farm] Improve subgraph queries.

QYuQianchen:
Fixes # .

- Segment list of addresses to per 1000 to respect limits set in the subgraph query.
- Create 353 addreses for testing


bonustrack (approved):
Looks good!

PR #234, opened 2021-05-26
[inverse-xinv] Inverse xINV strategy

0xKiwi:
This is a PR to add a new strategy to snapshot that allows support for staked INV within Anchor, referred to as xINV.

A good portion of the INV supply is staked in Anchor, so it would be fantastic if this could be added :smiley:


bonustrack (approved):
Looks good!

PR #236, opened 2021-05-28
[masterchef-pool-balance] ability to have non integer weights

Janther:

- The `masterchef-pool-balance` strategy could benefit from the use of non integer values. Since `BigNumber` has to be integer we added a `weightDecimals` variable to make final calculations.


bonustrack (approved):
Looks good!

PR #243, opened 2021-06-01
[conquer-uranus] Created conquer-uranus strategy

ASSnonymus:
Fixes # .

- The abi for calling balanceOf has been changed
- 
-


bonustrack:
@ASSnonymus This strategy looks exactly the same as "erc20-balance-of" i recommend to use this one instead of creating a new one.

ASSnonymus:
> @ASSnonymus This strategy looks exactly the same as "erc20-balance-of" i recommend to use this one instead of creating a new one.

Yes, we have had problems with the balanceOf function. Possibly it is because the header of our function is different so I have modified the abi string. That is the reason why it is practically similar.

bonustrack:
@ASSnonymus The strategy erc20-balance-of works for me with your token see:
image
You can test it here: https://snapshot.org/#/playground/erc20-balance-of

ASSnonymus:
I have tested it and it works, however the voting does not work, it returned an error from the function balanceOf and map undefined. What can it be?

ASSnonymus:
Now it's working, I don't understand anything XD. Anyway, thank you very much @bonustrack .

PR #245, opened 2021-06-02
[modefi] feat: modefi strategy

OxAL7:
Fixes # .

- 
- 
-


bonustrack:
@OxAL7 Does the test work for you? For me it fail with:
```call revert exception (method="balanceOf(address)", errorSignature=null, errorArgs=[null], reason=null, code=CALL_EXCEPTION, version=abi/5.0.4)```
Please make sure the example file is properly set.

**Comment by OxAL7** [comment]
> @OxAL7 Does the test work for you? For me it fail with:
> `call revert exception (method="balanceOf(address)", errorSignature=null, errorArgs=[null], reason=null, code=CALL_EXCEPTION, version=abi/5.0.4)`
> Please make sure the example file is properly set.

Yeah I properly set the example.json file. I am not sure why it happens. I copy paste the `erc20-balance-of` and then modified the address of token.

**Comment by OxAL7** [comment]
@bonustrack I have fixed the issue by changing the `snapshot` value in the example.json.
Also I ran the lint command to fix the prettier issue. Please take a look and let me know your opinion.
Thank you!

File: src/strategies/modefi/examples.json [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+        "symbol": "MOD",
+        "decimals": 18
+      }
+    },
+    "network": "1",
+    "addresses": [
+      "0x6f0b57730f7d51476e93cc46093ec9139010caa2",
+      "0xe5aec22bb09186e205cf3366236f1fc8e9606fa4"
+    ],
+    "snapshot": "latest"
```
bonustrack, line 17: You should use a block number here, not "latest"
```suggestion
    "snapshot": 12560234
```

bonustrack: Can you try if it work with this number? My guess is that before you were using a block number that was before the token contract creation

OxAL7: Sure, I will try.

OxAL7: @bonustrack yay! it works!

bonustrack (approved):
Looks good!

PR #246, opened 2021-06-02
Add a full sync node for Heco chain

dogeswapfans:
Fixes # .

- Current rpc node isn't a full sync node, which will cause ""missing trie node" issue, try this new added node provided by Heco community.

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -275,7 +275,10 @@
     "shortName": "heco",
     "chainId": 128,
     "network": "Mainnet",
-    "rpc": ["https://http-mainnet.hecochain.com"],
+    "rpc": [
+      "https://http-mainnet-node.defibox.com",
+      "https://http-mainnet.hecochain.com"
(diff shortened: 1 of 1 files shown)


dogeswapfans:
Not sure is it necessary to delete previous node which isn't a full sync node.

bonustrack (approved):
Let's try

dogeswapfans:
@bonustrack Does it need to update the release of npm package?

bonustrack:
@dogeswapfans No, it should be live already, does it work for you? You may need to clear your cache

dogeswapfans:
Unfortunately, it seems this is still not a full sync node, same error output on the console

bonustrack:
I recommend get in touch with the officials, they might be able to help.

dogeswapfans:
Yes, we asked, they ask projects to build own full sync node, which is unfriendly to community projects like your team

PR #249, opened 2021-06-03
Run strategy test on other pull request events

ChaituVR:


bonustrack:
@ChaituVR Isn't "push" the event we need? Like this: https://github.com/snapshot-labs/snapshot.js/blob/master/.github/workflows/nodejs.yml#L3

ChaituVR:
@bonustrack on push there is no pull_request object to check title and label... looks like `synchronize` can take care of it
```
synchronize | commit(s) pushed to the pull request
```
https://frontside.com/blog/2020-05-26-github-actions-pull_request/

bonustrack (approved):

PR #252, opened 2021-06-06
[governor-alpha] Governor alpha strategy

ChaituVR:

### DIFF
--- a/src/strategies/governor-alpha/index.ts
+++ b/src/strategies/governor-alpha/index.ts
@@ -1,0 +1,57 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'Arr00';
+export const version = '0.1.0';
+
+const abi = [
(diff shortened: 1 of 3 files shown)


src/strategies/governor-alpha/index.ts
```diff
@@ -0,0 +1,57 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+
+export const author = 'snapshot-labs';
+export const version = '0.0.1';
+
+const abi = [
```
bonustrack, line 7: Can you use this instead:
`function getCurrentVotes(address account) external view returns (uint96)`
same format than here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/erc20-balance-of/index.ts#L7-L9

ChaituVR: @bonustrack Added that 👍 works fine

PR #253, opened 2021-06-06
Fix: Handle edge case when dao module minimum bond is zero

cbrzn:
-  If DAO Module has minimum bond set to zero, a default bond of 0.001 ETH is going to be set


bonustrack (approved):
Looks good!

PR #275, opened 2021-06-27
Test strategy with latest snapshot

ChaituVR:


bonustrack:
- We should test the format of the score return to be sure it's an array of object with a number as value and address as key.
- Remove 10 sec test on the "latest" test

test/index.spec.ts
```diff
@@ -51,27 +51,30 @@ describe(`\nTest strategy "${strategy}"`, () => {
...
   }, 2e4);
 
-  it('Should return an array', () => {
+  it('Should return an array of object with addresses', () => {
     expect(scores).toBeTruthy();
+    // Check array
     expect(Array.isArray(scores)).toBe(true);
+    // Check array contains a object
+    expect(typeof scores[0]).toBe('object');
+    // Check object contains atleast one address from example.json
```
bonustrack, line 65:
```suggestion
    // Check object contains at least one address from example.json
```

bonustrack (approved):

PR #278, opened 2021-06-29
Add strategy [iotex-staked-balance]

maswalker:
add strategy iotex-staked-balance


maswalker:
@ChaituVR @bonustrack

bonustrack (approved):
Looks good! It will be up in ~5mn

PR #282, opened 2021-06-30
[erc20-balance-of-quadratic-delegation] A new strategy for quadratic delegations 

ferittuncer:
This is a variant of erc20-balance-of-delegation strategy that preserves voting power in the case of quadratic voting.

The problem with erc20-balance-of-delegation is that if used with quadratic voting, the total voting power that the delegate has is less than expected.
In other words, delegation causes voting power loss. To fix this, we created this strategy.

Say there are two persons, with 100 vote credits each, in a quadratic voting setting. Person A delegates to person B.

Using erc-20-balance-of-delegation:

`Voting Credits of Person B = Voting Credits of Person A + Voting Credits of Person B = 200`

Person A could cast 10 (10^2=100) votes individually, similarly, Person B could cast 10 (10^2=100) votes individually, so in total, they could cast 20 votes.
But now, delegate can't cast 20 votes (20^2=400) because it needs 400 credits.

Solution:

`Voting Credits of Person B = (Sqrt(Voting Credits of Person A) + Sqrt(Voting Credits of Person B))^2 = 400`

Now the delegate can represent the sum of individual voting powers.


ChaituVR:
Hi @ferittuncer 

Make sure to add your strategy to https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/index.ts#L106

and make sure to test your strategy with 
```bash
npm run test --strategy=erc20-balance-of-quadratic-delegation

```

src/strategies/erc20-balance-of-quadratic-delegation/index.ts
```diff
@@ -0,0 +1,49 @@
+import { strategy as erc20BalanceOfStrategy } from '../erc20-balance-of';
+import { getDelegations } from '../../plugins/delegation/utils';
+
+export const author = 'bonustrack';
```
ChaituVR, line 4: This should be your Github username

ferittuncer:
Please don't merge yet as I haven't tested it actually, but coded blindly. Because I'm unable to run the test suite.

ChaituVR:
@ferittuncer If you are still having issues in your local system, you can check logs in the github actions on this PR 
image
which seems to be working fine to test your strategy

src/strategies/erc20-balance-of-quadratic-delegation/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const delegations = await getDelegations(
+    space,
+    network,
+    provider,
+    addresses,
+    options,
+    snapshot
+  );
+  if (Object.keys(delegations).length === 0) return {};
+  console.debug('Delegations', delegations);
```
ChaituVR, line 24: Can we remove the unwanted logs?

ferittuncer:
> @ferittuncer If you are still having issues in your local system, you can check logs in the github actions on this PR
> image
> which seems to be working fine to test your strategy

The build is successful but does it yield correct values?

ferittuncer:
As we verified the problem is indeed in my environment, I started to work on a VM now.

ferittuncer:
Screenshot from 2021-07-05 14-32-24

As I suspected, it's not working correctly. Logs indicate a linear sum of voting scores, instead of quadratic.

I'm having hard time to print my debugging console logs, I guess this is an issue with Jest (https://stackoverflow.com/a/54190964/1079908). If you have a tip on this, I would appreciate.

src/strategies/erc20-balance-of-quadratic-delegation/index.ts
ferittuncer: Actually, this is a `console.debug` not a `console.log` so it will only print when in debugging mode. Do you still think we should remove it? If yes, I'm going to remove when I'm done with debugging.

ferittuncer:
> Screenshot from 2021-07-05 14-32-24
> 
> As I suspected, it's not working correctly. Logs indicate a linear sum of voting scores, instead of quadratic.
> 
> I'm having a hard time printing my debugging console logs, I guess this is an issue with Jest (https://stackoverflow.com/a/54190964/1079908). If you have a tip on this, I would appreciate it.

Found the origin of this problem, and fixed it in the last commit.

bonustrack (approved):

bonustrack:
@ferittuncer Just merged! It should be live in ~10mn

PR #286, opened 2021-07-01
feat: Added Aave Space validation function and generic validation test suite

kartojal:
- Add Aave Space validation function for validation of new proposals
- Added a generic way to test "validation" functions via `npm run test:validation --validation=your-validation-name", by default it runs the basic validation.


package.json
```diff
@@ -55,6 +55,7 @@
     "build:cjs": "rollup -c rollup.cjs.config.js",
     "dev": "rollup -c -w",
     "test": "jest test/index.spec.ts --runInBand",
+    "test:validation": "jest test/validation.spec.ts --runInBand",
```
bonustrack, line 58: It would be better to have a build script before running the test to be sure it use latest version of the validation code, like this:
```suggestion
    "test:validation": "npm run build:cjs && jest test/validation.spec.ts --runInBand",
```

src/validations/aave-space-validation/index.ts
```diff
@@ -0,0 +1,56 @@
+import { call, getScores } from '../../utils';
```
bonustrack, line 1: The `call` function doesn't seem to be used you can remove it
```suggestion
import { getScores } from '../../utils';
```

src/validations/aave-space-validation/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import { BigNumber } from '@ethersproject/bignumber';
```
bonustrack, line 2: This one too
```suggestion
```

test/validation.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  let getValidationTime = null;
+
+  it('Validation should run without any errors', async () => {
+    try {
+      const performanceStart = performance.now();
+      isValid = await snapshot.utils.validations[validation](
+        example.userAddress,
+        example.space || defaultSpace,
+        {},
+        {}
```
bonustrack, line 37: Would be great to have the validation params from example here
```suggestion
        example.validation.params || {}
```

bonustrack (approved):
Perfect!

PR #288, opened 2021-07-05
[modefi-staking] Add Strategies - Modefi

Modefi:
Fixes # .

- Add Strategies
- 
-


src/strategies/modefiMultichain/index.ts
```diff
@@ -0,0 +1,104 @@
+import { subgraphRequest } from '../../utils';
+import getProvider from '../../utils/provider';
+import strategies from '..';
+
+export const author = 'Modefi';
+export const version = '0.0.1';
+
+const defaultGraphs = {
+  '56': 'https://api.thegraph.com/subgraphs/name/apyvision/block-info',
+  '250': 'https://api.thegraph.com/subgraphs/name/ord786/fantom-blocks'
```
ChaituVR, line 10: Looks like this is just a copy of `multichain` strategy, is there any reason to make it seperate? I think you can use `multichain` strategy in your space and that strategy can support this networt

https://github.com/snapshot-labs/snapshot.js/blob/master/src/strategies/multichain/examples.json#L41-L43

src/strategies/index.ts
```diff
@@ -201,6 +203,8 @@ export default {
   brightid,
   'inverse-xinv': inverseXINV,
   modefi,
+  'modefiStaking': modefiStaking,
```
ChaituVR, line 206: The strategy name should be like this.. can u plz rename the folder name too.. 

```suggestion
  'modefi-staking': modefiStaking,
```

ChaituVR (changes requested):

Modefi:
Strategy updated

ChaituVR (commented):
LGTM

bonustrack (approved):

PR #289, opened 2021-07-06
Test cases failing if addresses are different word case

ChaituVR:
Fixes issue when running `uniswap` strategy,

where

`0x79317fC0fB17bC0CE213a2B50F343e4D4C277704` is the output 
and `0x79317fc0fb17bc0ce213a2b50f343e4d4c277704` is given in example.json


bonustrack (approved):
Looks good

PR #290, opened 2021-07-06
[eth-with-balance] Add eth-with-balance strategy

AronVanAmmers:
Adds a new strategy `eth-with-balance` which gives a voting power of 1 to any address with a configurable minimum ETH balance.

### DIFF
--- a/src/strategies/eth-with-balance/index.ts
+++ b/src/strategies/eth-with-balance/index.ts
@@ -1,0 +1,29 @@
+import { strategy as ethBalanceOfStrategy } from '../eth-balance';
+import { verifyResults } from '../ocean-marketplace/oceanUtils';
+
+export const author = 'Aron van Ammers';
+export const version = '0.1.0';
+
+export async function strategy(
+  space,
+  network,
+  provider,
+  addresses,
+  options,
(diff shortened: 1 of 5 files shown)


src/strategies/eth-with-balance/index.ts
```diff
@@ -0,0 +1,29 @@
+import { strategy as ethBalanceOfStrategy } from '../eth-balance';
+import { verifyResults } from '../ocean-marketplace/oceanUtils';
+
+export const author = 'Aron van Ammers';
```
bonustrack, line 4: This should be your GitHub username
```suggestion
export const author = 'AronVanAmmers';
```

src/strategies/eth-with-balance/index.ts
```diff
@@ -0,0 +1,29 @@
+import { strategy as ethBalanceOfStrategy } from '../eth-balance';
+import { verifyResults } from '../ocean-marketplace/oceanUtils';
```
bonustrack, line 2: This doesn't seem to be used you can remove it
```suggestion
```

bonustrack (approved):
Looks good!

PR #292, opened 2021-07-07
[ruler-staked-token] Add ruler-staked-token strategy

drop-out-dev:
- Add a strategy for getting the number of tokens staked in the BonusRewards contract


bonustrack (approved):
Looks good!

PR #294, opened 2021-07-08
Feat: Support RealitioERC20

cbrzn:

-  Allows the user to submit a bond on RealitioERC20 oracles
-  Retrieve symbol and token information so it can be shown as expected on the UI
-


bonustrack (approved):
Looks good

PR #304, opened 2021-07-13
Test returned addresses should be either same case as input addresses or checksum addresses

ChaituVR:


bonustrack (approved):
Looks good!

PR #309, opened 2021-07-15
[zrx-voting-power] ZRX voting power strategy

lyaunzbe:
Hello! I work with 0x labs on the platform side. I am making this PR to hopefully add the zrx-voting-power strategy (calculates voting power based on staked + wallet balances). Thank you!

Fixes # .

- 
- 
-


OhArthits (approved):

src/strategies/zrx-voting-power/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      {
+        internalType: 'uint256',
+        name: 'votingPower',
+        type: 'uint256'
+      }
+    ],
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 42:
```suggestion
const abi = [
  'function getVotingPower(address account, bytes32[] operatedPoolIds) view returns (uint256 votingPower)'
]
```

ChaituVR (changes requested):

src/strategies/zrx-voting-power/index.ts
lyaunzbe: thank you, i've made this change

ChaituVR:
@lyaunzbe there is conflict, can u resolve it plz :) everything else looks good to me

lyaunzbe:
@ChaituVR should be all set now, thank you!

src/strategies/zrx-voting-power/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  network,
+  provider,
+  addresses,
+  options,
+  snapshot
+) {
+  const blockTag =
+    typeof snapshot === 'number'
+      ? snapshot
+      : await provider.getBlockNumber(snapshot);
```
bonustrack, line 35: You don't need to query for current block number here, 'latest' will work.
```suggestion
      : 'latest';
```

bonustrack (approved):
Looks good!

PR #303, opened 2021-07-12
[balancer-smart-pool] Add/balancer smart pool token score

kibagateaux:
Fixes #115  .

Changes proposed in this pull request:
- Uses shares in a Balancer Smart Pool to track ownership of a governance token held in pool
- Requires 3 addresses - ERC20 token address representing shares in pool, contract address of smart pool to check total governance token holdings, ERC20 governance token address to check in smart pool
- Ran `npm run test --strategy-balancer-smart-pool` and results looked correct. We will be using this at MetaFactory and we are fine testing in prod

PR #311, opened 2021-07-15
[balancer-smart-pool] update balancer-smart-pool strategy to compare weights to total supply

kibagateaux:
Updates #303

I made the assumption that Snapshot had all `addresses` before calling the strategy. This update checks total supply of the pool shares token in the calculation instead of using the sum of shares owned by `addresses`


ChaituVR (commented):
LGTM

PR #319, opened 2021-07-19
[tomb-finance] New strategy for tomb.finance

thecryptoundertaker:
New voting strategy for tomb.finance

Calculates the amount of TShares in:

user's wallet
LP in user's wallet
LP in Cemetery TSHARE-FTM LP pool
Masonry single-staking pool
auto-compounding vaults of TSHARE-FTM LP
As we are using the quadratic voting, to calculate the amount of votes the Math.sqrt() is used on the total amount of owned TShares.

### DIFF
--- a/src/strategies/tomb-finance/index.ts
+++ b/src/strategies/tomb-finance/index.ts
@@ -113,7 +113,6 @@ export async function strategy(
 
   multi = new Multicaller(network, provider, abi, { blockTag });
 
-  multi.call(`tshare.totalSupply`, TSHARE_TOKEN_ADDRESS, 'totalSupply');
   multi.call(`lp.tshareBalance`, TSHARE_TOKEN_ADDRESS, 'balanceOf', [
     TSHARE_LP_TOKEN_ADDRESS
   ]);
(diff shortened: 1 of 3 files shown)


src/strategies/tomb-finance/index.ts
```diff
@@ -0,0 +1,219 @@
...
+        internalType: 'address',
+        name: '',
+        type: 'address'
+      }
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 94:
```suggestion
const abi = [
  'function balanceOf(address) view returns (uint256 amount)',
  'function userInfo(uint256, address) view returns (uint256 amount, uint256 rewardDebt)',
  'function totalSupply() view returns (uint256)',
  'function strategy() view returns (address)'
];
```

src/strategies/tomb-finance/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    multi.call(
+      `vaultTokens.lpBalance.${token.address}`,
+      CEMETERY_ADDRESS,
+      'userInfo',
+      ['1', vaultStrategiesResult.vaultStrategies[token.address]]
+    );
+  });
+
+  addresses.forEach((address: any) => {
+    multi.call(`tshare.${address}`, TSHARE_TOKEN_ADDRESS, 'balanceOf', [
```
ChaituVR, line 136: Multicalls for each address won't scale, is there any other alternate way to calculate the score?

src/strategies/tomb-finance/index.ts
thecryptoundertaker: Fixed!

PR #320, opened 2021-07-19
Add new RPC - rpc.ftm.tools for Fantom Opera blockchain

thecryptoundertaker:
Introduce additional RPC address for Fantom Opera network.

The new RPC is https://rpc.ftm.tools which is an RPC aggregator, widely used in Fantom Opera project and by Fantom Opera users.

It has become very popular during the high load period on Fantom, which caused the main RPC to lag and fail to respond at all.


bonustrack (approved):
Sound good!

PR #321, opened 2021-07-19
[masterchef-pool-balance-price] Add  strategy to get the balance or price of any combination of tokens for a pool using a masterchef contract.

joaomajesus:
- Adding a new  masterchef strategy that allows to:
  - select which token of the uniPair to use, or use both.
  - get both the balance and the price of the tokens.
  - apply an anti-whale measure to the result to reduce the impact of big balances.


src/strategies/masterchef-pool-balance-price/index.ts
```diff
@@ -0,0 +1,560 @@
...
+        internalType: 'uint8',
+        name: '',
+        type: 'uint8'
+      }
+    ],
+    payable: false,
+    stateMutability: 'view',
+    type: 'function'
+  }
+];
```
ChaituVR, line 188:
```suggestion
const abi = [
  'function userInfo(uint256, address) view returns (uint256 amount, uint256 rewardDebt)',
  'function totalSupply() view returns (uint256)',
  'function getReserves() view returns (uint112 _reserve0, uint112 _reserve1, uint32 _blockTimestampLast)',
  'function token0() view returns (address)',
  'function token1() view returns (address)',
  'function decimals() view returns (uint8)'
];
```

src/strategies/masterchef-pool-balance-price/index.ts
```diff
@@ -0,0 +1,450 @@
...
+    network,
+    provider,
+    abi,
+    tokenCalls,
+    { blockTag }
+  );
+
+  for (let i = 0; i < userInfoChunks.length; i++) {
+    const value = userInfoChunks[i];
+    const score = await processValues(value, tokenResponse, network, provider, blockTag)
```
bonustrack, line 444: Behind this function i can see there is a lot of call and we can not predict the number of userInfoChunks, this might be too heavy to process, we usually recommend to limit a strategy with a maximum of 3 multicall or requests. Is there any way you can simplify the strategy so it does less calls?

joaomajesus: I made sure that scenario wouldn't happen I'm caching the price for each token on every execution of the strategy.
There are only 2 multicalls in total, behind the ProcessValues there are at most 2 requests to CoinGeco to fetch the prices. That call to CoinGecko is cached so there will be at the maximum only 2 requests. 
If the configuration is set to get the price for only 1 token, then the total will be 1 request.
If the configuration is not set to get the price then no request will be made.
If I did the single multicall like in the original masterchef-pool-balance, the test would fail at 400 addresses because of the size of the body.

Is 4 multicals or requests, a difference of 1, that impactful?

From the tests check it looked like the cache was working well.

bonustrack: Got it, thanks for the explanation

bonustrack (approved):
Looks good!

src/strategies/masterchef-pool-balance-price/index.ts
```diff
@@ -0,0 +1,450 @@
+import { formatUnits } from '@ethersproject/units';
+import { multicall } from '../../utils';
+import { BigNumber} from '@ethersproject/bignumber';
+import fetch from 'cross-fetch';
+import { inspect } from 'util';
```
bonustrack, line 5: @joaomajesus Can you just remove `inspect`? This wont work on browser

joaomajesus: ah! ok will do it right away

joaomajesus: done

PR #324, opened 2021-07-20
Changes for _ipfsUrl refactor

samuveth:


src/utils.ts
```diff
@@ -85,8 +85,13 @@ export async function subgraphRequest(url: string, query, options: any = {}) {
 }
 
 export function getUrl(uri) {
```
bonustrack, line 87: Can we add possibility to edit the gateway like this?
```suggestion
export function getUrl(uri, gateway?) {
```

samuveth: Like that?

bonustrack (approved):

src/utils.ts
bonustrack: Yes

PR #326, opened 2021-07-20
[zrx-voting-power] Hotfix for zrx-voting-power strategy

lyaunzbe:
Fixes a case-sensitivity related bug when fetching voting power, specific to addresses that also happen to be staking pool operators.


bonustrack (approved):
Looks good!

PR #334, opened 2021-07-23
[immutable-x] Add Immutable X Snapshot Strategy

igx-C:
- Add Immutable X (Layer-2) snapshot strategy
- Currently, vote weighting is simply a linear sum of a user's L1 and L2 (zk-rollup) IMX balance


src/strategies/immutable-x/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    _space,
+    network,
+    provider,
+    addresses,
+    options,
+    _snapshot
+): Promise<Record<string, number>> {
+    try {
+        return combineBalanceScores([
+            await getL1Balances(network, provider, options, addresses),
```
bonustrack, line 131: The strategy should check balance at a specific block number, but here I don't see any block number (_snapshot is not used), this will make result changing overtime, and doesn't make it really usable. Is Immutable X EVM compatible? Isn't possible to use it with Ethers.js provider?

igx-C: Unfortunately, we don't currently support block (batch on l2)-level snapshots in our system (this will change soon) but current idea is to take a snapshot once based on `latest` right before the proposal is published and not have it change again for the lifetime of the proposal. Is there a way to enforce this?

bonustrack (changes requested):
You need to list the strategy on the file `./src/strategies/index.ts` otherwise test will fail.

bonustrack:
@igx-C Actually we've moved the strategies to another repository here:  https://github.com/snapshot-labs/snapshot-strategies Could you make your PR there? Sorry for the trouble!

bonustrack:
Closing this PR for now, new strategies must be added into this repository now: https://github.com/snapshot-labs/snapshot-strategies, feel free to reopen this PR if needed.

PR #339, opened 2021-07-27
Ref get url

samuveth:
https://github.com/snapshot-labs/snapshot/pull/488


src/utils.ts
```diff
@@ -84,13 +84,19 @@ export async function subgraphRequest(url: string, query, options: any = {}) {
   return data || {};
 }
 
-export function getUrl(uri) {
+const defaultGateway = process.env.VUE_APP_IPFS_GATEWAY || gateways[0];
```
bonustrack, line 87: Snapshot.js should not use process.env, this is something for the UI only.

bonustrack:
This was merged already.

PR #342, opened 2021-07-28
[erc721-with-tokenid] Add ERC721 with tokenid strategy

dimsome:
Add Strategy for ERC721 token with a specified tokenId. Useful for POAPs to allow voting with a specified ID rather than all.


ChaituVR (commented):
LGTM 👍

bonustrack (approved):
Looks great!

PR #348, opened 2021-08-02
Remove strategies

bonustrack:


dapplion:
Why? Where are the strategies now?

samuveth:
@dapplion they are here https://github.com/snapshot-labs/snapshot-strategies

PR #353, opened 2021-08-17
Add Wallet type support for meta actions

samuveth:


src/sign/index.ts
```diff
@@ -106,11 +109,11 @@ export default class Client {
     return await this.sign(web3, address, message, type);
   }
 
-  async follow(web3: Web3Provider, address: string, message: Follow) {
+  async follow(web3: Wallet, address: string, message: Follow) {
```
bonustrack, line 112: Why not doing like this?
```suggestion
  async follow(web3: Web3Provider | Wallet, address: string, message: Follow) {
```

samuveth: I don't see when we would need `Web3Provider` there?

bonustrack: Someone may want to sign using his wallet, not a meta action

bonustrack: But yea is an edge case, doesnt matter much actually

bonustrack (approved):
Looks good

PR #354, opened 2021-08-17
Add PALM network support

gnkz:
Changes:
- Add PALM network configuration
- Add PALM network logo


ChaituVR:
@gnkz you sure the multi-call address is correct? any idea why I get this error:
image

gnkz:
Hey @ChaituVR ! Thanks for taking a look at this.

I'm pretty sure the address is correct, the contract is verified here: https://explorer.palm.io/address/0xfFE2FF36c5b8D948f788a34f867784828aa7415D/contracts

We are using the MakerDAO one https://github.com/makerdao/multicall/blob/master/src/Multicall.sol but with a different solidity version. Do you think that could be the issue?

gnkz:
I tested it using `@makerdao/multicall` and seems to work fine. Here is the code I used to test it https://github.com/gnkz/palm-multicall-test

gnkz:
I tested it using `@makerdao/multicall` and seems to work fine. Here is the code I used to test it https://github.com/gnkz/palm-multicall-test

gnkz:
Another test I did was to modify the `test/multicaller.ts` file like in this gist https://gist.github.com/gnkz/cae0728d116ed68df1c5e311c945a4a2 and it worked

ChaituVR:
@gnkz Can you try using `multicall` method in utils, it seems to be working fine with `Multicaller` but not with multical, not sure whats going wrong, maybe some thing wrong with `aggregate` method in contract?

gnkz:
Hey @ChaituVR ! I added a file under `test/multicall.ts` to test the `multicall` function and it works for me. Can you check that file and try to run it to see if it works for you? I can remove that file once we confirm that everything looks good. Thanks!

ChaituVR:
@gnkz Sorry for all the confusion, the issue is with my local where I am testing that method, It works great now, Thank you for your patience

bonustrack (approved):

PR #358, opened 2021-08-27
add commentBox in list of plugin

spiritbro1:
add snapshot plugin "Comment Box" onto snapshot.js


bonustrack:
@spiritbro1 Can you add an image for the plugin like this https://github.com/snapshot-labs/snapshot.js/blob/master/src/plugins/aragon/logo.png ?

spiritbro1:
> @spiritbro1 Can you add an image for the plugin like this https://github.com/snapshot-labs/snapshot.js/blob/master/src/plugins/aragon/logo.png ?

ok will do

spiritbro1:
@bonustrack logo added

PR #372, opened 2021-09-14
SafeSnap - Reality Module Contract compatibility 

carlosfebres:
### Summary 
Update SafeSnap plugin to use the new version of the [Reality Module](https://github.com/gnosis/zodiac-module-reality) (previously called `Dao Module`). 

### Fix
- The `executor` field has been renamed to `avatar`.
- Keep compatibility with the previous version.


bonustrack (approved):
Looks good!

PR #378, opened 2021-09-20
Set Version 0.2.8

ChaituVR:
⚠️ We've moved the voting strategies to another repository, if you want to add or edit strategy you must do it here: https://github.com/snapshot-labs/snapshot-strategies


bonustrack (approved):
Looks good. When you update the version you should use a branch name on this format: `v0.2.8` and keep the branch once merged.

ChaituVR:
@bonustrack ok 👍 will do that from next version

Issue #381, opened 2021-09-27
Custom Network Support


CjHare:
A new Snapshot user here, it looks like a great product 👍 

The use case have for Snapshot is developing (& testing) a custom strategy against an in-memory network. When the strategy is eventually live, the CI testing will remain as part of the CI/CD pipeline (so it'll be a feature needed beyond the initial strategy development).

Although running Snapshot against a network defined outside of the `networks.json` is not is not currently a feature, if it is a feature you'd like, I'll happily port the changes from my open PR https://github.com/snapshot-labs/snapshot-strategies/pull/79

Generally speaking, this change switches the sourcing of the `networks json` from being hard-coded import to a CLI option, with the default being the `network.json`

The use case(s) supported by such a change are:
- Running strategy tests on CI boxes 
- Running Snapshot against private networks

Let me. know if this is a change you'd like me to create a PR on, or if there's a better solution in this case with Snapshot 🙂 

### Desired behavior
Run Snapshot against a network whose definition is in a custom json file.

### Actual behavior
The neworks.json from Snapshot.js is used.

### Steps to reproduce the behavior
Running a test on the default strategy, with the network being a local Ethereum node.

ChaituVR:
Now it is very easy to add a network :) https://docs.snapshot.org/developer-guides/networks

PR #383, opened 2021-09-29
feat: supporting custom networks file

CjHare:
An implementation for the code side for https://github.com/snapshot-labs/snapshot.js/issues/381

Figured there may be changes wanted around precisely how or which methods of input are to be supported, or where that logic resides, so I've held off on updating the README. (When that's set, then I'll add the update to the docs).

Reviews are welcome 🙂

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -11,7 +11,7 @@ import validations from './validations';
 import { signMessage, getBlockNumber } from './utils/web3';
 import { getHash, verify } from './sign/utils';
 import gateways from './gateways.json';
-import networks from './networks.json';
+import { networks } from './utils/networks';
 
 export const SNAPSHOT_SUBGRAPH_URL = {
   '1': 'https://api.thegraph.com/subgraphs/name/snapshot-labs/snapshot',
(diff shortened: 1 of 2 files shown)


ChaituVR:
Nice Idea @CjHare

bonustrack (commented):
Files in src/ shouldn't include process.env or others env specific variables. Can't you just do like this to add a new network:

```ts
import networks from '@snapshot-labs/snapshot.js/src/networks.json';

networks[12345] = {
  id: '12345',
  name: 'Custom node',
  rpc: ['http://localhost:88383']
  multicall: '0x123456'
};
```

ChaituVR:
@bonustrack This one is actually for `snapshot-strategies` networks come from `snapshot.cjs` and won't be able to change them without modifying `node_modules` folder

CjHare:
> Files in src/ shouldn't include process.env or others env specific variables.
That does make sense. 

As @ChaituVR points out these changes avoids some code duplication in a `snapshot-strategies` PR by allowing the networks to be from an alternative source to the `networks.json` (in this case using a local file instead). 
What would be the recommendation to achieve that change within the Snapshot packages/project structure?

bonustrack:
@CjHare Sorry for late review, this PR is not relevant for this repository, the change should be on https://github.com/snapshot-labs/snapshot-strategies , I would suggest make custom network an option in getScoreDirect function instead of using npm vars or config vars.

PR #384, opened 2021-09-29
fix: update `Multicall` contract

johnletey:
When using the Moonbase Alpha network with Snapshot, we realized that we couldn't vote.

After doing some digging, we found that the issue was occurring when calling the `aggregate` function.

We have gone ahead and deployed a new `Multicall` contract (you can find the verified code [here](https://moonbase-blockscout.testnet.moonbeam.network/address/0xf09FD6B6FF3f41614b9d6be2166A0D07045A3A97/contracts)).

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -390,7 +390,7 @@
     "shortName": "Moonbase",
     "chainId": 1287,
     "network": "testnet",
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
Looks good!

PR #386, opened 2021-09-30
Addded new network klatyn

NirajBhattarai:
⚠️ We've moved the voting strategies to another repository, if you want to add or edit strategy you must do it here: https://github.com/snapshot-labs/snapshot-strategies


ChaituVR (changes requested):
Doesn't look like a valid node
I m getting following error:
```
Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.4.5)
```

NirajBhattarai:
> Doesn't look like a valid node I m getting following error:
> 
> ```
> Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.4.5)
> ```

changed chain id now can you check

> Doesn't look like a valid node I m getting following error:
> 
> ```
> Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.4.5)
> ```

ChaituVR:
image
Looks like the node is not accessible from everywhere

PR #390, opened 2021-10-04
Add schema for voting delay and period

samuveth:
For https://github.com/snapshot-labs/snapshot/pull/815

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -172,13 +172,15 @@
         "plugins": {
           "type": "object"
         },
-        "votingDelay": {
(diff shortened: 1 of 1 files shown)


src/schemas/space.json
```diff
@@ -171,6 +171,14 @@
         },
         "plugins": {
           "type": "object"
+        },
+        "votingDelay": {
```
bonustrack, line 175: Can you make "voting" a root param with delay and period as sub param like this:

{
  "voting": {
    "delay": 1651651,
    "period": 894984
  }
}

PR #420, opened 2021-11-08
Network param for getSpaceUri function

### LINKED ISSUE #401: Support Rinkeby ENS resolver
Once this is merged: https://github.com/snapshot-labs/snapshot.js/pull/400/files

The `getSpaceUri` makes an ENS lookup to find the space settings for a certain ENS name. Would be nice if this could also support a testnet resolver, like: https://rinkeby.etherscan.io/address/0xf6305c19e814d2a75429fd637d01f7ee0e77d615

This way, for testing and local development and everything, if you don't already have a mainnet ENS domain, you can register one e.g. on Rinkeby, and use that to create a new space. The exact implementation of this needs to be discussed, since you probably don't want this to result in real spaces. But some form of dev/local mode would be nice. The frontend would then also need an update to make that somehow configurable.

mktcode:
Fixed #401 

Adds support for different ens resolver addresses per network. Defaults to main net.

The frontend can then be updated like this:

```js
const uri = await getSpaceUri(
  props.spaceId,
  import.meta.env.VITE_DEFAULT_NETWORK
);
```
This would allow to use ENS names registered e.g on rinkeby for testing.


bonustrack:
How this will be used? For testing right with local snapshot and hub?

mktcode:
> How this will be used? For testing right with local snapshot and hub?

Exactly. I still don't have a main net ENS domain.

ChaituVR:
LGTM

bonustrack (approved):

PR #427, opened 2021-11-15
Add Cronos network

tendiefrenco:
Adding the Cronos network.


ChaituVR:
Hi @tendiefrenco this works but we should also add a image of the network here https://github.com/snapshot-labs/snapshot.js/tree/master/src/networks

bonustrack (approved):

PR #432, opened 2021-11-23
feat: add Share function of validation proposal

jinanton:
added function to proposal validation based on share of the total supply of the tokens.


ChaituVR:
@jinanton This validation is just passing different strategies to validate right. we can pass different strategy params from basic validation, we don't need separate validation for this (aave and noun validations are also similar, will be removed in future)

jinanton:
Ooh, now i see, thx.

PR #438, opened 2021-12-01
Fix ranked choice bug with single strategy

samuveth:
There was a bug with the strategy calculation when the proposal only had a single strategy.


bonustrack (approved):
Seem to work well, thanks!

PR #442, opened 2021-12-07
Fix unhandled rejection "_signTypedData"

samuveth:
Fix https://discord.com/channels/707079246388133940/747049240022482974/917090820295909419

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -46,10 +46,8 @@ export default class Client {
   }
 
   async sign(web3: Web3Provider | Wallet, address: string, message, types) {
-    let signer;
     // @ts-ignore
-    if (web3?.getSigner) signer = web3.getSigner();
-    else signer = web3;
+    const signer = web3?.getSigner ? web3.getSigner() : web3;
     if (!message.from) message.from = address;
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -47,13 +47,14 @@ export default class Client {
 
   async sign(web3: Web3Provider | Wallet, address: string, message, types) {
     let signer;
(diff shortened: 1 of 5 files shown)


src/sign/index.ts
```diff
@@ -53,7 +53,7 @@ export default class Client {
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
     const data: any = { domain, types, message };
-    const sig = await signer._signTypedData(domain, data.types, message);
+    const sig = await signer?._signTypedData(domain, data.types, message);
```
ChaituVR, line 56: Why do we get signer undefined? 🤔 what should happen if it is undefined? 
If it is like this we send empty sig to the hub, right? maybe would be good to show some error message?

ChaituVR (changes requested):

src/sign/index.ts
samuveth: I can't reproduce signer undefined. Anytime I log the signer it's available. Not sure why it's showing this error.

samuveth: Ok can reproduce now.

src/sign/index.ts
```diff
@@ -47,8 +47,9 @@ export default class Client {
 
   async sign(web3: Web3Provider | Wallet, address: string, message, types) {
     let signer;
-    if (web3 instanceof Wallet) signer = web3;
-    if (web3 instanceof Web3Provider) signer = web3.getSigner();
+    // @ts-ignore
+    if (web3?.getSigner) signer = web3.getSigner();
+    else signer = web3;
```
bonustrack, line 52: Could be simplified to
```suggestion
  const signer = web3?.getSigner ? web3.getSigner() : web3;
```

bonustrack (approved):
Looks good!

ChaituVR (approved):

PR #462, opened 2022-01-06
[aavegotchi-agip] update strategy for agip 17 voting power

candoizo:
- adds voting power of aavegotchi realm parcels from agip17


ChaituVR:
Hi, @candoizo strategies moved to a different repo, can you plz raise a PR there? Thanks
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/aavegotchi-agip/index.ts

PR #468, opened 2022-01-17
Add IDchain network

misterplus:
Add IDchain network so we can start revising the brightid strategy.
The current node added in this PR is a temporary archive node setup on AWS, can someone make sure it works with snapshot?
We're deploying another node on Digital Ocean in a few days and I'll change the file to point to the new address later.


misterplus:
I replaced the multicall address with a verified one, you can check the source code here: https://explorer.idchain.one/address/0x41d289c966D51342A515A19dE9c27d16079c94E0/contracts

bonustrack:
Thanks! Can you check if the node works @ChaituVR ?

ChaituVR:
Hi, @misterplus I get the following error:
```
 Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.4.5)
 ```
 Also, i see an error that the node URL is not HTTPS

misterplus:
> Hi, @misterplus I get the following error:
> 
> ```
>  Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.4.5)
> ```
> 
> Also, i see an error that the node URL is not HTTPS

Sorry about that, the current setup on AWS is temporary and we should have a HTTPS node link within a few days.
Also would you please provide a method for checking a node is an operational archive node or not, so we could test it ourselves before commiting here? Saves you guys some time. 😄

ChaituVR:
@misterplus  we use this code https://github.com/snapshot-labs/networks/blob/main/src/composables/useApp.ts#L104-L164 to check archive node or not, get current block and send a multicall

ChaituVR:
We use this interface https://snapshot-networks.on.fleek.co/# (you can Click on any network and click the "edit network.json" button and add your network to JSON)

misterplus:
@ChaituVR I've added two https nodes (tested and confirmed working), please take a took!

bonustrack (approved):

PR #472, opened 2022-01-20
Add Moonbeam network

mar1:
Added Moonbeam network (chainID 1284)

PR #473, opened 2022-01-20
Adding Moonbeam logo

mar1:
Adding Moonbeam logo proper format


ChaituVR:
@mar1 Please add this image in #472  itself :)

PR #478, opened 2022-01-21
add emerald

moneytulip:
Fixes # .

-  Add Emerald to the networks.json
-  Add Emerald logo


moneytulip:
in case there's any concern, the multiswap contract isn't verified because of an issue with the block explorer.  But it's solidity compiler 0.8.8 with 5000 optimizations.  I also deployed another contract without optimizations at `0x008854eF4d0fB3E88BFC98c67F4700f4f702A363` if you'd prefer that.

bonustrack (approved):

PR #483, opened 2022-01-30
Add timeout to node requests

ChaituVR:


src/utils/provider.ts
```diff
@@ -7,8 +7,9 @@ import networks from '../networks.json';
 const providers = {};
 
 export default function getProvider(network: string) {
-  const url: string = networks[network].rpc[0];
-  if (!providers[network]) providers[network] = new StaticJsonRpcProvider(url);
+  const url: any = networks[network].rpc[0];
+  const connectionInfo = typeof url === 'object' ? {...url, timeout: 30000} : {url, timeout: 30000};
```
mktcode, line 11: I was curious what the default is but docs don't say. Found 120s here: https://github.com/ethers-io/ethers.js/blob/32c9a0976276245ad3191260157fb814e5c00080/packages/web/src.ts/index.ts#L110

mktcode (approved):

PR #490, opened 2022-02-03
Fix package publish workflow

ChaituVR:
image


bonustrack (approved):

mktcode:
Would be easier to just use one job. By using `needs` you can access a previous jobs outputs but other than that they are still pretty much isolated, aren't they? If you have to repeat `yarn install` in the second job, what about `yarn build` then?

PR #491, opened 2022-02-03
fix: validate and allow only valid ethereum address

bigint:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/252


bigint:
cc @ChaituVR

mktcode (approved):
Another very good catch! Thanks!

PR #493, opened 2022-02-04
[safesnap] Created function to fetch ENS text records

carlosfebres:
I added a function to get ENS text records following this recommendation from @mktcode https://github.com/snapshot-labs/snapshot/pull/1596#discussion_r798851057


mktcode (approved):
Looks good! Thanks!

PR #497, opened 2022-02-05
Remove duplicate build and yarn install from lockfile

ChaituVR:
As suggested in https://github.com/snapshot-labs/lock/pull/38

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -8,18 +8,7 @@ on:
     types: [created]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: 14
-      - run: yarn install
-      - run: yarn build
-
   publish-npm:
-    needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -8,26 +8,15 @@ on:
...
-    needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: 14
           registry-url: https://registry.npmjs.org/
-      - run: yarn install
+      - run: yarn install --frozen-lockfile
```
mktcode, line 19: Now build is missing.

```suggestion
      - run: yarn install --frozen-lockfile
      - run: yarn build
```

ChaituVR: Ya ya, I know, ignored as we already have `prepublishOnly` script which build before publishing https://github.com/snapshot-labs/snapshot.js/blob/master/package.json#L53

mktcode: Does it?
https://github.com/yarnpkg/berry/issues/2305
https://yarnpkg.com/advanced/lifecycle-scripts

(Didn't know that too... 🤫)

ChaituVR: Yea with yarn we may have to change it to `prepack + prepublish` that's why we are still using `npm publish` 

which work fine in our current workflow
https://github.com/snapshot-labs/snapshot.js/runs/5079073430?check_suite_focus=true
image

mktcode: Ah ok cool! Got it.

mktcode (approved):

PR #508, opened 2022-02-17
BSC node change and reverted back to Ankr node

ChaituVR:


bonustrack (approved):
I dont see any change of node

ChaituVR:
Ya ya node was changed in one version and reverted in one version

PR #515, opened 2022-02-25
Add HPB Mainnet Support

rjgeek:
Add HPB Mainnet Support

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -575,6 +575,24 @@
     "start": 379560,
     "imageIPFS": "QmSJEdToMvmjqoPvVq7awwdMeYKhXUYZMBiax9yHtFGMSq"
   },
+  "269": {
+    "key": "269",
+    "name": "High Performance Blockchain Mainnet",
+    "shortName": "HPB",
+    "chainId": 269,
+    "network": "mainnet",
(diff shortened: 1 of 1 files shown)


ChaituVR (changes requested):
Looks good but is there a way to increase limit of the node? (number of addresses we can pass into multicall contract) right now i can only pass around 300 addresses (ideally it should work when we pass 1000 addresses)

xueqianLu:
> Looks good but is there a way to increase limit of the node? (number of addresses we can pass into multicall contract) right now i can only pass around 300 addresses (ideally it should work when we pass 1000 addresses)

I have increase the limit of http request size on rpc server. You can retry pass 1000 addresses.

rjgeek:
> Looks good but is there a way to increase limit of the node? (number of addresses we can pass into multicall contract) right now i can only pass around 300 addresses (ideally it should work when we pass 1000 addresses)

I have increase the limit of http request size on rpc server. You can retry pass 1000 addresses.

ChaituVR:
HI @rjgeek This looks good, can you update the branch with latest changes, should be good to merge

ChaituVR (approved):

PR #517, opened 2022-02-25
Add Support for Multicall V2

abigger87:
## Overview

This PR adds support for [Makerdao's MulticallV2](https://github.com/makerdao/multicall/blob/master/src/Multicall2.sol).

## Changes Introduced

- MulticallV2 Contracts from MakerDAO's [Multicall Repo](https://github.com/makerdao/multicall) added to [networks.json](https://github.com/snapshot-labs/snapshot.js/blob/master/src/networks.json)
- `execute2` function added to the [Multicaller util](https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils/multicaller.ts) that executes the Multicall V2 [`tryBlockAndAggregate` function](https://github.com/makerdao/multicall/blob/master/src/Multicall2.sol#L67)
- `multicall2` function added to [`utils.ts`](https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts)

### DIFF
--- a/test/multicaller.ts
+++ b/test/multicaller.ts
@@ -40,3 +40,26 @@ multi.execute().then((result) => {
   }
   */
 });
+
+multi.execute2(undefined, true).then((result) => {
+  console.log('Multicaller v2 result', result);
+  /* Multicaller result
+  {
+    '0x6B175474E89094C44Da98b954EedeAC495271d0F': {
+      name: 'Dai Stablecoin',
+      symbol: 'DAI',
+      decimals: 18
+    },
+    '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2': {
+      name: 'Wrapped Ether',
+      symbol: 'WETH',
+      decimals: 18
+    },
+    '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599': {
+      name: 'Wrapped BTC',
+      symbol: 'WBTC',
+      decimals: 8
+    }
+  }
+  */
+});
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.3.51",
+  "version": "0.3.52",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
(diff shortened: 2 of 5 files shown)


test/multicaller.ts
```diff
@@ -40,3 +40,26 @@ multi.execute().then((result) => {
   }
   */
 });
+
+multi.execute2(undefined, true).then((result) => {
```
bonustrack, line 44: @abigger87 Sorry for slow review! I just tested this and for me it fail with this error:
```
Error: cannot estimate gas; transaction may fail or may require manual gas limit (error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":43,\"error\":{\"code\":-32000,\"message\":\"execution reverted\"}}\n"```

Are you able to reproduce this error? When execute function is called it reset the multicaller stored calls, you may need to add calls again or copy "multi" before calling execute()

abigger87: Nope, it's not obvious to me how to test/reproduce this. Can you share the command you ran?

abigger87: Will make your suggestions anyway!

bonustrack: @abigger87 To test it you'll need to run the `build` script then `node ./test/multicaller.ts`. When I test again now with the changes it still give me error. How do you test?

abigger87: I'm getting an unpredictable gas limit error when I run `node ./test/multicaller.ts` but don't see any environment variable specification anywhere for an RPC. Is there any other config I need to do to get this to run? Open to taking this to pms/dms I'm on TG @andreaslb or twitter @andreasbigger 

Here's the error:

Screen Shot 2022-03-04 at 12 51 22 PM

abigger87: Seems it's an issue with`networks.json` which I didn't change except for adding the multicall2 addresses?

abigger87: Ok fixed. Bad multicall address.
When you get a chance can you give it a shot @bonustrack - Am getting a call exception, but all the parameters seem correct to me 🤔 

Here's `Multicall2` on mainnet: https://etherscan.io/address/0x5ba1e12693dc8f9c48aad8770482f4739beed696#code

Screen Shot 2022-03-04 at 2 21 12 PM

PR #526, opened 2022-03-08
feat(): added a new uri-strict validation

midgerate:
Related to https://github.com/snapshot-labs/snapshot/pull/1881

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -176,6 +176,14 @@ export function validateSchema(schema, data) {
   const ajv = new Ajv({ allErrors: true, allowUnionTypes: true, $data: true });
   // @ts-ignore
   addFormats(ajv);
+  ajv.addFormat('uri-strict', {
+    type: 'string',
+    validate: (str) => {
+      return /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/.test(
+        str
+      );
+    }
+  });
   const validate = ajv.compile(schema);
   const valid = validate(data);
   return valid ? valid : validate.errors;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -176,14 +176,6 @@ export function validateSchema(schema, data) {
   const ajv = new Ajv({ allErrors: true, allowUnionTypes: true, $data: true });
   // @ts-ignore
   addFormats(ajv);
-  ajv.addFormat('uri-strict', {
-    type: 'string',
-    validate: (str) => {
-      return /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/.test(
-        str
-      );
-    }
-  });
   const validate = ajv.compile(schema);
   const valid = validate(data);
   return valid ? valid : validate.errors;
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -176,6 +176,14 @@ export function validateSchema(schema, data) {
   const ajv = new Ajv({ allErrors: true, allowUnionTypes: true, $data: true });
   // @ts-ignore
   addFormats(ajv);
+  ajv.addFormat('uri-strict', {
+    type: 'string',
+    validate: (str) => {
+      return /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/.test(
```
mktcode, line 182: Where is this regex coming from?

`([a-z\.]{2,6})` is looking for a 2-6 character top level domain but also allows dots?
It would match `http://a.b.c`.

```suggestion
      return /^(https?:\/\/)?([\da-z\.-]+)\.([a-z]{2,6})([\/\w \.-]*)*\/?$/.test(
```

midgerate: Came from copilot. 

So we cant allow `http://a.b.c` ? We should I think.

midgerate: Well, you certainly are. I have no idea how regex works. I will use your suggestion :D 

regex is something I aspire to learn

midgerate: the url format I saw somewhere was deprecated. I will try out url and see how that goes.

mktcode: A super smart and also very pretty, benevolent and still humble and likeable guy wrote an article for beginners about it, not so long ago: https://medium.com/geekculture/regular-expressions-for-non-programmers-ed2047d2181f

mktcode: Tbh... I'm also not much more than a beginner. ;) But I hope this article does it's job in showing that a little bit of knowledge can already be super helpful.

midgerate: Thanks! I am primed with the knowledge now. 

So our regex is simpler and I now understand what you mean. I also tested ajv's validation, however it seems to be doing a lot more which I do not understand. The problem with ajv is that it forces use of http/https protocol to be in the website URL.

samuveth (dismissed):

src/utils.ts
mktcode: Well, what else do we want to support? And of course a URL needs a protocol. I think users are used to it when entering a website and also we don't need to do any work when displaying the link. Just relying on a valid URL scheme makes more sense imho.

samuveth: > Well, what else do we want to support? And of course a URL needs a protocol. I think users are used to it when entering a website and also we don't need to do any work when displaying the link. You know... stuff like `'https://' + website.replace('http://', '').replace('https://', '')` or similar. Just relying on a valid URL scheme makes more sense imho. But if we allow things like `mywebsite.org` then maybe they should at least be prepended with `http://` when storing them in the DB.

Yes, I was also thinking we should always prepend and just use the build in URL validation from AJV

mktcode: Yea forget about that. Let's just use the built-in URL validation.

midgerate: @mktcode removed the new validator and used url validation.

samuveth (approved):

samuveth:
@ChaituVR this will need a new version after merged :)

PR #542, opened 2022-03-30
Throw error on subgraph when errors exist in response

ChaituVR:


mktcode (approved):
Approved. I'm just not sure under which version we should publish this. Technically this changes the behavior of the function, potentially breaking things in other applications, so it should be a new major release. (1.0.0) On the other hand, not so long ago, we didn't do any versioning at all.

Well.. I checked a few projects using snapshot.js and this function is being used as well: https://github.com/Kilros0817/staking/blob/5463d1b3fc9d7aed42d4a9c5baaa3151cd26ddc4/sections/gov/components/helper.ts#L13

Maybe it won't be a huge problem to just publish this as a new 0.3.x and we will probably have a lot of breaking changes still. So we'll end up at something like 25.10.23 quickly. :D But... so what?

PR #543, opened 2022-03-31
Add uri validation for discussion

samuveth:


samuveth:
@bonustrack why `but should accept any scheme like snapshot://123455` this?

mktcode (approved):

mktcode (approved):
actually... 256 is fine. :D

PR #546, opened 2022-04-06
Sub-DAO schema

mktcode:
...for the sub-dao feature.

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -228,6 +228,7 @@
         },
         "children": {
           "type": "array",
+          "maxItems": 8,
           "title": "children",
           "items": {
             "type": "string"
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -221,6 +221,17 @@
               "collector"
             ]
           }
+        },
+        "parent": {
+          "type": "string",
+          "title": "parent"
+        },
+        "children": {
+          "type": "array",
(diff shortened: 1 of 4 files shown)


ChaituVR:
Looks good but still draft PR

mktcode:
Yea I still need to add proper validation for ens names and not just allow any string.

mktcode:
> Yea I still need to add proper validation for ens names and not just allow any string.

Might actually be a bad idea. There are spaces like Uniswap without an ENS domain as space id.

src/schemas/space.json
```diff
@@ -221,6 +221,17 @@
               "collector"
             ]
           }
+        },
+        "parent": {
+          "type": "string",
+          "title": "parent"
+        },
+        "children": {
```
bonustrack, line 229: Can you enforce a max number of item here? I would recommend 8. This is just to avoid someone adding 1k space there and overload GraphQL

mktcode: done. what about validation now? maybe it can just stay like that and we perform a check in the settings page. the best solution would probably be to add a new validator that checks spaces existences based on a hub url param or so.

bonustrack: We dont need to validate that space exist here, but on the frontend we can make it that you can only select existing space.

bonustrack (approved):

PR #551, opened 2022-04-09
Add customUrl format validation and use for discussion

samuveth:
Fixes bug where we couldn't validate URL and empty string for discussion. Also add some URl types like `snapshot://` that we might want to use in the future.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -178,7 +178,8 @@ export function validateSchema(schema, data) {
   // @ts-ignore
   addFormats(ajv);
 
-  // Custom URL format
+  // Custom URL format for to allow empty string values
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -173,10 +173,28 @@ export async function getScores(
   }
 }
 
+function isUrl(str) {
+  if (!str.length) return true;
```
mktcode, line 177: How is an empty string a URL? Shouldn't it return false then?

samuveth: EIP712 needs it to be a string otherwise it will throw an error. Yeah I will try to inline it.

mktcode (approved):

PR #565, opened 2022-04-25
Don't pass limit option

### LINKED ISSUE #489: Rename limit option
The limit options should be renamed because we are passing the `options` object to the contract and `limit doesn't work in that case

https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L50-L63

```
cannot override "limit" (operation="overrides", overrides=["limit"], code=UNSUPPORTED_OPERATION, version=contracts/5.5.0)
```

ChaituVR:
Right now, we pass everything coming from options, and cases like `limit` , if we pass it, it fails with an error, so better to pass overrides here if we want to pass the overrides https://docs.ethers.io/v5/api/contract/contract/#Contract-functionsCall

Fixes #489


ChaituVR:
Looks like we are not passing limit anywhere, so merging this

PR #569, opened 2022-04-29
Change gnosis and arbi nodes

ChaituVR:
Changes from https://github.com/snapshot-labs/snapshot-score/blob/master/src/provider.ts#L4-L5


src/networks.json
```diff
@@ -439,8 +439,8 @@
     "network": "mainnet",
     "multicall": "0xb5b692a88bdfc81ca69dcb1d924f59f0413a602a",
     "rpc": [
-      "https://xdai-archive.blockscout.com",
       "https://gno.getblock.io/mainnet/?api_key=6c1d1e6e-75d9-452f-a863-a694bff93d5c",
+      "https://xdai-archive.blockscout.com",
```
bonustrack, line 443: We not continue with getblock, it shouldnt be the primaru RPC, don't we have a better one?

PR #581, opened 2022-05-17
fix: Supporting 1271 signatures not being 65 bytes long nor rsv compatible

0xdex18:
Originally, the call to /api/msg would throw "invalid signature string; must be 65 bytes" if the signature has a different length or throw rsv errors, if the 1271 sig is 65 bytes long.

The 1271 signatures of Ambire Wallet have different length and verifying a message would throw errors.

this PR solved this issue, ensuring message verification works on smart wallets like Ambire, Gnosis Safe, Argent, etc

### DIFF
--- a/src/sign/utils.ts
+++ b/src/sign/utils.ts
@@ -9,12 +9,19 @@ export function getHash(data) {
 
 export async function verify(address, sig, data) {
   const { domain, types, message } = data;
-  const recoverAddress = verifyTypedData(domain, types, message, sig);
+
   const hash = getHash(data);
   console.log('Hash', hash);
   console.log('Address', address);
-  console.log('Recover address', recoverAddress);
-  if (address === recoverAddress) return true;
+
+  try {
+    const recoverAddress = verifyTypedData(domain, types, message, sig);
+    console.log('Recover address', recoverAddress);
+    if (address === recoverAddress) return true;
+  } catch (e) {
+    console.log('Could not recoverAddress:' + e.message);
+  }
+
   console.log('Check EIP1271 signature');
   return await verifyEIP1271(address, sig, hash);
 }


src/sign/utils.ts
```diff
@@ -9,12 +9,19 @@ export function getHash(data) {
...
-  if (address === recoverAddress) return true;
+
+  try {
+    const recoverAddress = verifyTypedData(domain, types, message, sig);
+    console.log('Recover address', recoverAddress);
+    if (address === recoverAddress) return true;
+  } catch (e) {
+    console.log('Could not recoverAddress:' + e.message);
+  }
+
```
ChaituVR, line 24: It can catch all other unknown errors too?

0xdex18: Well, there is a lot going on in verifyTypedData, and I would classify it in 2 categories

1 : (This is the error we should catch) the signature validation category, checking things like 
  - ecrecover
  - signature length
  - rsv matching
   
2 : the 712 data validation like checking if the domain / types / message are 712 compliant

what matters here is to catch the 1st category as ecrecover, siglen and rsv matching would fail in the case of 1271

it also however catches the typed data compliance error, which you might want to throw.


As I wanted to keep the PR as small as possible, I kept your original call to verifyTypedData, wrapping it in a try catch but if you do care about throwing a relevant error related to the 712 compliant data and separate the 2 types of errors,

I would suggest something like that:

```js
// first, checking the data structure (which is independent from the sig)
try {
TypedDataEncoder.hash(domain , types, message)
} catch (e) {
 throw Error('Invalid 712 data structure')
}

// second, continue with the original process, matching the message AND the signature
try {
  const recoverAddress = verifyTypedData(domain, types, message, sig);
  console.log('Recover address', recoverAddress);
  if (address === recoverAddress) return true;
} catch (e) {
  console.log('Could not recoverAddress:' + e.message);
}
```

src/sign/utils.ts
```diff
@@ -9,12 +9,19 @@ export function getHash(data) {
...
   console.log('Address', address);
-  console.log('Recover address', recoverAddress);
-  if (address === recoverAddress) return true;
+
+  try {
+    const recoverAddress = verifyTypedData(domain, types, message, sig);
+    console.log('Recover address', recoverAddress);
+    if (address === recoverAddress) return true;
+  } catch (e) {
+    console.log('Could not recoverAddress:' + e.message);
```
bonustrack, line 22:
```suggestion
    console.log('Could not recover address', e);
```

bonustrack (approved):
Looks good!

PR #598, opened 2022-05-31
Fuse node change

ChaituVR:


metrox-eth (approved):

DefiDebauchery:
Just a note that the RPC endpoint was added to HECO and not Fuse.

PR #622, opened 2022-06-21
Refactor voting and add tests

samuveth:
Changes:
- Replace `jest` with `vitest`
- Add example proposals for each voting type
- Add tests for each voting type
- Add a new filter method `filterVotesWithInvalidChoice` to each voting type
- Add typescript to all voting types

Wrote tests for all voting types before making any changes to any of the classes. Also compared the results in this version with live version results and everything should be as before. Some older (~3 months and older) ranked choice voting proposals might have results change, but I checked more recent ones stay unchanged.

Additionally there was a bug on several voting types where the results would show the wrong amount for each strategy e.g: https://snapshot.org/#/ens.eth/proposal/0x3fa1a972b7d40510f5c1fb0d1c43e41960866a35db75a1541572451fa96f7eab

Before: 
image

After: 
image


src/voting/approval.ts
```diff
@@ -1,8 +1,24 @@
...
+  return votes.filter((vote) => {
+    return (
+      Array.isArray(vote.choice) &&
+      // If choice index is not in choices, return false
+      vote.choice.every((choice) => choices?.[choice - 1] !== undefined) &&
+      // If any choice is duplicated, return false
+      vote.choice.length === new Set(vote.choice).size
+    );
+  });
+}
```
samuveth, line 11: @bonustrack @mktcode please let me know if you can think any cases that aren't covered here for approval voting.

src/voting/rankedChoice.ts
```diff
@@ -41,64 +71,81 @@ function irv(ballots, rounds) {
...
+    return this.votes.filter((vote) =>
+      isValidChoice(vote.choice, this.proposal.choices)
+    );
+  }
+
+  getScores(): number[] {
+    const finalRounde = getFinalRound(this.getValidVotes());
     return this.proposal.choices.map((choice, i) =>
-      getFinalRound(i, this.votes).reduce((a, b: any) => a + b[1][0], 0)
+      finalRounde
```
bonustrack, line 127: Rounde? Shouldnt this be round?

samuveth: Fixed

bonustrack (approved):
Looks good

PR #669, opened 2022-07-27
add nova network

nova-network-inc:
Additions # .

- Add Nova Network

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -413,7 +413,6 @@
     "start": 4992871,
     "logo": "ipfs://QmSZvT9w9eUDvV1YVaq3BKKEbubtNVqu1Rin44FuN4wz11"
   },
-    },
   "87": {
     "key": "87",
     "name": "Nova Network",
--- a/src/networks.json
+++ b/src/networks.json
@@ -413,6 +413,21 @@
     "start": 4992871,
     "logo": "ipfs://QmSZvT9w9eUDvV1YVaq3BKKEbubtNVqu1Rin44FuN4wz11"
   },
+    },
+  "87": {
+    "key": "87",
+    "name": "Nova Network",
+    "shortName": "Nova",
+    "chainId": 87,
+    "network": "mainnet",
+    "multicall": "0x7A5a7579eb8DdEd352848cFDD0a5530C4e56FF7f",
+    "rpc": [
+      "https://dev.rpc.novanetwork.io/"
+    ],
+    "explorer": "https://explorer.novanetwork.io/",
+    "start": 1000000,
+    "logo": "ipfs://QmTTamJ55YGQwMboq4aqf3JjTEy5WDtjo4GBRQ5VdsWA6U"
+  },
   "97": {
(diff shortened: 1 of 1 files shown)


src/networks.json
```diff
@@ -413,6 +413,20 @@
     "start": 4992871,
     "logo": "ipfs://QmSZvT9w9eUDvV1YVaq3BKKEbubtNVqu1Rin44FuN4wz11"
   },
+  "87": {
+    "key": "87",
+    "name": "Nova Network",
+    "shortName": "Nova",
+    "chainId": 87,
+    "network": "mainnet",
+    "multicall": "0x7A5a7579eb8DdEd352848cFDD0a5530C4e56FF7f",
```
ChaituVR, line 422: Is there a way to verify this contract?

src/networks.json
```diff
@@ -413,6 +413,20 @@
...
+    "rpc": [
+      "https://dev.rpc.novanetwork.io/"
+    ],
+    "explorer": "https://explorer.novanetwork.io/",
+    "start": 1000000,
```
ChaituVR, line 427: should be block number at multicall contract creation

ChaituVR (changes requested):

src/networks.json
nova-network-inc: Deployed a new contract and changed the block number.

src/networks.json
nova-network-inc: You can have a look here https://github.com/nova-network-inc/fusion-public-contracts/blob/main/NovaMulticall.sol the explorer isn't indexed, so doesn't support verification just yet. Working through that.

nova-network-inc:
All concerns addressed. Please let me know if there's anything else you might need 😃

ChaituVR:
Looks good, except that we need to verify multicall contract through explorer

nova-network-inc:
> Looks good, except that we need to verify multicall contract through explorer

Okay, https://novascan.io/ is indexed and verification via CLI, I'll do it and send the link through later on today.

nova-network-inc:
> Looks good, except that we need to verify multicall contract through explorer

I had to redeploy it to verify, and the explorer has a UI glitch which prevents it to load the artifact, but you will see the contract verified and all the calls in there.

https://novascan.io/address/0x07f6b55a5dae24ac8011cc2cb815de8316e2b965?tab=contract

Let me know if anything else is needed.

ChaituVR (approved):

ChaituVR:
Hey @nova-network-inc 

Nova network's node returns this error when used on our server, 
```
Error: unable to verify the first certificate
```
but works fine from any front end, any idea?

nova-network-inc:
> Hey @nova-network-inc
> 
> Nova network's node returns this error when used on our server,
> 
> ```
> Error: unable to verify the first certificate
> ```
> 
> but works fine from any front end, any idea?

It was a problem with the RPC's proxy. It has been fixed now. Thanks @ChaituVR

PR #674, opened 2022-08-02
Add timeperiod validation for proposal entries

stevef:
Signed-off-by: Steve Finkelstein <sf@stevefink.net>

- Add timeperiod validation for proposal entries


mktcode:
Looks good. FYI: We're [thinking](https://github.com/snapshot-labs/snapshot.js/pull/676) about moving the validations to the [strategies](https://github.com/snapshot-labs/snapshot-strategies) repo. But I think we can merge this first and then move.

mktcode:
I'll merge this now. Later we'll add the validation functions to the [strategies repo](https://github.com/snapshot-labs/snapshot-strategies) but we leave them here as well for backwards compatibility / to avoid a major version update for now.

mktcode (approved):

PR #676, opened 2022-08-03
fix circular dependencies

### LINKED ISSUE #544: Circular dependencies
I want to use my local copy of snapshot.js when working on the hub and/or UI. I use `yarn link` for that but I get this error:

image

Re-building it or running dev mode doesn't work. But what I noticed is this:

image

Circular dependencies... Well, that doesn't sound good. Some more searching later:

image

I checked out some [older commits](https://github.com/snapshot-labs/snapshot.js/commit/1f4c55eab605d525ff246aa8729666b821f84ec0) and it is like that for more than a year... at least. Imho already far too long and we just build up more and more technical debt. This should be fixed asap.

mktcode:
Fixes https://github.com/snapshot-labs/snapshot.js/issues/544

- move `voting` and `validation` from `utils` to top-level/index

In the frontend we import from the `src` directory directly (e.g. `'@snapshot-labs/snapshot.js/src/voting/quadratic'`) but score and hub use `snapshot.utils.`. This needs to be adjusted when updating snapshot.js over there. So technically we should bump version to `v1.0.0`.

Probably the remaining circular dependencies should be fixed too.

```
src/utils.ts -> src/utils/multicaller.ts -> src/utils.ts
src/utils.ts -> src/utils/blockfinder.ts -> src/utils.ts
src/utils.ts -> src/sign/utils.ts -> src/sign/eip1271.ts -> src/utils.ts
```

But since this PR already fixes #544 I don't really want to go down that rabbit hole now. Not this month at least.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,10 +2,12 @@ import Client712 from './sign';
 import Client from './client';
 import schemas from './schemas';
 import utils from './utils';
+import voting from './voting';
 
 export default {
   Client,
   Client712,
   schemas,
-  utils
+  utils,
+  voting
 };
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import Client712 from './sign';
 import Client from './client';
 import schemas from './schemas';
 import utils from './utils';
+import validations from './validations';
 import voting from './voting';
 
 export default {
@@ -9,5 +10,6 @@ export default {
   Client712,
   schemas,
   utils,
+  validations,
   voting
 };
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,7 +13,6 @@ import { signMessage, getBlockNumber } from './utils/web3';
 import { getHash, verify } from './sign/utils';
 import gateways from './gateways.json';
 import networks from './networks.json';
-import voting from './voting';
 
 interface Options {
   url?: string;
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.4.13",
+  "version": "1.0.0",
```
bonustrack, line 3: Can we keep v1 for a big refactoring?

bonustrack:
I think we need the validations in snapshot-strategies repo, snapshot.js should only contain code from us

mktcode:
Yep. Couldn't agree more. Although in regards to #544 it's the `voting` causing the problem. I would then maybe start by adding the validations to the strategy repo, update frontend and hub, and then remove validations here and merge the PR.

> Can we keep v1 for a big refactoring?

Not sure what a big refactoring would exactly mean. But in terms of semantic versioning it doesn't really matter. It's a convention to exactly not have such discussion actually. If the update contains a breaking change -> major version update. If it only contains new, non-breaking features -> minor version update. Bugfixes and code cleanup -> patch version update. The only problem with semver is that there's no guarantee that it's strictly followed. But I don't want to contribute to that problem.

PR #683, opened 2022-08-09
Don't remove quadratic and weighted votes with zero values

samuveth:


src/voting/quadratic.ts
```diff
@@ -46,7 +46,7 @@ export default class QuadraticVoting {
       Object.keys(voteChoice).length > 0 &&
       // If voteChoice object values are not a positive integer, return false
```
ChaituVR, line 47:
```suggestion
      // If voteChoice object values are a negative number, return false
```

src/voting/weighted.ts
```diff
@@ -46,7 +46,7 @@ export default class WeightedVoting {
       Object.keys(voteChoice).length > 0 &&
       // If voteChoice object values are not a positive integer, return false
```
ChaituVR, line 47:
```suggestion
      // If voteChoice object values are negative number, return false
```

ChaituVR (approved):

PR #692, opened 2022-08-16
Delegation subgraphs as JSON and add op and goerli delegation support

ChaituVR:


ChaituVR:
Merging this @bonustrack we could remove `SNAPSHOT_SUBGRAPH_URL` but may have to look for ways to export JSON to the `dist` folder but we may break other's code, so for now leaving it like this

PR #743, opened 2022-10-15
Enforce Checksum on Sign

### LINKED ISSUE #712: Enforce checksum address on sign
- I think you can just do it here https://github.com/snapshot-labs/snapshot.js/blob/master/src/sign/index.ts#L55
- using https://docs.ethers.io/v5/api/utils/address/#utils-getAddress from https://www.npmjs.com/package/@ethersproject/address
- We have to check all methods in this class (vote/proposal/settings etc. ) should make sure they work fine even if we pass an address in any case
- To run these methods you can follow https://docs.snapshot.org/snapshot.js

emanuel-sol:
Fixes #712 .

- Add ethers/address submodule in order to use getAddress function
- Preprocess address input by lowercasing it to prevent runtime errors

I was able to successfully test the changes in the snapshot UI locally using `npm link`, but I'm having a difficult time mocking requests with malformed case addresses. I could use some testing guidance here.

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -54,11 +54,7 @@ export default class Client {
     // @ts-ignore
     const signer = web3?.getSigner ? web3.getSigner() : web3;
     const checksumAddress = getAddress(address.toLowerCase());
-    if (message.from) {
-      message.from = getAddress(message.from.toLowerCase());
-    } else {
-      message.from = checksumAddress;
-    }
+    message.from = message.from ? getAddress(message.from.toLowerCase()) : checksumAddress
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
     const data: any = { domain, types, message };
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -1,7 +1,7 @@
 import fetch from 'cross-fetch';
 import { Web3Provider } from '@ethersproject/providers';
 import { Wallet } from '@ethersproject/wallet';
-import { getAddress } from '@ethersproject/address'
+import { getAddress } from '@ethersproject/address';
 import {
   Space,
   Proposal,
@@ -54,7 +54,9 @@ export default class Client {
     // @ts-ignore
     const signer = web3?.getSigner ? web3.getSigner() : web3;
     const checksumAddress = getAddress(address.toLowerCase());
-    message.from = message.from ? getAddress(message.from.toLowerCase()) : checksumAddress
+    message.from = message.from
+      ? getAddress(message.from.toLowerCase())
+      : checksumAddress;
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
     const data: any = { domain, types, message };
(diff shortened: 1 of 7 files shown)


ChaituVR (changes requested):

src/sign/index.ts
```diff
@@ -52,13 +53,14 @@ export default class Client {
...
+    const checksumAddress = getAddress(address.toLowerCase());
+    if (!message.from) message.from = checksumAddress;
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
     const data: any = { domain, types, message };
     const sig = await signer._signTypedData(domain, data.types, message);
-    // console.log('Sign', { address, sig, data });
-    return await this.send({ address, sig, data });
+    // console.log('Sign', { checksumAddress, sig, data });
+    return await this.send({ checksumAddress, sig, data });
```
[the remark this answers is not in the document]
emanuel-sol: Ah yes, I forgot ES6 object destructuring would change the property name. Snapshot-Hub would reject the promise as defined [here](https://github.com/snapshot-labs/snapshot-hub/blob/e810bffb72b20db38fd4bb727f5a3773c49df4f6/src/ingestor/typedData/index.ts#L18). I hard coded the property name to be `address` as a fix.

src/sign/index.ts
```diff
@@ -52,13 +53,14 @@ export default class Client {
   async sign(web3: Web3Provider | Wallet, address: string, message, types) {
     // @ts-ignore
     const signer = web3?.getSigner ? web3.getSigner() : web3;
-    if (!message.from) message.from = address;
+    const checksumAddress = getAddress(address.toLowerCase());
+    if (!message.from) message.from = checksumAddress;
```
ChaituVR, line 57: I feel we should do checksum for `message.from` too.

src/sign/index.ts
```diff
@@ -52,13 +53,14 @@ export default class Client {
   async sign(web3: Web3Provider | Wallet, address: string, message, types) {
     // @ts-ignore
     const signer = web3?.getSigner ? web3.getSigner() : web3;
-    if (!message.from) message.from = address;
+    const checksumAddress = getAddress(address.toLowerCase());
```
bonustrack, line 56: Why toLowercase here?

src/sign/index.ts
bonustrack: Of course

bonustrack: This checksum message.from only if its not part of the message, if its part of the message it wont be checksumed you should change that

src/sign/index.ts
emanuel-sol: I used the lowercase function for basic error checking in case we receive a malformed checksum address

bonustrack: In which case does it throw and error here? I dont think toLowercase change anything here

src/sign/index.ts
```diff
@@ -52,13 +53,18 @@ export default class Client {
...
+    if (message.from) {
```
bonustrack, line 57: This could be made much simpler

src/sign/index.ts
emanuel-sol: Yeah it's probably a good idea if we throw an error if we get a bad checksum address. I'll get rid of it

ChaituVR (approved):
Tested ACK

PR #762, opened 2022-12-01
Add template and change uri to customUrl

samuveth:
Part of https://github.com/snapshot-labs/snapshot/issues/3195

Also changed all occurences of `uri` to our `customUrl` type. And changed guidelines from simple string to `customUrl`


ChaituVR (approved):

samuveth:
@ChaituVR any idea why I still can't save settings with a `template`? Saving with `guidelines` works well though.

PR #764, opened 2022-12-04
feat: replace client endpoint to sequencer

bonustrack:


src/constants.json
```diff
@@ -1,3 +1,10 @@
 {
-  "version": "0.1.3"
+  "livenet": {
+    "hub": "https://hub.snapshot.org",
```
ChaituVR, line 3: I get `snapshot.min.js:15          POST https://hub.snapshot.org/ 404` error when i try to vote @bonustrack

ChaituVR (changes requested):

src/constants.json
ChaituVR: I use same code as here https://docs.snapshot.org/snapshot.js#init-client

bonustrack:
Can you debug this PR @ChaituVR ?

ChaituVR (commented):
There was a issue with `.replace` so it was not working. fixed it now 👍

ChaituVR (approved):

PR #776, opened 2023-02-03
feat: Add get owner and controller helpers

samuveth:
Related to https://github.com/snapshot-labs/snapshot/issues/3472

Add `getEnsOwner` and `getSpaceController` helper functions for Ui and Hub.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -311,7 +311,7 @@ export async function getEnsOwner(ens: string): Promise<string | null> {
   const provider = getProvider('1');
   const ensRegistry = new Contract(
     registryAddress,
-    'function owner(bytes32) view returns (address)',
+   ['function owner(bytes32) view returns (address)'],
     provider
   );
   return await ensRegistry.owner(ens);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -306,12 +306,15 @@ export async function getSpaceUri(
   }
 }
 
-export async function getEnsOwner(ens: string): Promise<string | null> {
+export async function getEnsOwner(
+  ens: string,
+  network = '1'
+): Promise<string | null> {
   const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
-  const provider = getProvider('1');
+  const provider = getProvider(network);
   const ensRegistry = new Contract(
     registryAddress,
-   ['function owner(bytes32) view returns (address)'],
+    ['function owner(bytes32) view returns (address)'],
     provider
   );
   const ensHash = hash(normalize(ens));
@@ -333,7 +336,7 @@ export async function getSpaceController(
     isUriAddress = isAddress(address);
     if (isUriAddress) return address;
   }
-  return await getEnsOwner(id);
+  return await getEnsOwner(id, network);
 }
 
 export async function getDelegatesBySpace(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -314,7 +314,8 @@ export async function getEnsOwner(ens: string): Promise<string | null> {
    ['function owner(bytes32) view returns (address)'],
     provider
   );
-  return await ensRegistry.owner(ens);
+  const ensHash = hash(normalize(ens));
+  return await ensRegistry.owner(ensHash);
 }
 
 export async function getSpaceController(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -306,12 +306,15 @@ export async function getSpaceUri(
   }
 }
 
-export async function getEnsOwner(
-  ens: string,
-  network = '1'
-): Promise<string | null> {
-  const provider = getProvider(network);
-  return provider.resolveName(ens);
+export async function getEnsOwner(ens: string): Promise<string | null> {
+  const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
+  const provider = getProvider('1');
+  const ensRegistry = new Contract(
+    registryAddress,
+    'function owner(bytes32) view returns (address)',
+    provider
+  );
+  return await ensRegistry.owner(ens);
 }
 
 export async function getSpaceController(
@@ -329,7 +332,7 @@ export async function getSpaceController(
     isUriAddress = isAddress(address);
     if (isUriAddress) return address;
   }
-  return await getEnsOwner(id, network);
+  return await getEnsOwner(id);
 }
 
 export async function getDelegatesBySpace(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,7 @@
 import fetch from 'cross-fetch';
 import { Interface } from '@ethersproject/abi';
 import { Contract } from '@ethersproject/contracts';
+import { isAddress } from '@ethersproject/address';
 import { hash, normalize } from '@ensdomains/eth-ens-namehash';
 import { jsonToGraphQLQuery } from 'json-to-graphql-query';
 import Ajv from 'ajv';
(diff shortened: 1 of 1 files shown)


ChaituVR:
Waiting for clarity from @bonustrack

src/utils.ts
```diff
@@ -293,14 +294,43 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
 
+export async function getEnsOwner(
+  ens: string,
+  network = '1'
+): Promise<string | null> {
+  const provider = getProvider(network);
+  return provider.resolveName(ens);
+}
+
+export async function getSpaceController(
```
bonustrack, line 316: What if the text record is an IPFS or https URL?

src/utils.ts
```diff
@@ -293,14 +294,43 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
-export async function getSpaceUri(id, network = '1') {
+export async function getSpaceUri(
+  id: string,
+  network = '1'
+): Promise<string | null> {
   try {
     return await getEnsTextRecord(id, 'snapshot', network);
   } catch (e) {
-    return false;
+    return null;
```
ChaituVR, line 304:
```suggestion
    console.log(e);
    return null;
```

src/utils.ts
```diff
@@ -293,14 +294,43 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
+    let isUriAddress = isAddress(spaceUri);
+    if (isUriAddress) return spaceUri;
+
+    const uriParts = spaceUri.split('/');
+    const position = uriParts.includes('testnet') ? 5 : 4;
+    const address = uriParts[position];
+    isUriAddress = isAddress(address);
+    if (isUriAddress) return address;
+  }
+  return await getEnsOwner(id, network);
```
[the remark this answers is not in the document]
samuveth: @ChaituVR I would prefer to just keep `snapshot`, like @bonustrack said here https://github.com/snapshot-labs/snapshot/issues/3472#issuecomment-1416024704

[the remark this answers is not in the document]
samuveth: Is there any reason we should use a new separate key?

samuveth:
@ChaituVR can we merge this?

src/utils.ts
```diff
@@ -293,14 +294,44 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
+    return null;
   }
 }
 
+export async function getEnsOwner(
+  ens: string,
+  network = '1'
+): Promise<string | null> {
+  const provider = getProvider(network);
+  return provider.resolveName(ens);
```
ChaituVR, line 314: Hi @samuveth `resolveName` will return the owner of the domain right? or controller? 
For many domains the controller is already set to a different address, for example, https://app.ens.domains/name/gitcoindao.eth/details We want to check the controller instead of the owner maybe.?

samuveth: @ChaituVR by owner you mean Registrant right? The `resolveName` will always return the Controller of the domain.

ChaituVR: It returns the ETH address. (marked in red) we want to get controller (marked in blue) right?
Untitled

if there is no ETH address set (example ensdao.eth) it returns null

samuveth: Hm interesting. But this is how we have handled so far on the frontend. I couldn't find any way in their docs to get the controller specifically @ChaituVR

samuveth: @ChaituVR 

> if there is no ETH address set (example ensdao.eth) it returns null

This only seems to be the case for `lookupAddress`. https://docs.ethers.org/v5/single-page/#/v5/api/providers/provider/-%23-Provider-lookupAddress

But `resolveName` will always return an address if the name is owned by someone. https://docs.ethers.org/v5/single-page/#/v5/api/providers/provider/-%23-Provider-ResolveName

samuveth: The other solution would be to directly get it from the contract like this 

```js
// Create a new instance of the ENS registry
const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
const ensRegistry = new ethers.Contract(registryAddress, ['function resolver(bytes32) view returns (address)', 'function owner(bytes32) view returns (address)'], provider);

// Get the resolver of the ENS address
const ensName = 'example.eth'; // Replace with your ENS name
const nodeHash = ethers.utils.namehash(ensName);
const resolverAddress = await ensRegistry.resolver(nodeHash);

// Get the controller of the ENS address from the resolver
const resolverAbi = ['function getController(address) view returns (address)'];
const resolver = new ethers.Contract(resolverAddress, resolverAbi, provider);
const controller = await resolver.getController(nodeHash);

```

ChaituVR: Not sure what's best but I feel getting it from resolver should be good I think, let's wait for @bonustrack :)

samuveth:
@ChaituVR does this work for you?

ChaituVR:
Fabien confirmed that we can go with the ENS owner.?

src/utils.ts
```diff
@@ -293,14 +294,47 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
 
+export async function getEnsOwner(ens: string): Promise<string | null> {
+  const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
+  const provider = getProvider('1');
+  const ensRegistry = new Contract(
+    registryAddress,
+    'function owner(bytes32) view returns (address)',
+    provider
+  );
+  return await ensRegistry.owner(ens);
```
ChaituVR, line 317: We should pass a ENS like here https://github.com/snapshot-labs/snapshot.js/blob/83f502497f3351d9e6047a047fa4c4aee919cfd0/src/utils.ts#L292 no? 🤔 It works this way? will run a test

ChaituVR: Yep not working 🤔 It should be like 
```suggestion
  const ensHash = hash(normalize(ens));
  return await ensRegistry.owner(ensHash);
```

src/utils.ts
```diff
@@ -293,14 +294,47 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
+    return null;
   }
 }
 
+export async function getEnsOwner(ens: string): Promise<string | null> {
+  const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
+  const provider = getProvider('1');
+  const ensRegistry = new Contract(
+    registryAddress,
+    'function owner(bytes32) view returns (address)',
```
ChaituVR, line 314: It throws an error here 
```suggestion
   ['function owner(bytes32) view returns (address)'],
```

src/utils.ts
```diff
@@ -293,14 +294,47 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
   try {
     return await getEnsTextRecord(id, 'snapshot', network);
   } catch (e) {
-    return false;
+    console.log(e);
+    return null;
   }
 }
 
+export async function getEnsOwner(ens: string): Promise<string | null> {
```
ChaituVR, line 309: So this function returns the ENS controller. not owner/registrant 😄 we accidentally solved our problem?

ChaituVR: Example try to call this function for: https://app.ens.domains/name/abc.eth/details

samuveth: Okay then if that the case I guess the controller is fine :D

ChaituVR: it was not reading controllers from goerli. fixed it

src/utils.ts
```diff
@@ -306,12 +306,15 @@ export async function getSpaceUri(
   }
 }
 
-export async function getEnsOwner(ens: string): Promise<string | null> {
+export async function getEnsOwner(
+  ens: string,
+  network = '1'
+): Promise<string | null> {
   const registryAddress = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
```
samuveth, line 313: @ChaituVR is goerli the same registry address?

src/utils.ts
```diff
@@ -293,14 +294,51 @@ export function getEnsTextRecord(ens: string, record: string, network = '1') {
...
+  id: string,
+  network = '1'
+): Promise<string | null> {
+  const spaceUri = await getSpaceUri(id, network);
+  if (spaceUri) {
+    let isUriAddress = isAddress(spaceUri);
+    if (isUriAddress) return spaceUri;
+
+    const uriParts = spaceUri.split('/');
+    const position = uriParts.includes('testnet') ? 5 : 4;
```
ChaituVR, line 334: I think better to check the IPNS format as before right? not required?

src/utils.ts
ChaituVR: Yep same one

src/utils.ts
ChaituVR: Maybe not required. Let's merge this then? works fine to me, not sure if i miss some case 😅

samuveth: Yes can merge, should cover all cases as this is how we have been doing it on the frontend too

ChaituVR (approved):

PR #788, opened 2023-03-04
Using local hub and sequencer

ChaituVR:
It is hard to use the local hub and sequencer in UI

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -55,6 +55,10 @@ export default class Client {
       'https://testnet.snapshot.org',
       constants.testnet.sequencer
     );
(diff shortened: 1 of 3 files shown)


bonustrack (approved):

src/sign/index.ts
```diff
@@ -55,6 +55,10 @@ export default class Client {
       'https://testnet.snapshot.org',
       constants.testnet.sequencer
     );
+    address = address.replace(
+      'https://localhost:3000',
```
bonustrack, line 59: Maybe we can do like this, even better (on others env too)
```suggestion
      constants.local.hub,
```

PR #790, opened 2023-03-06
Return choice as valid if choice is encrypted

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/3234
Sequencer changes: https://github.com/snapshot-labs/snapshot-sequencer/pull/23

------
- Accepts new param if a choice is encrypted or not
- If the choice is encrypted, count it as a valid choice
- else continue doing the same as before
- In the sequencer, if the shutter is enabled and the proposal is active, we pass true, else false

TODO:
- [ ] Test everything once again
- [ ] Test all voting types

### DIFF
--- a/src/voting/approval.ts
+++ b/src/voting/approval.ts
@@ -12,7 +12,7 @@ export default class ApprovalVoting {
     votes: ApprovalVote[],
     strategies: Strategy[],
     selected: number[],
-    options: Options = { shutter: false }
+    options: Options = { encryptedChoice: false }
   ) {
     this.proposal = proposal;
     this.votes = votes;
@@ -24,9 +24,13 @@ export default class ApprovalVoting {
   static isValidChoice(
     voteChoice: number[] | string,
     proposalChoices: string[],
-    shutter = false
+    encryptedChoice = false
   ): boolean {
-    if (shutter && typeof voteChoice === 'string' && voteChoice.length > 0) {
+    if (
+      encryptedChoice &&
+      typeof voteChoice === 'string' &&
+      voteChoice.length > 0
+    ) {
       return true;
     }
     return (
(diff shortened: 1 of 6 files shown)


bonustrack:
This mean all shutter votes are counted toward total score but on others proposals not? Not sure I like this approach

samuveth:
@ChaituVR, I agree with Fabien. Like discussed on the call, I think we should always count invalid votes towards the total score, not matter if Shutter or not.

ChaituVR:
I think maybe my changes confused you, would make more sense if you refer https://github.com/snapshot-labs/snapshot-sequencer/pull/23  I renamed variable names from `shutter` to `encryptedChoice` now to make it easy to understand

After spending some time on this, I believe that this approach is much better
- It simply assumes if a choice is encrypted, it is valid
- These changes are only for active proposals using `shutter`, if the proposal is closed everything should work the same as before and will be consistent across all proposals
- We will count the vote with Invalid choices toward the total vote count but not in the final score
- Similar to override strategies, only the proposal's scores are updated at the end of the proposal

I believe that this approach is **simple** and straightforward. Counting invalid choices toward the total score is tricky, and we would need to explain the results to everyone, especially with quadratic voting and weighted voting. There are a lot of things that we need to handle

If we still want to go this way, I would recommend that @samuveth handle the changes in snapshot.js as he built these voting systems.

samuveth:
> and we would need to explain the results to everyone, especially with quadratic voting and weighted voting. There are a lot of things that we need to handle

The results will not change if we handle it like I did in https://github.com/snapshot-labs/snapshot.js/pull/791/files#diff-9716c0d490798915599efd18f6e9204675c990cfd0da94b90104a5ab0d26676b. The scoresTotal will only be used when calculating the percentages of each choice on the Ui. For that we can simply show another bar in the results for invalid votes if there are any

PR #791, opened 2023-03-07
Count all votes to scoresTotal

samuveth:
Related to https://github.com/snapshot-labs/snapshot-sequencer/pull/23

### DIFF
--- a/src/voting/weighted.ts
+++ b/src/voting/weighted.ts
@@ -68,9 +68,14 @@ export default class WeightedVoting {
         .reduce((a, b: any) => a + b, 0)
     );
 
+    const validScoresTotal = this.getValidVotes().reduce(
+      (a, b: any) => a + b.balance,
+      0
+    );
+
     return results
       .map((res, i) => percentageOfTotal(i, results, results))
-      .map((p) => (this.getScoresTotal() / 100) * p);
+      .map((p) => (validScoresTotal / 100) * p);
   }
 
   getScoresByStrategy(): number[][] {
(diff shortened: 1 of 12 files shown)


src/voting/weighted.ts
```diff
@@ -84,18 +89,23 @@ export default class WeightedVoting {
       )
       .map((arr) => arr.map((pwr) => [pwr]));
 
+    const validScoresTotal = this.getValidVotes().reduce(
+      (a, b: any) => a + b.balance,
+      0
+    );
```
ChaituVR, line 95: The whole problem is with this method `getValidVotes`, now `getScoresByStrategy` and `getScores` return different scores before and after the proposal end while using shutter,

ChaituVR (changes requested):

src/voting/weighted.ts
ChaituVR: I mean with the shutter it is just strings. it may not work with these changes too. if we ignore strings, it changes scores before and after `end`

ChaituVR: With my changes it won't `getValidVotes ` and `getScoresByStrategy` return zero values only `getScoresTotal` changes once proposal is closed

samuveth: Not really sure what your point is to be honest. Maybe the method names are a bit confusing?

bonustrack: This PR is what I was thinking in our call, not sure I understand what you mean @ChaituVR, for Shutter we can use just the function `getScoresTotal`

ChaituVR: Yep, thanks guys. I now feel completely confused about this 😆 Can ignore my comments

ChaituVR (approved):
utACK

PR #800, opened 2023-03-26
Remove try catch in verify function

ChaituVR:


src/sign/utils.ts
```diff
@@ -14,13 +14,9 @@ export async function verify(address, sig, data, network = '1') {
...
   // console.log('Address', address);
 
-  try {
-    const recoverAddress = verifyTypedData(domain, types, message, sig);
-    // console.log('Recover address', recoverAddress);
-    if (address === recoverAddress) return true;
-  } catch (e) {
-    // console.log('Could not recoverAddress:' + e.message);
-  }
+  const recoverAddress = verifyTypedData(domain, types, message, sig);
```
bonustrack, line 17: Wouldn't this function fail in the case where sig is `0x`?

ChaituVR: Opps correct reverted it

bonustrack (approved):
utACK

PR #803, opened 2023-03-31
Added Polygon zkEVM (Chain ID: 1101)

oneski:
Addes Polygon zkEVM to networks.json

- adds polygon zkevm to networks.json (chain id: 1101)

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -1033,6 +1033,23 @@
     "start": 451,
     "logo": "ipfs://QmYeskHqrEvWHqeAuqett64LxfH52HUXZi2T9BAMmgKvBF"
   },
+  "1101": {
+    "key": "1101",
+    "name": "Polygon zkEVM Mainnet",
+    "shortName": "Polygon zkEVM",
+    "chainId": 1101,
+    "network": "mainnet",
+    "multicall": "0x0B89Fc9fC6EAba660Ab984996E133C21E23F535D",
+    "rpc": [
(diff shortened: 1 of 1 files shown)


src/networks.json
```diff
@@ -1033,6 +1033,23 @@
...
+  "1101": {
+    "key": "1101",
+    "name": "Polygon zkEVM Mainnet",
+    "shortName": "Polygon zkEVM",
+    "chainId": 1101,
+    "network": "mainnet",
+    "multicall": "0x0B89Fc9fC6EAba660Ab984996E133C21E23F535D",
+    "rpc": [
+      "https://zkevm-rpc.com",
+      "https://polygon-rpc.com/zkevm"
```
ChaituVR, line 1045: Hi @oneski You have any other nodes than these? these nodes have very less payload limit so they won't work for us,

Everything else looks good 👍

ChaituVR: Oh maybe it is just the network thing, it fails with the error `"failed to execute the unsigned transaction"` if we pass more addresses to get voting power. not sure what to do in this case @bonustrack Any idea? it works only if we pass less than 167 addresses

oneski: these are the public nodes for zkEVM network, let ask the core team to see if there are other options! I know Ankr and Alchemy have node options, but those require API keys.

ChaituVR:
Closing this as stale, feel free to reopen if needed

PR #814, opened 2023-04-19
refactor: reuse ajv instance

Sekhmet:
Creating ajv instance is expensive, around ~11ms of synchronous execution on my machine. This is where most of `validateSchema`'s runtime is spent, which is executed twice per vote, resulting in 22ms of execution time spent creating fresh ajv instances on every vote.

Reusing ajv instance lowers validation time to ~0.02ms, 500x improvement.

It makes sense to precompute `.compile` as well, but those improvements seem negligible in our current usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.4.77",
+  "version": "0.4.78",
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK

ChaituVR (approved):

PR #820, opened 2023-05-08
Update AirDAO Testnet; Add AirDAO Devnet

SvineruS:


src/networks.json
```diff
@@ -1453,7 +1453,7 @@
     "chainId": 22040,
     "network": "testnet",
     "testnet": true,
-    "multicall": "0x18f7BFad831D8ED9E16C4cD50D6041dD453Ca1EF",
+    "multicall": "0xFCE5b5ca100e58E5635B15b79F0163be4CbD258b",
```
ChaituVR, line 1456: Hi, @SvineruS Why do you want to change this?

SvineruS: Our testnet was redeployed on same chain id; The old contact no longer exists

src/networks.json
```diff
@@ -1453,7 +1453,7 @@
...
     "rpc": [
       "https://network-archive.ambrosus-test.io"
```
ChaituVR, line 1458: @SvineruS This node is not an archive node, do you have any alternate node?

src/networks.json
ChaituVR: This multicall contract is not verified and published here https://testnet.airdao.io/explorer/address/0xFCE5b5ca100e58E5635B15b79F0163be4CbD258b/contract/verify/

src/networks.json
```diff
@@ -1463,7 +1463,26 @@
...
+    "testnet": true,
+    "multicall": "0x55Be7dd776fAbe89d93bAC66ed1cf0Ab31bdd6eB",
+    "rpc": [
+      "https://network-archive.ambrosus-dev.io"
+    ],
+    "ws": [
+      "wss://network-archive.ambrosus-dev.io/ws"
+    ],
+    "explorer": {
+      "url": "https://devnet.airdao.io/explorer"
```
ChaituVR, line 1483: this link is not working https://devnet.airdao.io/explorer

ChaituVR (changes requested):

src/networks.json
SvineruS: :thinking: 
It's archive and working in https://snapshot-networks.on.fleek.co/
Also i can fetch balance on first block

src/networks.json
SvineruS: Yea, devnet explorer is down for now. Is it really required for dev net?

src/networks.json
SvineruS: verified now

src/networks.json
ChaituVR: I should work for other blocks too, at least from the `start` block.

src/networks.json
ChaituVR: Yep, need to verify the multicall contract code, start block etc

src/networks.json
SvineruS: it works for me. which block is not work for you?

PR #823, opened 2023-05-10
feat: add new space settings for nft claimer

### LINKED ISSUE #822: Add new space setting for NFT Claimer
New settings for space are needed for the new NFT Claimer project, PR coming

wa0x6e:
Fixes #822

Add new space settings for NFT claimer

New settings are nested in a `nftClaimer` key. Feel free to propose a new key name if it's not suitable


src/schemas/space.json
```diff
@@ -326,6 +326,39 @@
...
+              "maxLength": 42
+            },
+            "network": {
+              "type": "string",
+              "title": "Network",
+              "maxLength": 12
+            },
+            "enabled": {
+              "type": "boolean"
+            }
```
ChaituVR, line 359: I am not following much but do we need `enabled`? if we have `nftClaimer` other than object, can we consider it as enabled? or do we want to keep the settings when it is disabled?

wa0x6e: NFT Claimer can be enabled, then disabled later

ChaituVR: I mean if it is disabled, do we want to keep the previous settings?

wa0x6e: yes, if because he can enable it back

wa0x6e: And there may be already minted nft

ChaituVR (changes requested):

src/schemas/space.json
samuveth: I'd suggest you add Title, Description and Examples to each field, then you will be able to use `TuneForm` to easily generate the form on the Ui. Example here: https://tune-histoire.vercel.app/ see `Source` in the sidebar for the used Schema

wa0x6e: Added

src/schemas/space.json
```diff
@@ -330,20 +330,31 @@
           "type": "object",
           "properties": {
             "maxSupply": {
-              "type": "number",
-              "minimum": 0
+              "type": "integer",
```
samuveth, line 333: Why integer and not number?

wa0x6e: Because it's an integer, and not floating number

samuveth: I see

samuveth (approved):
utACK

wa0x6e:
Obsolete PR
All settings have been migrated to be store on-chain, and retrieve via subgraph

PR #833, opened 2023-05-30
Add minimum to body in proposal schema

samuveth:


src/schemas/proposal.json
```diff
@@ -56,10 +56,7 @@
           "type": "number",
           "title": "end",
           "minimum": 1000000000,
-          "maximum": 2000000000,
-          "exclusiveMinimum": {
-            "$data": "1/start"
-          }
```
ChaituVR, line 62: Oh, this never works 🥲 Is there some other way to limit this?

ChaituVR (approved):
tAck

src/schemas/proposal.json
ChaituVR: will handle on seq

samuveth: Thanks :) Yeah this was causing error with newer versions.

samuveth: And I couldn't find a solution too do it in another way on the schema. Sorry should have mentioned it in the proposal description

PR #837, opened 2023-05-31
feat: Add space delegation portal schema

samuveth:
The `delegation` wasn't used are far as I can tell, so I removed it as well.

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -237,8 +237,8 @@
         "delegationPortal": {
           "delegationType": {
             "type": "string",
-            "title": "Standard",
-            "description": "The standard used by your delegation contract",
+            "title": "Delegation type",
+            "description": "Specify the type of delegation that you are using",
             "anyOf": [
               { "const": "compound-governor", "title": "Compound governor" }
             ]
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -253,8 +253,8 @@
           "delegationApi": {
             "type": "string",
             "format": "uri",
-            "title": "Subgraph URL",
-            "description": "The URL of your delegation subgraph",
+            "title": "Delegation API",
+            "description": "The URL of your delegation API (e.g a subgraph)",
             "examples": [
               "https://api.thegraph.com/subgraphs/name/arr00/uniswap-governance-v2"
             ]
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -235,7 +235,7 @@
           "additionalProperties": false
         },
         "delegationPortal": {
-          "standard": {
+          "delegationType": {
             "type": "string",
             "title": "Standard",
             "description": "The standard used by your delegation contract",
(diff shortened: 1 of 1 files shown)


src/schemas/space.json
```diff
@@ -234,8 +234,31 @@
...
+            ]
+          },
+          "contract": {
+            "type": "string",
+            "format": "address",
+            "title": "Contract address",
+            "description": "The address of your delegation contract",
+            "examples": ["0x3901D0fDe202aF1427216b79f5243f8A022d68cf"]
+          },
+          "subgraphUrl": {
```
bonustrack, line 253: Depending on the standard the API may not be a subgraph, I would call it `delegationApi` instead.

src/schemas/space.json
```diff
@@ -234,8 +234,31 @@
           "required": ["name"],
           "additionalProperties": false
         },
-        "delegation": {
```
bonustrack, line 237: We shouldn't remove this, may break some stuff

src/schemas/space.json
```diff
@@ -234,8 +234,31 @@
...
-          "type": "boolean"
+        "delegationPortal": {
+          "standard": {
```
bonustrack, line 238: I would call this "delegationType" similar than on SX-UI. The type may not be a real "standard".

src/schemas/space.json
samuveth: It's not used anywhere and it's not a required field. How can I break anything?

src/schemas/space.json
```diff
@@ -234,8 +234,31 @@
...
+            "format": "address",
+            "title": "Contract address",
+            "description": "The address of your delegation contract",
+            "examples": ["0x3901D0fDe202aF1427216b79f5243f8A022d68cf"]
+          },
+          "delegationApi": {
+            "type": "string",
+            "format": "uri",
+            "title": "Subgraph URL",
+            "description": "The URL of your delegation subgraph",
```
samuveth, line 257:
```suggestion
            "title": "Delegation API",
            "description": "The URL of your delegation API (e.g a subgraph)",
```

src/schemas/space.json
```diff
@@ -234,8 +234,31 @@
           "required": ["name"],
           "additionalProperties": false
         },
-        "delegation": {
-          "type": "boolean"
+        "delegationPortal": {
+          "delegationType": {
+            "type": "string",
+            "title": "Standard",
+            "description": "The standard used by your delegation contract",
```
samuveth, line 241:
```suggestion
            "title": "Delegation type",
            "description": "Specify the type of delegation that you are using",
```

src/schemas/space.json
ChaituVR: Yea `delegation` was added before when we thought of refactoring delegation but we didn't work on it anymore, I think it should be fine, couldn't find any references with this, not even on hub

bonustrack (approved):
utACK

tamebadger:
Screenshot 2023-06-07 at 22 22 36
seeing this error popping up everywhere, is it perhaps related ?

ChaituVR:
Reverting this for now

PR #854, opened 2023-06-16
Fetch with timeout

ChaituVR:


src/utils.ts
```diff
@@ -135,8 +135,18 @@ export async function multicall(
   }
 }
 
+export function fetchWithTimeout(url, options): Promise<any> {
+  const timeout = 30000 || options.timeout;
```
ChaituVR, line 139:
```suggestion
  const timeout = options.timeout || 30000;
```

wa0x6e:
Superseded by https://github.com/snapshot-labs/snapshot.js/pull/904

PR #867, opened 2023-07-08
chore: fix test config

wa0x6e:

- Update all dependencies related to tests 
- Test coverage should be based on all files, and not only on tested files


ChaituVR (approved):
utAck

PR #880, opened 2023-08-07
Fallback to relayer if signature is 0x

ChaituVR:
- If signature is `0x` , it will use relayer instead of sequencer
- Should not break anything with existing flow

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -47,8 +47,9 @@ export const domain = {
 
 export default class Client {
   readonly address: string;
+  readonly options: any;
 
-  constructor(address: string = constants.livenet.sequencer) {
+  constructor(address: string = constants.livenet.sequencer, options = {}) {
     address = address.replace(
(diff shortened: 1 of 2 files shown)


src/sign/index.ts
```diff
@@ -70,11 +72,14 @@ export default class Client {
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
     const data: any = { domain, types, message };
     const sig = await signer._signTypedData(domain, data.types, message);
-    //console.log('Sign', { address: checksumAddress, sig, data });
     return await this.send({ address: checksumAddress, sig, data });
   }
 
   async send(envelop) {
+    let address = this.address;
+    console.log('sig', envelop.sig);
```
ChaituVR, line 80:
```suggestion
```

jonesk7734 (approved):
Tested working

samuveth (approved):
utACK

PR #882, opened 2023-08-09
API key in headers to score-api

ChaituVR:
How to test?

Try running `ts-node test/vp.ts` and check the headers in requests

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -28,16 +28,16 @@ interface Strategy {
 }
 
 export const SNAPSHOT_SUBGRAPH_URL = delegationSubgraphs;
-const SCORE_API_KEY = process?.env?.KEYCARD_SECRET || '';
-const SCORE_API_HEADERS = {
-  Accept: 'application/json',
-  'Content-Type': 'application/json'
-};
-if (SCORE_API_KEY) SCORE_API_HEADERS['X-API-KEY'] = SCORE_API_KEY;
-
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -227,11 +234,11 @@ export async function getScores(
     };
     const res = await fetch(scoreApiUrl, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
+      headers: SCORE_API_HEADERS,
       body: JSON.stringify({ params })
     });
     const obj = await res.json();
-    return obj.result.scores;
+    return options.returnValue ? obj.result[options.returnValue] : obj.result;
```
ChaituVR, line 241: This is to remove this duplicate code https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/scores.ts#L46-L74 refer to the comment on top of the function

src/utils.ts
```diff
@@ -28,6 +28,12 @@ interface Strategy {
 }
 
 export const SNAPSHOT_SUBGRAPH_URL = delegationSubgraphs;
+const SCORE_API_KEY = process?.env?.KEYCARD_SECRET || '';
+const SCORE_API_HEADERS = {
```
wa0x6e, line 32: Why are you using UPPER_CASE syntax like its a constant ?

wa0x6e (approved):
tAck

PR #886, opened 2023-08-11
feat(config): add ability to define brovider URL

Todmy:
- add extra variables to functions for configuring hardcoded endpoints/data
- those variables are optional, so it should work the same way for our users

### DIFF
--- a/src/sign/eip1271.ts
+++ b/src/sign/eip1271.ts
@@ -48,8 +48,8 @@ export async function verifyOldVersion(
   return returnValue.toLowerCase() === magicValue.toLowerCase();
 }
 
-export async function verify(address, sig, hash, network = '1') {
-  const provider = getProvider(network);
+export async function verify(address, sig, hash, network = '1', opts) {
+  const provider = getProvider(network, opts);
   if (await verifyDefault(address, sig, hash, provider)) return true;
   return await verifyOldVersion(address, sig, hash, provider);
 }
--- a/src/sign/eip1271.ts
+++ b/src/sign/eip1271.ts
@@ -48,8 +48,8 @@ export async function verifyOldVersion(
   return returnValue.toLowerCase() === magicValue.toLowerCase();
 }
 
-export async function verify(address, sig, hash, network = '1', opts) {
-  const provider = getProvider(network, opts);
+export async function verify(address, sig, hash, network = '1', options = {}) {
+  const provider = getProvider(network, options);
   if (await verifyDefault(address, sig, hash, provider)) return true;
   return await verifyOldVersion(address, sig, hash, provider);
 }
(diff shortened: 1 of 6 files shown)


samuveth (approved):
utACK

src/sign/eip1271.ts
```diff
@@ -48,8 +48,8 @@ export async function verifyOldVersion(
   return returnValue.toLowerCase() === magicValue.toLowerCase();
 }
 
-export async function verify(address, sig, hash, network = '1') {
-  const provider = getProvider(network);
+export async function verify(address, sig, hash, network = '1', opts) {
+  const provider = getProvider(network, opts);
```
ChaituVR, line 52: It is not optional? then we should change it in sequencer too? 🤔

src/utils.ts
```diff
@@ -392,15 +398,18 @@ export async function getSpaceController(
...
-  return await getEnsOwner(id, network);
+  return await getEnsOwner(id, network, opts);
 }
 
 export async function getDelegatesBySpace(
   network: string,
   space: string,
-  snapshot = 'latest'
+  snapshot = 'latest',
+  options: any = {}
```
ChaituVR, line 408: `opts` or `options` everywhere 😋

src/sign/eip1271.ts
Todmy: it's optional

src/utils.ts
```diff
@@ -392,15 +398,18 @@ export async function getSpaceController(
...
 ) {
-  if (!delegationSubgraphs[network]) {
+  const subgraphNetwork =
```
ChaituVR, line 410: This should be subgraphUrl

src/utils.ts
```diff
@@ -392,15 +398,18 @@ export async function getSpaceController(
...
+    options.subgraphNetwork || SNAPSHOT_SUBGRAPH_URL[network];
+  if (!subgraphNetwork) {
```
ChaituVR, line 412:
```suggestion
  const subgraphUrl =
    options.subgraphUrl || SNAPSHOT_SUBGRAPH_URL[network];
  if (!subgraphUrl) {
```

src/utils.ts
```diff
@@ -432,10 +441,7 @@ export async function getDelegatesBySpace(
   while (true) {
     params.delegations.__args.skip = page * PAGE_SIZE;
 
-    const pageResult = await subgraphRequest(
-      delegationSubgraphs[network],
-      params
-    );
+    const pageResult = await subgraphRequest(subgraphNetwork, params);
```
ChaituVR, line 444:
```suggestion
    const pageResult = await subgraphRequest(subgraphUrl, params);
```

ChaituVR (approved):
utAck

PR #895, opened 2023-09-08
feat(proposal-update): add ability to update proposal

Todmy:


src/sign/types.ts
```diff
@@ -29,6 +29,19 @@ export interface Proposal {
...
+  proposal: string;
+  from: string;
+  space: string;
+  timestamp: number;
+  type?: ProposalType;
+  title?: string;
+  body?: string;
+  discussion?: string;
+  choices?: string[];
+  plugins?: string;
```
wa0x6e, line 42: Are those type really optional ? Isn't the signing failing if they're missing ?

ChaituVR: this is resolved?

wa0x6e: I believe so

wa0x6e (approved):
tAck

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.5.9",
+  "version": "0.5.10",
```
ChaituVR, line 3: Feel like this is a slightly bigger update :P
```suggestion
  "version": "0.6.0",
```

src/sign/types.ts
```diff
@@ -126,6 +139,21 @@ export const proposalTypes = {
   ]
 };
 
+export const updateProposalTypes = {
+  updateProposal: [
```
ChaituVR, line 143: All other types start with a capital letter, maybe better to make it `UpdateProposal`?

ChaituVR (approved):

PR #904, opened 2023-09-21
Migrate ofetch

wa0x6e:

Current `fetch` request does not support timeout, not keep alive connections.

## 💊 Fixes / SolutionM

Migrate the fetch library from `cross-fetch` to `ofetch`

## 🚧 Changes

- Add support for keep alive connections
- Add timeout to all fetch requests

## ‼️ Breaking changes

- `ofetch` is handling error response differently, and is rejecting a Promise on all non-200 errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
     "@ethersproject/wallet": "^5.6.2",
     "ajv": "^8.11.0",
     "ajv-formats": "^2.1.1",
-    "cross-fetch": "^3.1.6",
     "json-to-graphql-query": "^2.2.4",
-    "lodash.set": "^4.3.2"
+    "lodash.set": "^4.3.2",
+    "ofetch": "^1.3.3"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^18.1.0",
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -87,6 +87,7 @@ export default class Client {
         Accept: 'application/json',
         'Content-Type': 'application/json'
       },
+      timeout: 20e3,
       body: JSON.stringify(envelop)
     };
     return new Promise((resolve, reject) => {
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -1,4 +1,4 @@
-import fetch from 'cross-fetch';
+import { ofetch as fetch } from 'ofetch';
 import { Web3Provider } from '@ethersproject/providers';
 import { Wallet } from '@ethersproject/wallet';
 import { getAddress } from '@ethersproject/address';
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -143,6 +143,7 @@ export async function subgraphRequest(url: string, query, options: any = {}) {
       'Content-Type': 'application/json',
       ...options?.headers
     },
+    timeout: 20e3,
     body: JSON.stringify({ query: jsonToGraphQLQuery({ query }) })
   });
   let responseData: any = await res.text();
@@ -184,7 +185,7 @@ export function getUrl(uri, gateway = gateways[0]) {
 
 export async function getJSON(uri, options: any = {}) {
   const url = getUrl(uri, options.gateways);
-  return fetch(url).then((res) => res.json());
+  return fetch(url, { timeout: 30e3 }).then((res) => res.json());
 }
 
 export async function ipfsGet(
@@ -193,7 +194,7 @@ export async function ipfsGet(
   protocolType = 'ipfs'
 ) {
   const url = `https://${gateway}/${protocolType}/${ipfsHash}`;
-  return fetch(url).then((res) => res.json());
+  return fetch(url, { timeout: 20e3 }).then((res) => res.json());
 }
 
 export async function sendTransaction(
@@ -235,6 +236,7 @@ export async function getScores(
     const res = await fetch(scoreApiUrl, {
       method: 'POST',
       headers: scoreApiHeaders,
+      timeout: 60e3,
       body: JSON.stringify({ params })
     });
     const obj = await res.json();
@@ -268,6 +270,7 @@ export async function getVp(
   const init = {
     method: 'POST',
     headers: scoreApiHeaders,
+    timeout: 60e3,
     body: JSON.stringify({
       jsonrpc: '2.0',
       method: 'get_vp',
@@ -309,6 +312,7 @@ export async function validate(
   const init = {
     method: 'POST',
     headers: scoreApiHeaders,
+    timeout: 30e3,
     body: JSON.stringify({
       jsonrpc: '2.0',
       method: 'validate',
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,4 +1,4 @@
-import fetch from 'cross-fetch';
+import { ofetch as fetch } from 'ofetch';
 import { Interface } from '@ethersproject/abi';
 import { Contract } from '@ethersproject/contracts';
 import { isAddress } from '@ethersproject/address';
--- a/yarn.lock
+++ b/yarn.lock
@@ -1789,13 +1789,6 @@ create-hmac@^1.1.0, create-hmac@^1.1.4, create-hmac@^1.1.7:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-cross-fetch@^3.1.6:
-  version "3.1.6"
-  resolved "https://registry.yarnpkg.com/cross-fetch/-/cross-fetch-3.1.6.tgz#bae05aa31a4da760969756318feeee6e70f15d6c"
-  integrity sha512-riRvo06crlE8HiqOwIpQhxwdOk4fOeR7FVM/wXoxchFEqMNUjvbs3bfo4OTgMEMHzppd4DxFBDbyySj8Cv781g==
-  dependencies:
-    node-fetch "^2.6.11"
-
 cross-spawn@^6.0.5:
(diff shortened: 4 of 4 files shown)


src/utils.ts
```diff
@@ -234,25 +228,25 @@ export async function getScores(
       strategies,
       addresses
     };
+
     const res = await fetch(scoreApiUrl, {
       method: 'POST',
       headers: scoreApiHeaders,
+      timeout: 60e3,
```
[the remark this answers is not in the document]
wa0x6e: - subgraphRequest: 20s (most of errors are timeout errors. Better timing out earlier, and let people retry)
- getScores: 60s (very generous, better than default 5min. Most requests resolves in less than 1s, 30s on heavy traffic. We start getting a lot of 500 errors when response time >1min)
- getVp: 60s (same as above)
- validate: 30s (should resolve in a few seconds usually)
- getJSON: 30s (very generous timeout for a simple JSON fetch, lower than default 5min)
- ipfsGet: 20s (pineapple should respond within 15s, all internal request has a 15s timeout)
- send: 20s (from dashboard, request to sequencer always resolve below 20s)


Those values are just starting values, and may be adjusted later, depending on results.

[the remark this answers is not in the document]
Todmy: thank you for the detailed explanation

Todmy (approved):
utACK

Todmy (approved):
utACK

wa0x6e:
Fixed the `fetch` request from `Client`, please review again, as it's a core function

package.json
Todmy, line 1: you need to update this package version

src/sign/index.ts
```diff
@@ -88,16 +88,17 @@ export default class Client {
...
-      fetch(address, init)
-        .then((res) => {
-          if (res.ok) return resolve(res.json());
-          throw res;
-        })
-        .catch((e) => e.json().then((json) => reject(json)));
-    });
+
+    try {
+      return await fetch(address, init);
```
Todmy, line 95: you don't need `await` here

Todmy (changes requested):

src/sign/index.ts
wa0x6e: I think there's a little difference when used inside a try/catch
https://jakearchibald.com/2017/await-vs-return-vs-return-await/#returning

Todmy: you are right, haven't thought about this

Todmy (approved):
utACK

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.6.2",
+  "version": "1.0.0-beta.0",
```
bonustrack, line 3: Why this change? We shouldn't go from a normal version to a beta. Also not sure we should bump to v1. I dont see any change in function or parameters

bonustrack:
@Todmy Can you test this PR, utACK is not enough

src/utils.ts
```diff
@@ -186,7 +181,15 @@ export function getUrl(uri, gateway = gateways[0]) {
 
 export async function getJSON(uri, options: any = {}) {
   const url = getUrl(uri, options.gateways);
-  return fetch(url).then((res) => res.json());
+  return fetch(url, {
+    timeout: 30e3,
+    parseResponse: JSON.parse,
```
Todmy, line 186: why do we need this prop only for `getJSON` and `ipfsGet`?

test/e2e/utils.spec.ts
Todmy, line 1: have you missed tests for `subgraphRequest`?

Todmy (changes requested):

src/utils.ts
```diff
@@ -234,25 +244,25 @@ export async function getScores(
...
+    const body = await fetch(scoreApiUrl, {
       method: 'POST',
       headers: scoreApiHeaders,
-      body: JSON.stringify({ params })
+      timeout: 60e3,
+      body: { params }
     });
-    const obj = await res.json();
-
-    if (obj.error) {
```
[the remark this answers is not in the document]
wa0x6e: `getScores()` should always return a JSON-RPC error with status code 4xx or 5xx, so this case has been moved inside the catch block

src/utils.ts
```diff
@@ -281,19 +292,20 @@ export async function getVp(
...
       }
-    })
+    }
   };
 
   try {
-    const res = await fetch(options.url, init);
-    const json = await res.json();
-    if (json.error) return Promise.reject(json.error);
-    if (json.result) return json.result;
```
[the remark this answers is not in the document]
wa0x6e: AFAIK, no 200 response should ever return an object with the `error` key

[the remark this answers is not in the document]
wa0x6e: > Also, consider that the getVp function returned the result if body.result otherwise it returned undefined. It can be an error from the past, but better to check

What do you mean by that ?

src/utils.ts
```diff
@@ -234,25 +244,25 @@ export async function getScores(
...
+      : body.result[options.returnValue || 'scores'];
   } catch (e) {
-    if (e.errno) {
-      return Promise.reject({ code: e.errno, message: e.toString(), data: '' });
-    }
-    return Promise.reject(e);
+    return Promise.reject(
+      e.data?.error || {
+        code: e.status || 0,
+        message: e.statusText || e.toString(),
```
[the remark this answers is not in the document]
wa0x6e: `errno` was removed from this PR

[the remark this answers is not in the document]
wa0x6e: Previous `if (e.errno) {` was only triggered on very few network error, such as invalid hostname, timeout, and Abort.

Status code 400 or 500 does not go to the catch block.

AFAIK, none of our services are handling that error, and the `message` field is not used.

src/utils.ts
```diff
@@ -281,19 +292,20 @@ export async function getVp(
...
 
   try {
-    const res = await fetch(options.url, init);
-    const json = await res.json();
-    if (json.error) return Promise.reject(json.error);
-    if (json.result) return json.result;
+    const body = await fetch(options.url, init);
+    return body.result;
   } catch (e) {
-    if (e.errno) {
```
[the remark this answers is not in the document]
wa0x6e: We remove a node specific error, and replace it with a cross env error

src/utils.ts
wa0x6e: Removed it, found another way to handle this

test/e2e/utils.spec.ts
wa0x6e: Added. They were already covered by blockfinder.spec.ts though

src/utils.ts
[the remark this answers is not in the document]
Todmy: I mean [after this line](https://github.com/snapshot-labs/snapshot.js/pull/904/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351L291) it returns nothing(`undefined`). But your implementation always returns data. As I said, it could be an error in the previous version, but better to check in places where we use `getVp`

Todmy:
@wa0x6e could you please test your tests? I run them and have some failed
Screenshot 2023-09-29 at 11 44 32
a lot of them related to the `https://httpstat.us/200?sleep=5000` but there are other errors

Todmy (changes requested):
tests should be fixed

src/utils.ts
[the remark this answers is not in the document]
wa0x6e: I think you're reading the wrong commit, the line you're mentioning have been removed

wa0x6e:
> @wa0x6e could you please test your tests? I run them and have some failed Screenshot 2023-09-29 at 11 44 32 a lot of them related to the `https://httpstat.us/200?sleep=5000` but there are other errors

Tests are passing on my local, and on github actions 🤔

What are the errors you're seeing ?

wa0x6e:
Added retry to flaky tests depending on third party services

Todmy:
issue persists. The reason why tests fail on my local host is that I use node@18. This package is used for projects with node@18 as well.

wa0x6e:
Hmmmm… have to start adding node 18 to ci

wa0x6e:
> Hmmmm… have to start adding node 18 to ci

https://github.com/snapshot-labs/snapshot.js/pull/911

wa0x6e:
> issue persists. The reason why tests fail on my local host is that I use node@18. This package is used for projects with node@18 as well.

Tests fixed for node 18

Todmy (approved):
tACK

test/e2e/utils/blockfinder.spec.js
ChaituVR, line 1: Why `.js` 🤔

wa0x6e: Because all existing previous test were using `.js`

wa0x6e:
I believe this should not be merged on `master`, but on a dedicated v1 branch. We could then npm publish the beta from that branch

ChaituVR:
Maybe we can leave it on this branch, can i publish it to beta tag?

wa0x6e:
> Maybe we can leave it on this branch, can i publish it to beta tag?

I don’t think it’s possible to have 2 versions on the same branch, it’ll create a merge conflict

ChaituVR:
Published this branch to https://www.npmjs.com/package/@snapshot-labs/snapshot.js/v/0.8.0-beta.0

IndigoFlo (approved):

PR #909, opened 2023-09-30
fix: Add timestamp to follow, unfollow, subscribe, unsubscribe, delete-space and alias

ChaituVR:
Problem:
- Without timestamp in types, `id` generated from hash is always same, so it is updating the previous message in the `messages` table
- Example: `follow` a space -> `unfollow` that space -> `follow` that space again -> Notice that it updates an existing record 
- This is blocking https://github.com/snapshot-labs/snapshot-sequencer/pull/172/

How to test:
- Update snapshot.js in node_modules with new changes in snapshot-sequencer
-  `follow` a space -> `unfollow` that space -> `follow` that space again -> Notice that `id` returned should be different


wa0x6e (approved):
utAck

PR #911, opened 2023-10-02
chore: run CI tests on more node versions

wa0x6e:
Run the github action tests on more node versions.

It's currently only ran on node 14, which is dead for many years.

Adding 16 (which is EOL but is still used by many of our services), 18 (current) and 20.

This will avoid scenario like https://github.com/snapshot-labs/snapshot.js/pull/904 which fails on node 18

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['14', 16', '18', '20']
+        node-version: ['14', '16', '18', '20']
 
     steps:
       - uses: actions/checkout@v2.3.4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,9 +11,9 @@ jobs:
         node-version: ['14', '16', '18', '20']
 
     steps:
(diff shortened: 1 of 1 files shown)


ChaituVR:
But this will increase number of actions 
image

we may hit concurrent runs soon, which may cause some delay to merge things https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions

wa0x6e:
We can 2:
- the extreme: so 14 and 20
- the most used: 16 and 18

wdyt?

ChaituVR:
that would be perfect 👌

wa0x6e:
> that would be perfect 👌

Which one?

.github/workflows/test.yml
```diff
@@ -8,12 +8,12 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: ['14', '16', '18', '20']
```
ChaituVR, line 11: I think extremes we want to support 
```suggestion
        node-version: ['14', '20']
```

ChaituVR (approved):
utAck

PR #913, opened 2023-10-05
fix: Add schema for alias

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot-sequencer/pull/189

We need to check if alias address is a valid address


wa0x6e (approved):
tAck

PR #927, opened 2023-10-31
fix: File naming conflict

samuveth:
Was getting these error on FE after updating to Vite 4.5.

Seems there was a conflict with naming with `types.ts` and `types.json` import here https://github.com/snapshot-labs/snapshot.js/blob/e716ec157e436fe1084bfb2f39c5b1a8bf75de51/src/sign/index.ts#L40

I tried changing the import to `./types.ts` but got:
```
An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.ts(5097)
```

Instead of adding that rule `allowImportingTsExtensions` I decided to change the file name to something more fitting `hashedTypes.json`


ChaituVR (approved):
utAck

PR #928, opened 2023-10-31
fix: check for invalid arguments in `getScores`

### LINKED ISSUE #923: fix: add more validation rules to `getScores`

wa0x6e:

From https://github.com/orgs/snapshot-labs/projects/12/views/7?pane=issue&itemId=43073796

The `getScores` passes the arguments as is to score-api without check. There are some basic check possible from the SDK, to filter out invalid requests, and lighten the score-api server load.

This will also avoid log pollution on score-api, making the log more relevant

## 💊 Fixes / Solution

Fix #923  

Add some basic argument validation on the SDK side, and return a meaningful error when argument are not valid.

## 🚧 Changes

- Add check that `addresses` should always be an array, and contains always at least on valid address
- Add check that network is not empty, and valid (referenced in networks.json)
- Check that the given `snapshot` is in the network range (should be greater than the network's `start` block)
- Check that `network` in strategies are valid
- Fully tests `getScores`

## 🛠️ Tests

- Send some request to `getScores` functions, with invalid data
- The function should throw an error with an error message


src/utils.ts
```diff
@@ -222,6 +223,33 @@ export async function getScores(
...
+  }
+  strategies.forEach((strategy) => {
+    if (strategy.network && !networks[strategy.network]) {
+      throw new Error(
+        `Invalid network (${strategy.network}) in strategy ${strategy.name}`
+      );
+    }
+  });
+  if (typeof snapshot === 'number' && snapshot < networks[network].start) {
+    throw new Error(
```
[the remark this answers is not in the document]
wa0x6e: Correct

src/utils.ts
```diff
@@ -222,6 +223,37 @@ export async function getScores(
...
+          `Invalid network (${strategy.network}) in strategy ${strategy.name}`
+        );
+      }
+    });
+  } catch (e) {
+    return Promise.reject(e);
+  }
+  if (!isValidSnapshot(snapshot, network)) {
+    return Promise.reject(
+      `Snapshot (${snapshot}) must be greater than network start block (${networks[network].start})`
```
ChaituVR, line 253:
```suggestion
      `Snapshot (${snapshot}) must be 'latest' or greater than network start block (${networks[network].start})`
```

src/utils.ts
```diff
@@ -497,6 +529,21 @@ export function getNumberWithOrdinal(n) {
...
+}
+
+function isValidAddress(address: string) {
+  return isAddress(address) && address !== EMPTY_ADDRESS;
+}
+
+function isValidSnapshot(snapshot: number | string, network: string) {
+  return (
+    (typeof snapshot === 'number' && snapshot >= networks[network].start) ||
+    snapshot === 'latest'
```
ChaituVR, line 543:
```suggestion
    snapshot === 'latest' ||
    (typeof snapshot === 'number' && snapshot >= networks[network].start)
```

src/utils.spec.js
```diff
@@ -0,0 +1,187 @@
...
+          'addresses is an empty array',
+          { addresses: [] },
+          /addresses can not be empty/i
+        ],
+        ['network is not valid', { network: 'mainnet' }, /invalid network/i],
+        ['network is empty', { network: '' }, /invalid network/i],
+        [
+          'snapshot is smaller than start block',
+          { snapshot: 1234 },
+          /snapshot \([0-9]+\) must be greater than network start block/i
```
ChaituVR, line 82:
```suggestion
          /snapshot \([0-9]+\) must be 'latest' or greater than network start block/i
```

src/utils.ts
```diff
@@ -249,6 +250,37 @@ export async function getScores(
...
+    return Promise.reject('addresses should be an array of addresses');
+  }
+  if (addresses.length === 0) {
+    return Promise.reject('addresses can not be empty');
+  }
+  for (const address of addresses) {
+    if (!isValidAddress(address)) {
+      return Promise.reject(`Invalid address: ${address}`);
+    }
+  }
```
ChaituVR, line 263: I feel like better to use `.some`

ChaituVR:
```suggestion
  if (addresses.some(address => !isValidAddress(address))) {
    return Promise.reject(`Invalid address: ${address}`);
  }
```

src/utils.ts
```diff
@@ -249,6 +250,37 @@ export async function getScores(
...
+    strategies.forEach((strategy) => {
+      if (strategy.network && !isValidNetwork(strategy.network)) {
+        throw new Error(
+          `Invalid network (${strategy.network}) in strategy ${strategy.name}`
+        );
+      }
+    });
+  } catch (e) {
+    return Promise.reject(e);
+  }
```
ChaituVR, line 277: Why different approach here (try,catch), can be the same `Promise.reject` like every where else

wa0x6e: Because it can not reject inside forEach because of the anonymous function. Will use `.some` as suggested above

ChaituVR:
```suggestion
  const invalidStrategy = strategies.find(strategy => strategy.network && !isValidNetwork(strategy.network));
  
  if (invalidStrategy) {
    return Promise.reject(new Error(`Invalid network (${invalidStrategy.network}) in strategy ${invalidStrategy.name}`));
  }
```

ChaituVR (approved):
tACK

PR #929, opened 2023-10-31
fix: check for invalid arguments in `validate`

### LINKED ISSUE #924: fix: add more validations rules to `validate`

wa0x6e:

From https://github.com/orgs/snapshot-labs/projects/12/views/7?pane=issue&itemId=43073796

The `validate` passes the arguments as is to score-api without check. There are some basic check possible from the SDK, to filter out invalid requests, and lighten the score-api server load.

This will also avoid log pollution on score-api, making the log more relevant

## 💊 Fixes / Solution

Fix #924

Add some basic argument validation on the SDK side, and return a meaningful error when argument are not valid.

## 🚧 Changes

- Add check that author address is valid
- Add check that network is not empty, and valid (referenced in networks.json)
- Check that the given `snapshot` is in the network range (should be greater than the network's `start` block)
- Check that `network` in strategies are valid
- Fully tests `validate`

## 🛠️ Tests

- Send some request to `validate` functions, with invalid data
- The function should throw an error with an error message

### DIFF
--- a/src/utils.spec.js
+++ b/src/utils.spec.js
@@ -67,7 +67,7 @@ describe('utils', () => {
         [
           'snapshot is smaller than start block',
           { snapshot: 1234 },
-          /snapshot \([0-9]+\) must be greater than network start block/i
+          /snapshot \([0-9]+\) must be 'latest' or greater than network start block/i
         ]
       ];
 
--- a/src/utils.spec.js
+++ b/src/utils.spec.js
@@ -1,0 +1,164 @@
+import { describe, test, expect, vi, afterEach } from 'vitest';
+import * as crossFetch from 'cross-fetch';
+import { validate } from './utils';
+
+vi.mock('cross-fetch', async () => {
+  const actual = await vi.importActual('cross-fetch');
+
+  return {
+    ...actual,
+    default: vi.fn()
+  };
+});
+const fetch = vi.mocked(crossFetch.default);
+
+describe('utils', () => {
+  afterEach(() => {
+    vi.resetAllMocks();
+  });
+
+  describe('validate', () => {
+    const payload = {
+      validation: 'basic',
+      author: '0xeF8305E140ac520225DAf050e2f71d5fBcC543e7',
+      space: 'fabien.eth',
+      network: '1',
+      snapshot: 7929876,
+      params: {
+        minScore: 0.9,
+        strategies: [
+          {
+            name: 'eth-balance',
+            params: {}
+          }
+        ]
+      }
+    };
+
+    function _validate({
+      validation,
+      author,
+      space,
+      network,
+      snapshot,
+      params,
+      options
+    }) {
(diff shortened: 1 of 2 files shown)


wa0x6e:
@ChaituVR do we allow empty author (0x000…) or should we also add a check for that?

ChaituVR:
> @ChaituVR do we allow empty author (0x000…) or should we also add a check for that?

We have a check on score-api is it important to check everything here also?

wa0x6e:
> > @ChaituVR do we allow empty author (0x000…) or should we also add a check for that?
> 
> We have a check on score-api is it important to check everything here also?

This one is less important, but does not cost to add, and benefit will be less requests to score-api

ChaituVR:
Lets add it then 😅

src/utils.ts
```diff
@@ -333,6 +334,19 @@ export async function validate(
...
+  if (!isValidAddress(author)) {
+    return Promise.reject(`Invalid author: ${author}`);
+  }
+
+  if (!isValidNetwork(network)) {
+    return Promise.reject(`Invalid network: ${network}`);
+  }
+  if (!isValidSnapshot(snapshot, network)) {
+    return Promise.reject(
+      `Snapshot (${snapshot}) must be greater than network start block (${networks[network].start})`
```
ChaituVR, line 346:
```suggestion
      `Snapshot (${snapshot}) must be 'latest' or greater than network start block (${networks[network].start})`
```

src/utils.ts
```diff
@@ -530,6 +544,21 @@ export function getNumberWithOrdinal(n) {
...
+
+function isValidAddress(address: string) {
+  return isAddress(address) && address !== EMPTY_ADDRESS;
+}
+
+function isValidSnapshot(snapshot: number | string, network: string) {
+  return (
+    (typeof snapshot === 'number' && snapshot >= networks[network].start) ||
+    snapshot === 'latest'
+  );
```
ChaituVR, line 559:
```suggestion
  return (
    snapshot === 'latest' ||
    (typeof snapshot === 'number' && snapshot >= networks[network].start)
  );
```

ChaituVR (approved):
tAck

src/utils.spec.js
```diff
@@ -0,0 +1,164 @@
...
+          'author is an invalid address',
+          { author: 'test-address' },
+          /invalid author/i
+        ],
+        ['network is not valid', { network: 'mainnet' }, /invalid network/i],
+        ['network is empty', { network: '' }, /invalid network/i],
+        [
+          'snapshot is smaller than start block',
+          { snapshot: 1234 },
+          /snapshot \([0-9]+\) must be greater than network start block/i
```
ChaituVR, line 70:
```suggestion
          /snapshot \([0-9]+\) must be 'latest' or greater than network start block/i
```

PR #936, opened 2023-11-10
feat: Check if network is mainnet or testnet depending on env

ChaituVR:
- Add a new keyword `snapshotKeyword` to check if the network exists in networks.json or not
- If snapshot's env is `mainnet` it will check only mainnet networks
- if snapshot's env is `testnet` it will check only testnet networks

To Do: 
- [ ] On UI and sequencer, need to update `validateSchema` function to pass the env in the third optional param 
```js
validateSchema(schema, data, { snapshotEnv: 'mainnet' })
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -37,7 +37,12 @@ const scoreApiHeaders = {
   'Content-Type': 'application/json'
 };
 
-const ajv = new Ajv({ allErrors: true, allowUnionTypes: true, $data: true });
+const ajv = new Ajv({
+  allErrors: true,
+  allowUnionTypes: true,
+  $data: true,
+  passContext: true
+});
 // @ts-ignore
 addFormats(ajv);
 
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -68,6 +73,28 @@ ajv.addFormat('ethValue', {
...
+);
+const testnetNetworkIds = Object.keys(networks).filter(
+  (id) => networks[id].testnet
+);
+
+ajv.addKeyword({
+  keyword: 'snapshotNetwork',
+  validate: function (schema, data) {
+    // @ts-ignore
+    const snapshotEnv = this.snapshotEnv || 'default';
```
wa0x6e, line 88: Why don't we default to `mainnet` ?

ChaituVR: We now support all networks 🙈 so i thought it is the default case

we can change it later because,
If we make mainnet default, then we need to change the sequencer code before merging this, else demo.snapshot.org will also accept only mainnet networks
Also, UI still displays all networks.

wa0x6e: So everything works like before, without the 3rd param ?

ChaituVR: Yep should work

samuveth (approved):
utACK

PR #944, opened 2023-11-24
fix: Add start param to network that missing it

ChaituVR:
Some networks are missing `start` param

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -90,6 +90,7 @@
     "explorer": {
       "url": "https://ubiqscan.io"
     },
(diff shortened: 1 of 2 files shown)


src/networks.json
```diff
@@ -90,6 +90,7 @@
     "explorer": {
       "url": "https://ubiqscan.io"
     },
+    "start": 0,
```
ChaituVR, line 93:
```suggestion
    "start": 1,
```

wa0x6e (approved):
utAck

PR #945, opened 2023-11-25
fix: Only check for mainnet on production

samuveth:


src/utils.ts
```diff
@@ -78,21 +78,16 @@ const networksIds = Object.keys(networks);
...
-  (id) => networks[id].testnet
-);
 
 ajv.addKeyword({
   keyword: 'snapshotNetwork',
-  validate: function (schema, data) {
-    // @ts-ignore
-    const snapshotEnv = this.snapshotEnv || 'default';
+  validate: function (schema, data, options) {
+    const snapshotEnv = options?.snapshotEnv || 'default';
```
ChaituVR, line 85: I tried many things previously but there is no option to pass `options` here, this is called internally in agv, so I was using `.call` in line `435`
```suggestion
  validate: function (schema, data) {
    // @ts-ignore
    const snapshotEnv = this.snapshotEnv || 'default';
```

ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #946, opened 2023-12-12
fix: Ranked Choice total voting power

ChaituVR:
Right now, In ranked choice voting

we calculate the total voting power of all votes after all rounds, which is the same as total voting power of all votes anyway

This is causing problems with shutter, If we want to calculate total voting power after all rounds of RCV, we need all choices to be decrypted, so in this change we are just calculating total voting power of all votes like all other voting types


samuveth (approved):
tACK

PR #949, opened 2023-12-22
feat: Add new keyword to check max length based on space type (proposal body)

ChaituVR:
This will add a new keyword 
`maxLengthWithSpaceType` to allow higher body length limit to turbo spaces

```JSON
"maxLengthWithSpaceType": {
  "default": 10000,
  "turbo": 20000
}
```

so on sequencer or UI while calling `validateSchema` we can pass a new param `spaceType` which can be `default` by default or if we pass `turbo` max length will be `20000`

How to test:
- Run `yarn test` and Try passing bigger `body` to `proposal.json`
- and try adding `spaceType` to `test/schema.spec.ts`
```javascript 
const isValid = validateSchema(schema, example, {
  snapshotEnv: 'mainnet',
  spaceType: 'turbo'
});
```


bonustrack:
We will need to allow more voting strategies too for Turbo spaces

ChaituVR:
> We will need to allow more voting strategies too for Turbo spaces

Yes will create a new PR if we agree on the concept of adding new keyword like here 
follow status here https://discord.com/channels/955773041898573854/1187284357799481484/1187287152426504282

test/schema.spec.ts
```diff
@@ -18,7 +20,44 @@ describe.each([
...
+      snapshotEnv: 'mainnet'
+    });
     expect(isValid).toBe(true);
   });
 });
+
+// Tests for turbo spaces
+
+describe.each([
+  { schemaType: 'space', schema: schemas.space, example: space },
```
wa0x6e, line 33: Why do you test the space validation here, as the given `space` is not a Turbo space ?

wa0x6e (changes requested):

test/schema.spec.ts
```diff
@@ -18,7 +20,44 @@ describe.each([
...
+    expect(isValid).toBe(true);
+  });
+});
+
+describe.each([
+  // { schemaType: 'space', schema: schemas.space, example: space },
+  {
+    schemaType: 'proposal',
+    schema: schemas.proposal,
+    example: proposalTurbo,
```
wa0x6e, line 50: This file name is confusing, as it's named turbo, but will not pass any turbo test

wa0x6e (changes requested):

test/schema.spec.ts
ChaituVR: Yes we will have turbo on space too, changed it here https://github.com/snapshot-labs/snapshot.js/pull/951/files#diff-e755a2ecce42f09829d5c7dc1de8853d1d00ef56eaadc2709601c87b9be8ddfb So i thought better to add a placeholder

test/schema.spec.ts
ChaituVR: Sorry I don't get what you mean 🙈 

Here we test if turbo payload returns error.

wa0x6e (approved):
utAck

PR #951, opened 2023-12-23
feat: Add new keyword to check max items based on space type (strategies)

ChaituVR:
This is very similar to #949


wa0x6e (approved):
utAck

PR #956, opened 2024-01-24
feat: fix fetch delegates error when `skip` value is greater than 5000

wa0x6e:

More context here https://discord.com/channels/955773041898573854/1031838169282392144/1199230970361036880

When fetching delegates from subgraph, when there are more than 5000 delegates, the graphql query returns the error 

```
Errors found in subgraphRequest: URL: https://subgrapher.snapshot.org/api.thegraph.com/subgraphs/name/snapshot-labs/snapshot-arbitrum, Status: 500,  Response: [{"message":"The `skip` argument must be between 0 and 5000, but is 6000"}]
```

## 💊 Fixes / Solution

This PR introduces another way to paginate, using the `timestamp_gte` filter, like we already to on the votes query on the hub

## 🚧 Changes

- Edit the subgraph query to use `timestamp_gte` filter to paginate results
- Fix typing of `getDelegatesBySpace` function
- Add more tests

## Notes

This change may make the query less performant than before, since we now have to filter out potential duplicates.

## 🛠️ Tests

- Test fetching delegates for the arbitrum node, for a recent block
- It should be able to return more than 5000 results


src/utils.ts
```diff
@@ -538,34 +538,62 @@ export async function getDelegatesBySpace(
...
+    const pivot = result[result.length - 1];
+    params.delegations.__args.where.timestamp_gte = pivot?.timestamp || 0;
+
+    let pageDelegations: Delegation[] =
+      (await subgraphRequest(subgraphUrl, params)).delegations || [];
+
+    // Removing duplicates results between pages because of the _gte filter
+    if (pivot) {
+      // List of delegation that can potentially be duplicates in the new page results
+      const pivotDelegations = result.filter((d) => isSameDelegate(d, pivot));
```
ChaituVR, line 588: There can be multiple records with the same `timestamp` better to do filter duplicates once pagination is completed, 

I am also worried about what happens if there are 1000 records with the same timestamp. (highly unlikely)

ChaituVR (changes requested):

src/utils.ts
wa0x6e: Yes, this implementation loosely assumes that there will not be more than 1000 duplicates

Records are sorted by timestamp, so checking for duplicate on each page will be more easier than doing it at once at the end, on an array of 5000 items.

src/utils.ts
```diff
@@ -538,34 +538,62 @@ export async function getDelegatesBySpace(
...
+
+      pageDelegations = pageDelegations.filter(
+        (d) => !pivotDelegations.some((p) => isSameDelegate(p, d))
+      );
```
ChaituVR, line 592:
```suggestion
      pageDelegations = pageDelegations.filter(
         (d) => !result.some((p) => isSameDelegate(p, d))
      );
```

wa0x6e: the `pivotDelegations` is the potential duplicates, so if we use timestamp_gte = 1 on page 1, when we got to page 2, `pivotDelegations` will only contain delegate with `timestamp` == 1.

It was added to decrease the number of search loops, so if `pivotDelegations` is 2 elements, we have 2000 loop (2 * 1000).

By removing it, we will end up with 1000 * 1000 on page 2, and 2000 * 1000 on page 3, etc ...

wa0x6e: The best solution will be to implement a search loop which stop searching after the first negative return, since the array is sorted by timestamp, which will drop our loop count to single digit most of the time.

This will make the code less readable, so I go the route of computation is cheaper than dev time for maintenance

ChaituVR: No I think i am not able to explain it correctly 😄 but i mean `pivotDelegations` can just be `1` record, but there can be multiple duplicates in new page because we using `timestamp_gte`

src/utils.ts
```diff
@@ -587,34 +587,62 @@ export async function getDelegatesBySpace(
...
     params.delegations.__args.block = { number: snapshot };
   }
 
+  const isSameDelegate = (a: Delegation, b: Delegation) => {
+    return (
+      a.delegator === b.delegator &&
+      a.delegate === b.delegate &&
+      a.space === b.space
+    );
+  };
```
ChaituVR, line 625: Should be outside of this function

src/utils.ts
ChaituVR: Maybe i am confused but Imagine if the first page returns
```JSON
[{delegate: "A", timestamp: 1},{delegate: "B", timestamp: 2},{delegate: "C", timestamp: 2}]
```
second page returns 
```JSON
[{delegate: "B", timestamp: 2},{delegate: "C", timestamp: 2},{delegate: "D", timestamp: 3}]
```

so now `pivotDelegations` will be `[{delegate: "C", timestamp: 2}]`, but it should be `[{delegate: "B", timestamp: 2},{delegate: "C", timestamp: 2}]`

ChaituVR: Maybe we should filter with just `timestamp `

```javascript
const pivotDelegations = result.filter((d) => isSameTimestamp(d, pivot));

pageDelegations = pageDelegations.filter(
  (d) => !pivotDelegations.some((p) => isSameTimestamp(p, d))
);
```

test/e2e/utils/getDelegatesBySpace.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+
+    expect(results.length).toEqual(fixtures.length);
+    expect(
+      results.every((a: any) => {
+        return fixtures.some((b) => {
+          return (
+            a.delegator === b.delegator &&
+            a.space === b.space &&
+            a.delegate === b.delegate
```
ChaituVR, line 25: better to check timestamp also

src/utils.ts
wa0x6e: Maybe we should just extract this whole logic inside a utils/delegation.ts

src/utils.ts
wa0x6e: You're absolutely correct, the pivotDelegation filter should not use the `isSameDelegate`, but another filter to extract all delegates with same timestamp

wa0x6e: Code refactored, should be more easier to understand

Logic extracted to delegation.ts, and split into multiple functions for easier reading and testing

ChaituVR (approved):
tAck

src/utils/delegation.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (snapshot !== 'latest') {
+    params.delegations.__args.block = { number: snapshot };
+  }
+
+  while (true) {
+    params.delegations.__args.where.timestamp_gte =
+      result[result.length - 1]?.timestamp || 0;
+
+    const newResults: Delegation[] =
+      (await subgraphRequest(subgraphUrl, params)).delegations || [];
```
ChaituVR, line 56: Very highly unlikely but, could we return an error if all delegations of a page have same timestamp 😄

wa0x6e: Implemented, now handling case when there are more than 1000 duplicates, it should be able to handle up to 5000 delegations with same timestamp

src/utils/delegation.ts
```diff
@@ -0,0 +1,120 @@
...
+  const skip = 0;
+  const spaceIn = ['', space];
+  if (space.includes('.eth')) spaceIn.push(space.replace('.eth', ''));
+
+  while (true) {
+    const newResults = await fetchData({
+      url: subgraphUrl,
+      spaces: spaceIn,
+      first: PAGE_SIZE,
+      skip,
```
ChaituVR, line 37: Could we hard code these on `fetchData`?

wa0x6e: Sure, it was added for support of `skip` params in case of > 1000 duplicates, but can remove it

src/utils/delegation.ts
```diff
@@ -0,0 +1,113 @@
...
+      throw new Error('Unable to paginate delegation');
+    }
+
+    if (newResults.length < PAGE_SIZE) break;
+  }
+
+  return result;
+}
+
+export function mergeWithoutDuplicates(a: Delegation[], b: Delegation[]) {
```
ChaituVR, line 54: `a` and `b` are too confusing, maybe we can use something like `existing` and `new` 😄

src/utils/delegation.ts
```diff
@@ -0,0 +1,113 @@
...
+function isSameDelegation(a: Delegation, b: Delegation): boolean {
+  return (
+    a.delegator === b.delegator &&
+    a.delegate === b.delegate &&
+    a.space === b.space &&
+    a.timestamp === b.timestamp
+  );
+}
+
+function getDelegationsFromTimestamp(results: Delegation[], timestamp: number) {
```
ChaituVR, line 74:
```suggestion
function filterDelegationsByTimestamp(results: Delegation[], timestamp: number) {
```

src/utils/delegation.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (newResults.length < PAGE_SIZE) break;
+  }
+
+  return result;
+}
+
+export function mergeWithoutDuplicates(a: Delegation[], b: Delegation[]) {
+  const pivot = a[a.length - 1];
+  const pivotDelegations = pivot
+    ? getDelegationsFromTimestamp(a, pivot.timestamp)
```
ChaituVR, line 57:
```suggestion
    ? filterDelegationsByTimestamp(a, pivot.timestamp)
```

src/utils/delegation.ts
wa0x6e: There's no concept of new or ancient in this case, it's returning same thing no matter the order

ChaituVR, line 54: I just realized that subgraph returns `id` property, i think we can use this property to filter duplicates here than worrying about timestamps?

wa0x6e: the `id` field is somewhat expensive:

```
 {
        "id": "0x000000085d9a759bb5c3d459d638739c0f48deb0-stgdao.eth-0x7265a60acaeaf3a5e18e10bc1128e72f27b2e176",
        "delegator": "0x000000085d9a759bb5c3d459d638739c0f48deb0",
        "space": "stgdao.eth",
        "delegate": "0x7265a60acaeaf3a5e18e10bc1128e72f27b2e176",
        "timestamp": 1681033668
      },
```

We still need the `timestamp` property, and adding the `id` field will increase the amount of data, 35kb vs 43kb for 1000 delegations.

We decrease the amount of computation, at the cost of more data.

For cleaner duplication check, we could also use our own unique keys. So instead of using arrays, we could just use something like:

```
const results = new Map<string, Delegation>();

....

const newResults = await fetchData(....)

newResults.forEach((result: Delegation) => {
  const key = uniqueKey(result);
  if (!results.has(key)) results.set(key, result);
})

function uniqueKey(d: Delegation( {
  return `${d.delegator}-${d.delegate}-${d.space}`
}

**Comment by wa0x6e** [comment]
Another refactoring done, now using `Map` to hold results, with more simpler duplication check

File: src/utils/delegation.ts [hunk 10]
```diff
@@ -0,0 +1,117 @@
...
+function delegationKey(delegation: Delegation) {
+  return `${delegation.delegator}-${delegation.delegate}-${delegation.space}`;
+}
+
+function allDuplicates(delegations: Delegation[]) {
+  return (
+    delegations.length === PAGE_SIZE &&
+    delegations[0].timestamp === delegations[delegations.length - 1].timestamp
+  );
+}
```
ChaituVR, line 63:
```suggestion
function allDuplicates(delegations: Delegation[]) {
  return (
    delegations.length === PAGE_SIZE &&
    delegations[0].timestamp === delegations[delegations.length - 1].timestamp
  );
}

function delegationKey(delegation: Delegation) {
  return `${delegation.delegator}-${delegation.delegate}-${delegation.space}`;
}
```

src/utils/delegation.ts
```diff
@@ -0,0 +1,117 @@
...
+      pivot = delegation.timestamp;
+    });
+
+    if (newResults.length < PAGE_SIZE) break;
+  }
+
+  return [...result.values()];
+}
+
+function allDuplicates(delegations: Delegation[]) {
```
ChaituVR, line 54:
```suggestion
    if (checkAllDuplicates(newResults)) {
      throw new Error('Unable to paginate delegation');
    }

    newResults.forEach((delegation) => {
      concatUniqueDelegation(result, delegation);
      pivot = delegation.timestamp;
    });

    if (newResults.length < PAGE_SIZE) break;
  }

  return [...result.values()];
}

function checkAllDuplicates(delegations: Delegation[]) {
```

ChaituVR (approved):
tAck

PR #965, opened 2024-02-26
Add subscription field

ChaituVR:
To keep track of subscirptions

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -415,7 +415,7 @@
     },
     "start": 332890,
     "logo": "ipfs://bafkreibbrwgwdcfh755dqh4ndta6tja77qbaa36bfv4pstub6prhhore7a",
-    "subscription": "Feb 26, 2024"
+    "subscription": "Feb 26, 2025"
   },
   "204": {
     "key": "204",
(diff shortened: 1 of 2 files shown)


src/networks.json
```diff
@@ -414,7 +414,8 @@
       "url": "https://www.oklink.com/manta"
     },
     "start": 332890,
-    "logo": "ipfs://bafkreibbrwgwdcfh755dqh4ndta6tja77qbaa36bfv4pstub6prhhore7a"
+    "logo": "ipfs://bafkreibbrwgwdcfh755dqh4ndta6tja77qbaa36bfv4pstub6prhhore7a",
+    "subscription": "Feb 26, 2024"
```
ChaituVR, line 418:
```suggestion
    "subscription": "Feb 26, 2025"
```

wa0x6e: Why no timestamp ?

ChaituVR: I thought it should be more human-readable, and if we do `new Date("Feb 26, 2025")` it will return a timestamp

bonustrack:
@ChaituVR We don't need this, we keep track of this in Notion with others infos

PR #974, opened 2024-03-14
feat: Add quorumType to space settings

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/4600

This will add ability to send a new space settings to store optimistic quorum (default to undefined)

## How to test:
- Run `yarn test` 
- Change values in `examples/space.json`
- Try passing different values to `quorumType` 
- If `quorumType` exist, It should accept only `optimistic` and reject all

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -304,6 +304,10 @@
               "type": "number",
               "minimum": 0
             },
+            "quorumType": {
+              "type": "string",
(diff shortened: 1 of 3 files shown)


ChaituVR:
Merging this for now to save time @wa0x6e but let me know if I have to change some names, this should not break anything

PR #981, opened 2024-03-18
fix: UI doesn't show any error if sequencer is down or unavailable

ChaituVR:
- Right now if there is a network issue (internet issue or if sequencer/relayer is down, we couldn't show a error because of this
- This creates lot of confusions with safe voting, especially when it doesn't show any message while voting 
- It return following error 
- image
- UI is stuck like this without any error
image

## How to test
Use the newly added test.html to test signing when sequencer is not available 
- Run `yarn build` (without the fix)
- Run `serve -p 9000` 
- Go to  http://localhost:9000/test/test.html
- Try voting , and should return a Uncaught error
- Now with new fix run `yarn build`
- Run `serve -p 9000` 
- Go to  http://localhost:9000/test/test.html
- Try voting, and you see a caught error of `Network request failed`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.11.12",
+  "version": "0.11.13",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
utAck

PR #983, opened 2024-03-23
Feat: Pass chain ID for trust wallet

### LINKED ISSUE #972: Unable to sign using trust wallet browser
If you try to vote on snapshot.org on trust wallet's browser, you will see it fails silently

image

After some debugging, I got to know the error `provided chainid does not match currently active chain`
and i found that we are not passing `chainId` here https://github.com/snapshot-labs/snapshot.js/blob/master/src/sign/index.ts#L49 

and if we pass it, it works fine ✅ But may cause problems with other wallets or wallets using other chains
```[tasklist]
### Tasks
- [ ] Add chainId and try on metamask
- [ ] Add chanid and try on metamask using different chain
- [ ] Try on trust wallet with different chains
- [ ] Try on wallet connect with different chains
```

ChaituVR:
Fixes #972 

Pass `chainId` into payload for trust wallet

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -83,7 +83,7 @@ export default class Client {
       ...domain
     };
     // @ts-ignore
-    if(window?.ethereum?.isTrust) {
+    if (window?.ethereum?.isTrust) {
       domainData.chainId = (await signer.provider.getNetwork()).chainId;
     }
     const data: any = { domain: domainData, types, message };
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -43,7 +43,11 @@ import constants from '../constants.json';
 const NAME = 'snapshot';
 const VERSION = '0.1.4';
 
-export const domain = {
+export const domain:{
+  name: string,
+  version: string,
+  chainId?: number
+} = {
   name: NAME,
   version: VERSION
   // chainId: 1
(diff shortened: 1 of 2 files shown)


src/sign/index.ts
```diff
@@ -74,8 +78,16 @@ export default class Client {
...
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
-    const data: any = { domain, types, message };
-    const sig = await signer._signTypedData(domain, data.types, message);
+
+    const domainData = {
+      ...domain
+    };
+    // @ts-ignore
+    if(window?.ethereum?.isTrust) {
```
bonustrack, line 86:
```suggestion
    if (window?.ethereum?.isTrust) {
```

src/sign/index.ts
```diff
@@ -74,8 +78,16 @@ export default class Client {
     message.from = message.from ? getAddress(message.from) : checksumAddress;
     if (!message.timestamp)
       message.timestamp = parseInt((Date.now() / 1e3).toFixed());
-    const data: any = { domain, types, message };
-    const sig = await signer._signTypedData(domain, data.types, message);
+
+    const domainData = {
+      ...domain
+    };
```
bonustrack, line 84: Isn't this the same?
```suggestion
    const domainData = domain;
```

ChaituVR: This will create a object reference and update `domain` when you update `domainData` 

Could confuse devs in future, if we add more params

bonustrack (approved):
utACK

PR #984, opened 2024-03-25
fix: Reject with json if request fails with json

ChaituVR:
Right now sequencer sends a response object with an error, we have removed it here #981 as it is avoiding some errors like network errors

### DIFF
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -108,15 +108,10 @@ export default class Client {
         .then((res) => {
           if (res.ok) return resolve(res.json());
           if (res.headers.get('content-type')?.includes('application/json'))
-            return res
-              .json()
-              .then((e) => reject(e))
-              .catch((e) => {
-                reject(e);
-              });
+            return res.json().then(reject).catch(reject);
           throw res;
         })
-        .catch((e) => reject(e));
+        .catch(reject);
     });
   }
 
--- a/src/sign/index.ts
+++ b/src/sign/index.ts
@@ -108,7 +108,12 @@ export default class Client {
         .then((res) => {
           if (res.ok) return resolve(res.json());
           if (res.headers.get('content-type')?.includes('application/json'))
-            return res.json().then((e) => reject(e));
+            return res
+              .json()
+              .then((e) => reject(e))
+              .catch((e) => {
+                reject(e);
+              });
           throw res;
         })
         .catch((e) => reject(e));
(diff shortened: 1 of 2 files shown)


wa0x6e (commented):
What happens when the requested header is json, but server is not returning json ? `res.json()` will fail in this case

src/sign/index.ts
```diff
@@ -107,6 +107,13 @@ export default class Client {
       fetch(address, init)
         .then((res) => {
           if (res.ok) return resolve(res.json());
+          if (res.headers.get('content-type')?.includes('application/json'))
+            return res
+              .json()
+              .then((e) => reject(e))
```
wa0x6e, line 113: SHould not reject here

wa0x6e (changes requested):

src/sign/index.ts
```diff
@@ -107,6 +107,13 @@ export default class Client {
       fetch(address, init)
         .then((res) => {
           if (res.ok) return resolve(res.json());
+          if (res.headers.get('content-type')?.includes('application/json'))
```
wa0x6e, line 110: This is already assumed in line 101 above

ChaituVR: that's for request no? 

here we are checking response headers 
Untitled

wa0x6e: Correct, my bad, I thought it was checking request headers, and not response.

src/sign/index.ts
```diff
@@ -107,9 +107,11 @@ export default class Client {
...
+            return res.json().then(reject).catch(reject);
           throw res;
         })
-        .catch((e) => reject(e));
+        .catch(reject);
```
wa0x6e, line 114: This will reject with an Error object, instead of json, no ? Same with line 111.

What's the goal of this PR ? To always return a JSON object when rejecting ?

ChaituVR: > This will reject with an Error object, instead of json, no ? Same with line 111.

Correct, this is for unknown errors like network error
for example `TypeError: Network request failed` 

we can still return object in this case also, but this will help us debug much better

src/sign/index.ts
ChaituVR: If we don't reject, UI can't catch it https://github.com/snapshot-labs/snapshot/blob/master/src/composables/useClient.ts#L28-L36

wa0x6e (approved):
utAck

Issue #992, opened 2024-05-01
`ReferenceError: window is not defined` while trying to vote with snapshot.js


ChaituVR:
### Expected behavior
Should work without error

### Steps to reproduce the behavior
- `npm install @snapshot-labs/snapshot.js` in any nodejs project
- Try to vote with the code here https://gist.github.com/ChaituVR/db9f78a2f2a7e8347843a6a83393b032
- See the error

franzns:
Thanks for opening the issue. Let me know if I can help

fffbbbccc:
Hey all, i am experiencing the same issue while i was trying to quickly prototype something in node. 
First of all, i had some issues because i didnt check the ethers.js dependencies of snapshot.js and since its using 5.x, the signTypedData function is different (6.x uses signTypedData and 5.x uses _signTypeData). 

Anyhow, after getting this out of the way, i realized the this error is coming from this line [here](https://github.com/snapshot-labs/snapshot.js/blob/2b4650af3123d968fe555246184dca3a35ff072f/src/sign/index.ts#L86C1-L88C6) 
```
    if (window?.ethereum?.isTrust) {
      domainData.chainId = (await signer.provider.getNetwork()).chainId;
    }
```
The generated node code looks like this :
```
if (!((_a = window === null || window === void 0 ? void 0 : window.ethereum) === null || _a === void 0 ? void 0 : _a.isTrust)) return [3 /*break*/, 2];
```
and i think it throw the error because its trying to access window object without first checking if the window object exists, which ofc doesnt in a node env. 
As a horrendous hack, i modified the generated code and replaced it with 
```
if (typeof window === 'undefined' || !window.ethereum || !window.ethereum.isTrust) return [3 /*break*/, 2];
```
which **worked**.  So i figured i am on the right track. 

I opened a pr which will hopefully fix this here : https://github.com/snapshot-labs/snapshot.js/pull/998

ChaituVR:
Thank you @fffbbbccc it is fixed now

PR #998, opened 2024-05-11
added a check for window to avoid breaking node envs

fffbbbccc:
Fixes
addresses https://github.com/snapshot-labs/snapshot.js/issues/992
-  added a check for window to avoid breaking node envs


ChaituVR (approved):
utAck

PR #999, opened 2024-05-17
Fix: Pass api key in headers and detect html error

ChaituVR:
### Summary:
- If the API key is passed in the URL, send it in headers
- Detech if the response is HTML and return an error
- Fix test cases

### How to test: 
- Start your own score API server
- Change score API to return HTML
- Change test files in `test` folder, for example `vp.ts` with the local score API server


src/utils.ts
```diff
@@ -38,6 +39,37 @@ const scoreApiHeaders = {
   'Content-Type': 'application/json'
 };
 
+const DEFAULT_SCORE_API_URL = 'https://score.snapshot.org';
+
+const formatScoreAPIUrl = (url = DEFAULT_SCORE_API_URL) => {
```
wa0x6e, line 44: Maybe use a simple `function` for these new functions, instead of anonymous function, for better stacktrace.

src/utils.ts
```diff
@@ -317,9 +348,9 @@ export async function getScores(
     );
   }
 
-  const url = new URL(scoreApiUrl);
-  url.pathname = '/api/scores';
-  scoreApiUrl = url.toString();
+  const urlObject = new URL(scoreApiUrl);
+  urlObject.pathname = '/api/scores';
+  const { url, headers } = formatScoreAPIUrl(urlObject.toString());
```
wa0x6e, line 353: We could move this pathname assignation, and delegate it to `formatScoreAPIUrl` as an optional second args

```suggestion
  const { url, headers } = formatScoreAPIUrl(scoreApiUrl, '/api/scores');
```

And add

```
if (pathname) urlObject.pathname = pathname;
```

inside `formatScoreAPIUrl`

src/utils.ts
```diff
@@ -38,6 +39,37 @@ const scoreApiHeaders = {
...
+  };
+};
+
+const parseScoreAPIResponse = async (res: any) => {
+  let json: any = await res.text();
+  try {
+    json = JSON.parse(json);
+  } catch (e) {
+    return Promise.reject({
+      code: 500,
```
wa0x6e, line 64: We could get the true status code from `res`, and only fallback to 500 if missing

src/utils.ts
```diff
@@ -38,6 +39,37 @@ const scoreApiHeaders = {
...
+      message: 'Failed to parse response from score API',
```
wa0x6e, line 65: Same for `message`, I think it's available from `res`, so we could get something more descriptive like `Network timeout`

src/utils.ts
```diff
@@ -329,16 +360,12 @@ export async function getScores(
...
-      headers: scoreApiHeaders,
+      headers,
       body: JSON.stringify({ params })
     });
-    const obj = await res.json();
-
-    if (obj.error) {
-      return Promise.reject(obj.error);
-    }
+    const obj = await parseScoreAPIResponse(res);
```
wa0x6e, line 368: Let's call this variable `response`, like on line 431, for consistency

src/utils.ts
```diff
@@ -455,9 +481,8 @@ export async function validate(
   };
 
   try {
-    const res = await fetch(options.url, init);
-    const json = await res.json();
-    if (json.error) return Promise.reject(json.error);
+    const res = await fetch(url, init);
+    const json = await parseScoreAPIResponse(res);
```
wa0x6e, line 485: Same here, better to use `response` for consistency

test/e2e/utils/getScores.spec.ts
```diff
@@ -29,4 +29,22 @@ describe('test getScores', () => {
       data: ''
     });
   });
+  test('getScores should not pass API Key if it is passed in URL', async () => {
```
wa0x6e, line 32:
```suggestion
  
  test('getScores should not pass API Key if it is passed in URL', async () => {
```

wa0x6e: Missing new line

wa0x6e (changes requested):

src/utils.ts
ChaituVR: Passing it in data as in this case error will be in HTML and can be unpredictable.if it is not HTML error we handle it seperately

wa0x6e:
Just noticed that this issue may be fixed with https://github.com/snapshot-labs/snapshot.js/pull/957

PR #1007, opened 2024-06-03
Add Split-Delegation to delegate portal options

samepant:
This allows for space settings to use `split-delegation` in their settings, which is the first step to integrating it into the Snapshot settings UI.


ChaituVR (approved):
utAck

PR #1009, opened 2024-06-05
feat: add support for starknet address

wa0x6e:
Add a new validatin rule for validating evm/starknet address

Inspired by how Argent X is validating starknet address https://github.com/argentlabs/argent-starknet-recover/blob/main/schema.ts#L43

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -89,7 +89,7 @@ addErrors(ajv);
 ajv.addFormat('address', {
   validate: (value: string) => {
     try {
-      return isAddress(value);
+      return /^0x[0-9a-fA-F]{62,64}$/.test(value) || isAddress(value);
     } catch (err) {
       return false;
     }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -99,7 +99,7 @@ ajv.addFormat('address', {
 ajv.addFormat('evmOrStarknetAddress', {
   validate: (value: string) => {
     try {
-      return /^0x[0-9a-fA-F]{62,64}$/.test(value);
+      return isAddress(value) || /^0x[0-9a-fA-F]{62,64}$/.test(value);
     } catch (err) {
       return false;
     }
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -89,7 +89,7 @@ addErrors(ajv);
 ajv.addFormat('address', {
   validate: (value: string) => {
     try {
-      return isAddress(value);
+      return /^0x[0-9a-fA-F]{62,64}$/.test(value) || isAddress(value);
```
ChaituVR, line 92: i think it is better to do `isAddress` first, as it is our primary check

```suggestion
      return isAddress(value) || /^0x[0-9a-fA-F]{62,64}$/.test(value);
```

ChaituVR (changes requested):

src/utils.ts
ChaituVR: This may allow stark addresses here https://github.com/snapshot-labs/snapshot.js/blob/master/src/schemas/alias.json#L11 should be fine?

wa0x6e: Should be fine I think.

wa0x6e: > i think it is better to do `isAddress` first, as it is our primary check

Will do more tests to make sure that it does not throw an error, and move on to the `catch` block, before checking for starknet address

src/utils.ts
```diff
@@ -96,6 +96,16 @@ ajv.addFormat('address', {
   }
 });
 
+ajv.addFormat('evmOrStarknetAddress', {
+  validate: (value: string) => {
+    try {
+      return /^0x[0-9a-fA-F]{62,64}$/.test(value);
```
ChaituVR, line 102: This won't check evm adress right?

ChaituVR (changes requested):

ChaituVR (approved):
utAck

PR #1011, opened 2024-06-07
feat: add more fields to user profile

wa0x6e:
Used in https://github.com/snapshot-labs/sx-monorepo/pull/394

Add `cover`, `twitter` and `github` to user profile


ChaituVR (approved):
utAck

PR #1023, opened 2024-06-18
feat: support additional statement properties

wa0x6e:
Related to https://github.com/snapshot-labs/sx-monorepo/pull/407

This PR adds new properties to `Statement`, which will be need in the new SX space user profile page.

New fields:

- discourse (for user discourse handle)
- status (active/inactive)
- network (same as follow/unfollow)


ChaituVR (approved):
utAck

PR #1025, opened 2024-06-18
Fix make about field optional

wa0x6e:
This PR removes `minLength` on `about` and `discourse` properties, thus allowing empty string as valid value.


ChaituVR (approved):
utAck

PR #1033, opened 2024-07-10
fix: Use ENS name's resolver to get text record

ChaituVR:
### Summary

Right now we call both the old resolver and the new resolver to get the text record,
which is causing this issue: https://discord.com/channels/955773041898573854/1030772407226601522/1260542812106002526

This change will get the resolver of the ENS name and use that resolver's text record

### How to test:
- call getSpaceController before and after the fix
```ts
  snapshot.utils.getSpaceController('buzzedbears.eth', '1').then(console.log).catch(console.error);
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -537,7 +537,8 @@ export async function getEnsTextRecord(
   options: any = {}
 ) {
   const {
-    ensResolvers: ensResolversOpt = networks[network]?.ensResolvers || networks['1'].ensResolvers,
+    ensResolvers: ensResolversOpt = networks[network]?.ensResolvers ||
+      networks['1'].ensResolvers,
     broviderUrl,
     ...multicallOptions
   } = options;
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -535,25 +537,34 @@ export async function getEnsTextRecord(
   options: any = {}
 ) {
   const {
-    ensResolvers: ensResolversOpt,
+    ensResolvers: ensResolversOpt = networks[network]?.ensResolvers ||
```
wa0x6e, line 540: Where do we use this `ensResolverOpts` ? Seems not to be used anywhere on snapshot repo

wa0x6e:
Running the test script still give me `0xDe30040413b26d7Aa2B6Fc4761D80eb35Dcf97aD`

ChaituVR:
> Running the test script still give me `0xDe30040413b26d7Aa2B6Fc4761D80eb35Dcf97aD`

Weird, i tried again and works fine for me , it return `0xCdd598D1588503e1609Ae1e50cdB74473fFB0090`
Maybe you are using the previously generated `dist` folder?

src/utils.ts
ChaituVR: Removed it

wa0x6e:
> > Running the test script still give me `0xDe30040413b26d7Aa2B6Fc4761D80eb35Dcf97aD`
> 
> Weird, i tried again and works fine for me , it return `0xCdd598D1588503e1609Ae1e50cdB74473fFB0090` Maybe you are using the previously generated `dist` folder?

Correct

wa0x6e (approved):
tAck

PR #1039, opened 2024-07-19
chore: fix tests

wa0x6e:
Update tests to reflect that starknet verification can verify on different chain


ChaituVR:
what to do with this one? it work with both? 😄

wa0x6e:
It should not work with both. I don’t know why test tests passed on the previous code, but rerun them now fail as expected

ChaituVR (approved):
tAck

PR #1040, opened 2024-07-24
feat: add support for more starknet signature format

wa0x6e:
Fixing https://github.com/snapshot-labs/sx-monorepo/pull/383#issuecomment-2248228622

SDK currently only support Argent x signature format: `[r, s]` (2 items array)

This PR:

- add support for signature returned by Braavos wallet: `[r,s,v]` (3 items array)
- throw an error when the contract is not deployed
- validate the results return by `is_valid_signature` before returning


src/verify/starknet.ts
```diff
@@ -67,16 +67,24 @@ export default async function verify(
...
+    await contractAccount.is_valid_signature(
+      getHash(data, address),
+      sig.slice(-2)
+    );
 
-  await contractAccount.isValidSignature(getHash(data, address), [
-    sig[0],
-    sig[1]
-  ]);
+    return true;
```
ChaituVR, line 82: Looks like we are checking `is_valid_signature` and irrespective of the response, we return true, maybe that's the reason it accepts testnet signatures also?

wa0x6e: The function is always throwing and error when not `true`

wa0x6e: Added a check to compare returned result before returning `true`

src/verify/starknet.ts
```diff
@@ -67,16 +68,28 @@ export default async function verify(
...
+    );
 
-  return true;
+    return BigNumber.from(result).eq(BigNumber.from('370462705988'));
+  } catch (e: any) {
+    if (e.message.includes('Contract not found')) {
+      throw new Error('Contract not deployed');
+    }
+
+    throw e;
```
ChaituVR, line 93: Can we return `false` if error contains `INVALID_SIG`? 

this will help sequencer to return `invalid signature` if someone pass a invalid sig, right now we return `signature validation failed`

ChaituVR (changes requested):

src/verify/starknet.ts
ChaituVR, line 93:
```suggestion
    if (e.message.includes('INVALID_SIG')) return false;
    throw e;
```

src/verify/starknet.ts
```diff
@@ -67,16 +68,28 @@ export default async function verify(
...
+  try {
+    const contractAccount = new Contract(
+      ABI,
+      address,
+      getProvider(network, options)
+    );
+
+    if (sig.length < 2) {
+      throw new Error('Invalid signature');
+    }
```
ChaituVR, line 80: maybe here also
```suggestion
    if (sig.length < 2) {
      return false
    }
```

ChaituVR (approved):
tAck

PR #1041, opened 2024-07-28
fix: handle error when the ens is not valid

wa0x6e:
Fix https://snapshot-labs.sentry.io/issues/5639825201/?query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=15

This PR will handle error, and return `null` when the given ENS name is not valid


ChaituVR (approved):
utAck

PR #1047, opened 2024-08-20
feat: add support for variables in subgraphRequest function

wa0x6e:
This PR adds support for variable when sending graphql request via the `subgraphRequest` utils function.

Injecting `variables` via the `options` argument is very ugly, but in order to be backward compatible, this seems to be the only way.

### How to test

```ts
import { VariableType } from 'json-to-graphql-query';

const result = await subgraphRequest(
    'https://hub.snapshot.org/graphql',
    {
      __variables: {
        id: 'String!'
      },
      space: {
        __args: {
          id: new VariableType('id')
        },
        id: true,
        name: true
      }
    },
    {
      variables: {
        id: 'aave.eth'
      }
    }
  );
  console.log(result);
```

It should print 

```ts
{ space: { id: 'aave.eth', name: 'Aave' } }
```


ChaituVR:
Where do we want to use this?

wa0x6e:
In sequencer, in the delegate statement pr. Some graphql api does not work with the hardcoded variables inside the query, since they don't use string type, but enum

src/utils.ts
```diff
@@ -254,14 +254,16 @@ export async function multicall(
 }
 
 export async function subgraphRequest(url: string, query, options: any = {}) {
+  const body: Record<string, any> = { query: jsonToGraphQLQuery({ query }) };
+  if (body) body.variables = options.variables;
```
ChaituVR, line 258: I think body will always exist
```suggestion
  if (options.variables) body.variables = options.variables;
```

ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #1050, opened 2024-08-23
feat: allow getDelegatesBySpace to fetch delegates for all spaces

Sekhmet:
With this change space can be passed as null and this will result in all delegates to be returned instead of per-space results only.

This will be used on new delegate-registry-api.

### DIFF
--- a/src/utils/delegation.ts
+++ b/src/utils/delegation.ts
@@ -13,7 +13,7 @@ type Delegation = {
 
 export default async function getDelegatesBySpace(
   network: string,
-  space: string,
+  space: string | null,
   snapshot: string | number = 'latest',
   options: any = {}
 ) {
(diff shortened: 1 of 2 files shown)


src/utils/delegation.ts
```diff
@@ -26,7 +26,7 @@ export default async function getDelegatesBySpace(
 
   let pivot = 0;
   const result = new Map<string, Delegation>();
-  const spaceIn = buildSpaceIn(space);
+  const spaceIn = space ? buildSpaceIn(space) : null;
```
ChaituVR, line 29: It takes too much time to get all delegations 😅 (100.418 seconds) but yea works

ChaituVR (approved):
tAck

src/utils/delegation.ts
Sekhmet: Yeah, this will be reused on the server to only do it once every 24 hours for all spaces on given network so it shouldn't be terrible.

PR #1051, opened 2024-08-28
feat: add network property to delegation portal

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/177

Will add a new `delegationNetwork` property to delegation portal, to save the delegation contract network (chain ID)

Property will be required, and on the UI side, will always set to something on existing spaces, thanks to https://github.com/snapshot-labs/snapshot-hub/pull/916

Updating UI with this update will be backward compatible, as long as the hub PR is merged first


bonustrack (approved):
utACK

PR #1060, opened 2024-09-24
fix: enforce lowecase valid hostname for space domain

wa0x6e:
From https://github.com/snapshot-labs/snapshot-sequencer/pull/439#discussion_r1773253485

This PR enforce that the `domain` property in space settings is a valid hostname, and always lowercase


ChaituVR (approved):
utAck

PR #1067, opened 2024-10-07
fix: loosen some address validation to accept lowercase

wa0x6e:
This PR:

- add a new `evmAddress` validation, which accept checksummed and lowercase EVM address (existing `address` validation is only accepting checksummed address)
- Migrate existing regex address validation to use `evmAddress` 
- Migrate `delegationPortalNetwork` from `address` to `evmAddress` (this is used only on v2, which does not enforce checksummed address)
- Migrate recently treasuries address to also accept lowercase address

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -147,9 +147,7 @@
           "maxItems": 100,
           "items": {
             "type": "string",
-            "pattern": "^0x[a-fA-F0-9]{40}$",
-            "minLength": 42,
-            "maxLength": 42
+            "format": "evmAddress"
           },
           "title": "members",
           "uniqueItems": true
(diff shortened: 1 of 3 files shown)


ChaituVR:
We sure that UI settings send checksum addresses instead of lowercase? else it may break

wa0x6e:
> We sure that UI settings send checksum addresses instead of lowercase? else it may break

The changes are moving strict existing validation (only checksum) to checksum+lowercase

ChaituVR:
> The changes are moving strict existing validation (only checksum) to checksum+lowercase

Ah I see its a new validation 😄 Is there any reason to allow lowercase addresses here? because it was allowed before?

ChaituVR (approved):
utAck

wa0x6e:
> > The changes are moving strict existing validation (only checksum) to checksum+lowercase
> 
> Ah I see its a new validation 😄 Is there any reason to allow lowercase addresses here? because it was allowed before?

Yes, it was allowed before, and the strict validation was triggering errors when saving spaces with lowercase treasuries address

PR #1070, opened 2024-10-09
fix: add meaninful error message for anyof validation

wa0x6e:
This PR fixes the error message for validation with `anyOf`, which were showing `Must match a schema in anyOf.`

This PR adds a more meaningful error message for validation using `anyOf`


ChaituVR (approved):
tAck

PR #1069, opened 2024-10-08
feat: add support for starknet treasuries and delegations

wa0x6e:
Toward https://github.com/snapshot-labs/sx-monorepo/pull/851
Continuing https://github.com/snapshot-labs/snapshot.js/pull/1066#issuecomment-2399931257

This PR adds support for starknet addresses and network for delegations and treasuries

PR #1071, opened 2024-10-09
fix: filter out starknet testnet on mainnet

wa0x6e:
Following on https://github.com/snapshot-labs/snapshot.js/pull/1069

This PR do not validate starknet testnet on mainnet, like for the regular network selector


ChaituVR (approved):
tAck

PR #1073, opened 2024-10-14
fix: make delegationNetwork optional

wa0x6e:
This PR change the delegation network to be optional (will set a default on sequencer before saving)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.12.24",
+  "version": "0.12.25",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
utAck

wa0x6e:
This PR needs something else on sequencer to be complete.

We need to populate the missing delegationNetwork field by the default value if missing. (better to have valid data structure, that have some fallback default on fetch)

ChaituVR:
> This PR needs something else on sequencer to be complete.
> 
> We need to populate the missing delegationNetwork field by the default value if missing. (better to have valid data structure, that have some fallback default on fetch)

Handled in https://github.com/snapshot-labs/snapshot-sequencer/pull/459

PR #1077, opened 2024-10-17
feat: increase app max length

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/235
Discussion https://discord.com/channels/955773041898573854/1296435871431786587

Increase the max length of the proposal and vote property to 128


bonustrack (approved):
utACK

PR #1081, opened 2024-10-27
feat: add-email-subscription-to-profile

wa0x6e:
Fixes # .

- 
- 
-

### DIFF
--- a/src/schemas/profile.json
+++ b/src/schemas/profile.json
@@ -51,6 +51,24 @@
           "title": "farcaster",
           "pattern": "^[a-z0-9-]*$",
           "maxLength": 17
+        },
+        "emailSubscription": {
+          "type": "object",
+          "properties": {
+            "email": {
+              "type": "string",
+              "title": "email",
+              "maxLength": 256
+            },
+            "subscriptions": {
+              "type": "array",
+              "title": "subscriptions",
+              "uniqueItems": true,
+              "items": { "type": "string" }
+            }
+          },
+          "required": [],
+          "additionalProperties": false
         }
       },
       "required": [],
--- a/test/examples/profile-addEmailSubscription.json
+++ b/test/examples/profile-addEmailSubscription.json
@@ -1,0 +1,8 @@
+{
+  "name": "Vitalik",
+  "avatar": "ipfs://NnVTdgcFciJUoxqgsNe9Hy2R7rgvEb5jDu5Pa5mmEG3UghbubTf2M78jUXbKLqKM",
+  "about": "This is my about info",
+  "emailSubscription": {
+    "email": "test@email.com"
+  }
+}
--- a/test/schema.spec.ts
+++ b/test/schema.spec.ts
@@ -7,6 +7,8 @@ import spaceStarknetDelegation from './examples/space-starknet-delegation.json';
 import proposalTurbo from './examples/proposal-turbo.json';
 import vote from './examples/vote.json';
 import profile from './examples/profile.json';
+import profileAddEmailSubscription from './examples/profile-addEmailSubscription.json';
+import profileUpdateEmailSubscription from './examples/profile-updateEmailSubscription.json';
 import statement from './examples/statement.json';
 import alias from './examples/alias.json';
 import schemas from '../src/schemas';
@@ -24,6 +26,16 @@ describe.each([
   { schemaType: 'proposal', schema: schemas.proposal, example: proposal },
   { schemaType: 'vote', schema: schemas.vote, example: vote },
   { schemaType: 'profile', schema: schemas.profile, example: profile },
+  {
+    schemaType: 'profile',
+    schema: schemas.profile,
+    example: profileAddEmailSubscription
+  },
+  {
+    schemaType: 'profile',
+    schema: schemas.profile,
+    example: profileUpdateEmailSubscription
+  },
   { schemaType: 'statement', schema: schemas.statement, example: statement },
   { schemaType: 'zodiac', schema: schemas.zodiac, example: space },
   { schemaType: 'alias', schema: schemas.alias, example: alias }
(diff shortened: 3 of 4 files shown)


src/schemas/profile.json
```diff
@@ -51,6 +51,24 @@
           "title": "farcaster",
           "pattern": "^[a-z0-9-]*$",
           "maxLength": 17
+        },
+        "emailSubscription": {
+          "type": "object",
+          "properties": {
+            "email": {
```
ChaituVR, line 58: Maybe emailHash?

src/schemas/profile.json
```diff
@@ -51,6 +51,24 @@
...
+              "maxLength": 256
+            },
+            "subscriptions": {
+              "type": "array",
+              "title": "subscriptions",
+              "uniqueItems": true,
+              "items": { "type": "string" }
+            }
+          },
+          "required": [],
```
ChaituVR, line 70: Users can pass subscriptions without email? maybe we should make it required

test/examples/profile-addEmailSubscription.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "Vitalik",
+  "avatar": "ipfs://NnVTdgcFciJUoxqgsNe9Hy2R7rgvEb5jDu5Pa5mmEG3UghbubTf2M78jUXbKLqKM",
+  "about": "This is my about info",
+  "emailSubscription": {
+    "email": "test@email.com"
```
ChaituVR, line 6: We will hash this email while saving to IPFS? i think users may not be able to verify their signature using tools like https://signator.io/

Not sure what's the alternative here

ChaituVR (changes requested):

src/schemas/profile.json
wa0x6e: We need to pass whole readable email, to save in the database.

After, it will not be needed anymore, and `email` will be omitted. To unsubscribe, we pass `email: null`

src/schemas/profile.json
wa0x6e: Email will not be available, since not returned by the hub.

test/examples/profile-addEmailSubscription.json
wa0x6e: In that case, we could just remove email from the object before saving to ipfs

ChaituVR: but that will still show it as wrong signature, as we remove one field from signed message

wa0x6e: Will move out the email subscription creation from profile

src/schemas/profile.json
wa0x6e: Added a new `Subscription` schema. Sequencer will not create IPFS for this schema

src/schemas/profile.json
```diff
@@ -51,6 +51,12 @@
           "title": "farcaster",
           "pattern": "^[a-z0-9-]*$",
           "maxLength": 17
+        },
+        "emailSubscriptions": {
+          "type": "array",
+          "title": "email subscriptions",
+          "uniqueItems": true,
+          "items": { "type": "string" }
```
ChaituVR, line 59: This can be sent with `subscribtion` no? do we need it on profile?

wa0x6e: This is for the edition, so that any changes are saved on IPFS

Anything send with `Subscription` will not be saved on ipfs, since containing the email address

ChaituVR: Users won't be able to update their email? we can add these to `Subscription` and make email optional right?

wa0x6e: Updating email => send email with Subscription again

If we add `emailSubscriptions` array to `Subscription`, (which make more sense), we would also skip saving those update on ipfs (not a big deal I guess, means that anything related to subscription will not be saved and tracked on ipfs).

wa0x6e: Moved to `Subscription`, into the `metadata` field

scripts/generateHashWithTypes.ts
```diff
@@ -1,12 +1,11 @@
 const { createHash } = require('crypto');
 
 const types = {
-  Statement: [
+  DeleteSubscription: [
     { name: 'from', type: 'address' },
-    { name: 'timestamp', type: 'uint64' },
-    { name: 'space', type: 'string' },
-    { name: 'about', type: 'string' },
-    { name: 'statement', type: 'string' }
```
ChaituVR, line 9: why we remove statement here? 🤔

wa0x6e:
I am thinking that maybe, we should not use a generic `subscription` name, and directly use `email-subscription`, to avoid confusion with the existing `subscription` type returned by the hub.

ChaituVR:
> I am thinking that maybe, we should not use a generic `subscription` name, and directly use `email-subscription`, to avoid confusion with the existing `subscription` type returned by the hub.

agree

test/schema.spec.ts
```diff
@@ -7,6 +7,7 @@ import spaceStarknetDelegation from './examples/space-starknet-delegation.json';
 import proposalTurbo from './examples/proposal-turbo.json';
 import vote from './examples/vote.json';
 import profile from './examples/profile.json';
+import subscription from './examples/subscription.json';
```
ChaituVR, line 10: i think we should rename this file also to avoid confusions

test/examples/subscription.json
```diff
@@ -0,0 +1,5 @@
+{
+  "type": "email",
+  "value": "my-email@snapshot.org",
+  "metadata": "['summary', 'newProposal']"
```
ChaituVR, line 4: not sure what this file is for 🤔

ChaituVR (changes requested):

PR #1089, opened 2024-11-29
feat: Add env BROVIDER_URL

EthanOK:
Fixes # .

- Add env BROVIDER_URL

### DIFF
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -1,3 +1,4 @@
+import 'dotenv/config';
 import {
   JsonRpcBatchProvider,
   StaticJsonRpcProvider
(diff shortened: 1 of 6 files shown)


src/utils/provider.ts
```diff
@@ -1,3 +1,4 @@
+import 'dotenv/config';
```
ChaituVR, line 1: Not sure if we want to use this on front end, it works?

src/utils/provider.ts
```diff
@@ -10,7 +11,7 @@ export type ProviderOptions = {
   broviderUrl?: string;
 };
 
-const DEFAULT_BROVIDER_URL = 'https://rpc.snapshot.org';
+const DEFAULT_BROVIDER_URL = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
```
ChaituVR, line 14: Which project is asking for this?

PR #1090, opened 2024-11-29
feat: Add setBroviderUrl before getProvider

EthanOK:
Fixes # .

- Add setBroviderUrl

### DIFF
--- a/test/provider.ts
+++ b/test/provider.ts
@@ -1,0 +1,9 @@
+const snapshot = require('../');
+
+const provider_DEFAULT = snapshot.utils.getProvider('1');
+console.log(provider_DEFAULT.connection.url == 'https://rpc.snapshot.org/1');
+
+snapshot.utils.setBroviderUrl('https://rpc.custom.org');
+const provider_CUSTOM = snapshot.utils.getProvider('56');
+console.log(provider_CUSTOM.connection.url == 'https://rpc.custom.org/56');
+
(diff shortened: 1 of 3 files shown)


test/provider.ts
```diff
@@ -0,0 +1,9 @@
+const snapshot = require('../');
+
+const provider_DEFAULT = snapshot.utils.getProvider('1');
+console.log(provider_DEFAULT.connection.url == 'https://rpc.snapshot.org/1');
+
+snapshot.utils.setBroviderUrl('https://rpc.custom.org');
```
ChaituVR, line 6: You can actually use second param in getProvider to pass a different brovider URL, you tried?

EthanOK: I tried. 
I'm using getVp and validate, but I want to use my own provider rpc. https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/utils/vp.ts#L22
https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/validations/basic/index.ts#L11

I use  snapshot-strategies package, don't direct add second param in getProvider.
So:
```
import strategies from '@snapshot-labs/strategies';
import snapshot from "@snapshot-labs/snapshot.js";

// change custom rpc
snapshot.utils.setBroviderUrl('https://rpc.custom.org');
// It use my custom rpc in getVp 
strategies.utils.getVp(...)
```

PR #1114, opened 2025-01-30
feat: allow provider timeout customization

wa0x6e:
This PR adds the ability to set a custom timeout value for the provider


ChaituVR:
Not related but Tests are failing because ipfs.snapshot.box is slow loading ipfs for first time 🤔

PR #1115, opened 2025-01-30
fix: remove limits depending on space turbo status

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/315

This PR will remove all limits depending on the space `turbo` status.

This will effectively remove the length validation on the sdk, and defer those validation to sequencer.
This will also remove body lenght and strategies count limit on v1 UI (no impact on v2)

## TODO 

- [ ] Merge only after same validations have been added to sequencer


ChaituVR (approved):
looks good, will merge after sequencer to make sure it is ready to handle limits

PR #1117, opened 2025-02-01
fix: remove min length validation from space symbol

wa0x6e:
This PR will make the symbol property on Space truly optional, by removing the min length validation


bonustrack (approved):
utACK

PR #1134, opened 2025-03-16
feat: offchain validate name owner

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/485

This PR adds support for checking a domain owner offchain, using the domain DNS records for top level donain (
https://docs.ens.domains/registry/dns#offchain-verification-gasless), and resolved address for subdomain

For offchain subdomains, it will return the resolved address as owner, since offchain subdomain do not have the concept of owner/controller

Also edit the function signature to always return an address, to be combatible with sx

### Test

```js
const owner = await snapshot.utils.getEnsOwner(DOMAIN);
```

- It should return results as before for regular .eth domains
- It should return results for regular domains, using the dns txt records to prove ownership (like `defi.app`)
- It should return results for subdomain, such as `lucemans.cb.id`

Those offchain subdomains should resolved to these addresses

```json
{
    "defi.app": "0x7aeB96261e9dC2C9f01BaE6A516Df80a5a98c7eB",
    "lucemans.cb.id": "0x4e7abb71BEe38011c54c30D0130c0c71Da09222b",
    "lucemans.uni.eth": "0x225f137127d9067788314bc7fcc1f36746a3c3B5",
    "ulas.clv.eth": "0xd883E2d407361049dd3DACE2fF4278D75BC11d2A"
  }
```


src/utils.ts
```diff
@@ -639,14 +680,23 @@ export async function getEnsOwner(
     );
     owner = await ensNameWrapperContract.ownerOf(ensHash);
   }
+
+  if (owner === EMPTY_ADDRESS && domainType === 'other-tld') {
+    owner = await getDNSOwner(ens);
+  }
+
+  if (owner === EMPTY_ADDRESS && domainType === 'subdomain') {
+    owner = await provider.resolveName(ens);
```
ChaituVR, line 689: But this may affect existing ens subdomains, for example, https://app.ens.domains/2.snapspace.eth?tab=records owner can be different than the resolver, same as normal ens 🤔

src/utils.ts
ChaituVR, line 1: > Also edit the function signature to always return an address, to be combatible with sx

What is related to sx here? don't get it sorry 🙈 maybe I miss some context

ChaituVR (changes requested):

src/utils.ts
wa0x6e: Trying to refactor the sx code to use the function from the sdk to avoid duplicate code, see https://github.com/snapshot-labs/sx-monorepo/pull/1264

but thinking that maybe better to moving everything to stamp will be simpler

src/utils.ts
ChaituVR: ah got it 👍 

looks like it is throwing error on testnet
image
 
 Try with `console.log(await snapshot.utils.getEnsOwner('lucemans.cb.id', 11155111));` returning `0x0000000000000000000000000000000000000000` on master

src/utils.ts
```diff
@@ -219,6 +221,43 @@ ajv.addFormat('domain', {
...
 
+function getDomainType(domain: string): DomainType {
+  const isEns = domain.endsWith('.eth');
+
+  const tokens = domain.split('.');
+
+  if (tokens.length === 2 && !isEns) return 'other-tld';
+  else if (tokens.length > 2) return 'subdomain';
+  else if (isEns) return 'ens';
+  else throw new Error('Invalid domain');
```
ChaituVR, line 232: Returning `Uncaught (in promise) Error: Invalid domain` if i try `console.log(await snapshot.utils.getEnsOwner('abcd'));` maybe it should be `0x00` like before?

src/utils.ts
ChaituVR: If i send `console.log(await snapshot.utils.getEnsOwner('abcd.cb.id'));` it throws this error
image

src/utils.ts
```diff
@@ -219,6 +221,43 @@ ajv.addFormat('domain', {
...
+    {
+      headers: {
+        accept: 'application/dns-json'
+      }
+    }
+  );
+
+  const data = await response.json();
+  if (data.Status === 3) return EMPTY_ADDRESS;
+  if (data.Status !== 0) throw new Error('Failed to fetch DNS Owner');
```
ChaituVR, line 248: Maybe should return empty address in this case also maybe?

src/utils.ts
wa0x6e: Will finally not take the stamp route, and instead will implement the same function on sx, instead of importing this sdk.

Still need the return signature to be compatible with sx, for consistency

src/utils.ts
wa0x6e: The error throwing is to distinguish non-existent domain from fetch error, same as we don't rescue error on contract call failure, and just return an empty address

src/utils.ts
wa0x6e: > If i send `console.log(await snapshot.utils.getEnsOwner('abcd.cb.id'));` it throws this error image

This is from the coinbase offchain resolver contract, which throw an error instead of returning the empty address, like the resolver on uni.eth or linea.eth

Will contact coinbase about that

wa0x6e: Added tests for all mentioned scenario

src/utils.ts
```diff
@@ -639,14 +685,44 @@ export async function getEnsOwner(
...
+      owner = await provider.resolveName(ens);
+    } catch (e: any) {
+      // mute error from coinbase, when the subdomain is not found
+      // most other resolvers just return an empty address
+      const coinbaseError = e.message.includes(
+        'response not found during CCIP fetch'
+      );
+      // mute error from missing offchain resolver on sepolia
+      const sepoliaMissingResolverError =
+        e.message.includes('UNSUPPORTED_OPERATION') && network === '11155111';
```
ChaituVR, line 709: Is there some other way than hardcoding sepolia? this testnet can change time to time

ChaituVR: Maybe we can just ignore all unsupported operations irrespective of the network?

ChaituVR (commented):
Everything else is working well

src/utils.ts
```diff
@@ -30,6 +30,8 @@ interface Strategy {
   params: any;
 }
 
+type DomainType = 'ens' | 'tld' | 'other-tld' | 'subdomain';
+
```
ChaituVR, line 34:
```suggestion
type DomainType = 'ens' | 'tld' | 'other-tld' | 'subdomain';

const MUTED_ERRORS = [
  // mute error from coinbase, when the subdomain is not found
  // most other resolvers just return an empty address
  'response not found during CCIP fetch',
  // mute error from missing offchain resolver (mostly for sepolia)
  'UNSUPPORTED_OPERATION'
];
```

src/utils.ts
```diff
@@ -639,14 +685,43 @@ export async function getEnsOwner(
...
+    } catch (e: any) {
+      const mutedErrors = [
+        // mute error from coinbase, when the subdomain is not found
+        // most other resolvers just return an empty address
+        'response not found during CCIP fetch',
+        // mute error from missing offchain resolver (mostly for sepolia)
+        'UNSUPPORTED_OPERATION'
+      ];
+
+      if (mutedErrors.every((error) => !e.message.includes(error))) {
```
ChaituVR, line 710:
```suggestion
      if (MUTED_ERRORS.every((error) => !e.message.includes(error))) {
```

ChaituVR (approved):
tAck

PR #1138, opened 2025-04-01
feat: return space controller for .shib names

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/523

This PR will return the space controller for .shib names, for both their mainnet and testnet.

~~As of now, they don't have an official api for returning name owner, it's using the same api service as their name service lookup (https://www.shibariumscan.io/name-domains)~~

PR updated to use d3 (their name partner) api to return names info.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -716,9 +716,9 @@ export async function getEnsOwner(
   return owner || EMPTY_ADDRESS;
 }
 
-export async function getSpaceController(
+async function getEnsSpaceController(
   id: string,
-  network = '1',
+  network: string,
   options: any = {}
 ): Promise<string> {
   const spaceUri = await getSpaceUri(id, network, options);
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -735,6 +735,44 @@ export async function getSpaceController(
...
+export async function getShibariumNameOwner(
+  id: string,
+  network: string
+): Promise<string> {
+  if (!id.endsWith('.shib')) {
+    return EMPTY_ADDRESS;
+  }
+
+  const response = await fetch(
+    `https://bens.services.blockscout.com/api/v1/${network}/domains:lookup?name=${id}&only_active=true`,
```
ChaituVR, line 747: What is different from this service and API use in stamp https://github.com/stamp-labs/stamp/pull/362/files#diff-dbbfcf7aecae7147af6069bcbaabc0db16a1dccfb2878230de0771bea5aed776R7-R8 ?

wa0x6e:
This one is searching by domain, whereas in stamp, it’s searching by
address.

ChaituVR (approved):
tAck, but not sure how we will use this on sequencer, because we pass only default network here 
https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/settings.ts#L89-L91

wa0x6e:
> tAck, but not sure how we will use this on sequencer, because we pass only default network here https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/settings.ts#L89-L91

We would need to edit sequencer, to use shib network (109 and 157) when the name is ending by .shib, and default network otherwise

wa0x6e:
> tAck, but not sure how we will use this on sequencer, because we pass only default network here https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/settings.ts#L89-L91

Would be fixed by https://github.com/snapshot-labs/snapshot-sequencer/pull/527

wa0x6e:
You can merge and release a new version once https://github.com/snapshot-labs/snapshot-sequencer/pull/527 will be ready to be merged

wa0x6e:
PR updated to uses stamp, now depends on https://github.com/stamp-labs/stamp/pull/366

ChaituVR (approved):
tAck

PR #1149, opened 2025-05-20
feat: accept starknet address when fetching vp or validation

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/562

This PR allows starknet addresses when calling `get_vp`, `validate` and `getScores`


bonustrack (approved):
utACK

PR #1153, opened 2025-05-26
fix: accept both lowercase and checksum starknet address as valid

wa0x6e:
This PR loosen the starknet address validator, to accept both checksum (current) and lowercase (new) addresses as valid. It stil require that addresses are fully padded.,

This will align with the `evmAddress` validation, which is currently accepting both lowercase and checksum.

Without this patch, the validator will reject all checksum starknet address. In the ui side, we don't check for case when sending starknet address (like adding member in the space settings), so valid data on the UI will be rejected by sequencer.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -132,7 +132,9 @@ ajv.addFormat('evmAddress', {
 ajv.addFormat('starknetAddress', {
   validate: (value: string) => {
     try {
-      return validateAndParseAddress(value) === value;
+      return (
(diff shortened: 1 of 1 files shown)


bonustrack:
Is there a way to double vote if we support both casing? I believe in the hub db we have only checksummed address, not lowercased ones

wa0x6e:
> Is there a way to double vote if we support both casing? I believe in the hub db we have only checksummed address, not lowercased ones

We already support both casing for evm address

On the backend side, we have another compare after lowercasing everything

PR #1157, opened 2025-06-05
feat: add farcaster handle to space profile

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/565

Adds farcaster username to space profile

name validation from https://github.com/farcasterxyz/protocol/blob/main/docs/SPECIFICATION.md#5-fname-specifications : max 16 charaters, start by number or letter, and can only contain lowercase letter, number and hyphen

### DIFF
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -78,6 +78,12 @@
           "pattern": "^[A-Za-z0-9_-]*$",
           "maxLength": 39
         },
+        "farcaster": {
+          "type": "string",
+          "title": "farcaster",
+          "pattern": "^[a-z0-9][a-z0-9-]*$",
+          "maxLength": 16
+        },
         "email": {
           "type": "string",
           "title": "email",
--- a/src/schemas/space.json
+++ b/src/schemas/space.json
@@ -80,9 +80,7 @@
         },
         "farcaster": {
           "type": "string",
-          "title": "farcaster",
-          "pattern": "^[a-z0-9][a-z0-9-]*$",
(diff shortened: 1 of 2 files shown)


src/schemas/space.json
```diff
@@ -78,6 +78,12 @@
           "pattern": "^[A-Za-z0-9_-]*$",
           "maxLength": 39
         },
+        "farcaster": {
+          "type": "string",
+          "title": "farcaster",
+          "pattern": "^[a-z0-9][a-z0-9-]*$",
```
ChaituVR, line 84: I think we need to allow ens names as well, for example `alice.eth` as mentioned here https://docs.farcaster.xyz/learn/what-is-farcaster/usernames

ChaituVR (changes requested):

src/schemas/space.json
wa0x6e: Removed the regex pattern to allow any string (we don't have validation in place for validating ENS name)

bonustrack:
Can we add a max length of 32 too?

src/schemas/space.json
```diff
@@ -78,6 +78,10 @@
           "pattern": "^[A-Za-z0-9_-]*$",
           "maxLength": 39
         },
+        "farcaster": {
+          "type": "string",
+          "title": "farcaster"
+        },
```
ChaituVR, line 84: Some spam space can pass a huge string, should be fine?

ChaituVR (approved):
utAck

PR #1159, opened 2025-06-13
feat: Apechain Delegate Registry support

Sekhmet:
This will support Apechain Delegate Registry for offchain spaces.


wa0x6e (approved):
utAck

PR #1167, opened 2025-07-03
fix: use a dedicated key for brovider id

wa0x6e:
Brovider is using a special string for starknet id (`sn`), which is not the chain id, like all the other networks.

We're currently using the `key` property to pass that for starknet, but it's not compatible with a lot of implementation, which is using `key` as chain ID.

This PR will add a new dedicated `broviderId` key, for networks having a custon brovider id, and fallback to the network id when missing. The `key` will be back to be the chain ID

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.14.2",
+  "version": "0.14.3",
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
utACK

ChaituVR (approved):
utACk

PR #1171, opened 2025-07-16
feat: return owner of .sonic domains (#1168)

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/598

This PR improves the .sonic domain name owner 

- now poll the registry for the owner, instead of the NFT contract (registry address not in the docs, obtained via direct contact with ud)
- function has been refactored to be future-proof, and ready to handle more chain/tld from unstoppable domains

Updated the test, since the `boorger.sonic` domain has now changed owner (EOA wallet, instead of UD custody wallet)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -43,14 +43,17 @@ const ENS_ABI = [
   'function text(bytes32 node, string calldata key) external view returns (string memory)',
   'function resolver(bytes32 node) view returns (address)' // ENS registry ABI
 ];
-const SONIC_ABI = [
+const UD_MAPPING = {
+  '146': {
+    tlds: ['.sonic'],
+    registryContract: '0xde1dadcf11a7447c3d093e97fdbd513f488ce3b4'
+  }
+};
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -43,14 +43,17 @@ const ENS_ABI = [
   'function text(bytes32 node, string calldata key) external view returns (string memory)',
   'function resolver(bytes32 node) view returns (address)' // ENS registry ABI
 ];
-const SONIC_ABI = [
-  'function ownerOf(uint256 tokenId) external view returns (address address)'
+const UD_MAPPING = {
```
ChaituVR, line 46: UD is confusing; maybe we should just use `Unstoppable` directly everywhere

ChaituVR (approved):
utAck

src/utils.ts
bonustrack: Unstoppable is incomplete, I think UD is fine

PR #1172, opened 2025-07-28
feat: add timeout support to all fetch calls

wa0x6e:
All `fetch` calls now have a default timeout of 30s.

This can be disabled by setting timeout to `0`.

All functions using `fetch` will now timeout after 30s


ChaituVR (approved):
tAck

PR #1176, opened 2025-08-09
fix: fix operator precedence

wa0x6e:
This PR fix an operator precedence issue, which was returning scores as 0 instead of the existing value, when summing with `undefined`


ChaituVR (approved):
utAck

PR #1177, opened 2025-08-12
refactor: improve ranked choice voting readability and add test

wa0x6e:
This PR improves comprehension of the ranked choice voting system by adding more documentation, and add more test.

### Add more docs

Add more docs to some function, to help comprehension, and understand returned value.

### Add more tests

Add more tests, justifying that the fix https://github.com/snapshot-labs/snapshot.js/pull/1176 is really needed, and not just an edge case.

This fix is not related to invalid votes, etc..., the scenario can happen if the irv function pass on the first round, and the first round do no contain all the votes.

### Adjust `getChoiceString`  

Update function to handle invalid choice: they're now ignored, instead of showing `undefined`, like on the approval/quadratic/weighted voting system. 

This is consistent with the score computation, where those invalid votes are already ignored

### DIFF
--- a/src/voting/rankedChoice.ts
+++ b/src/voting/rankedChoice.ts
@@ -86,7 +86,7 @@ function irv(
  * // Candidate 2 wins with 150 total voting power (60+50+40 from 3 strategies)
  * // Candidate 1 has 120 total voting power (70+30+20 from 3 strategies)
  */
-function getFinalRound(
+export function getFinalRound(
   votes: RankedChoiceVote[]
 ): [string, [number, number[]]][] {
   const rounds = irv(
--- a/src/voting/rankedChoice.ts
+++ b/src/voting/rankedChoice.ts
@@ -66,6 +66,26 @@ function irv(
       );
 }
 
+/**
+ * Runs the complete Instant Runoff Voting (IRV) algorithm and returns the final results.
+ *
+ * Executes all elimination rounds until a winner is determined or fewer than 3 candidates remain.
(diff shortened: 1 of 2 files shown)


wa0x6e:
Not an important PR, don't need a release after merge, can wait next fix/feature

src/voting/rankedChoice.ts
```diff
@@ -148,16 +187,57 @@ export default class RankedChoiceVoting {
...
+   * // Returns: "(1st) Alice" - invalid choice 5 is filtered out
+   */
   getChoiceString(): string {
     return this.selected
-      .map((choice) => {
-        if (this.proposal.choices[choice - 1])
-          return this.proposal.choices[choice - 1];
-      })
+      .map((choice) => this.proposal.choices[choice - 1])
+      .filter(Boolean)
```
ChaituVR, line 240: Should be fine with reduced array length?

ChaituVR (approved):
tAck

src/voting/rankedChoice.ts
ChaituVR: We are filtering some choices here (before we are not filtering) so there will be difference in array length, i think it's fine but want to double check because PR is just refactor 😅

wa0x6e: This is a little inconsistency I found while writing tests.

We are filtering out invalid choices everywhere, but single choice.

In which case this can break ? I don't think we check for arrays length anywhere before calling this function

ChaituVR: Okies

PR #1180, opened 2025-09-03
fix: parseStarknetResult to handle multiple outputs

ChaituVR:
AI gave this fix but it is straight forwarrd, but testing it is the main task

### Summary:
- Right now starknet multicall is parseing only first param in the output
- This PR will handle all the params

### How to test:
- Checkout this branch
- Run `npm test -- --run test/integration/multicall/starknet-contracts.spec.ts`
- Notice  in the logs that all params in output are handled
- Run `git restore -s origin/master  -- src/multicall/starknet.ts` to checkout master branch's multicall
- Comment out all `expect` in `test/integration/multicall/starknet-contracts.spec.ts`
- Run `npm test -- --run test/integration/multicall/starknet-contracts.spec.ts`
- Notice that all the params except first param are `0x0`


test/integration/multicall/starknet-contracts.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    );
+    console.log('='.repeat(60));
+    console.log(
+      '\n🎉 All contract calls completed successfully via multicall!'
+    );
+
+    // Final assertions
+    expect(identityString).toBe('Staking Delegation Pool');
+    expect(poolMemberInfoResult[0]).toBeDefined(); // reward_address exists
+    expect(stakerInfoResult[0]).toBeDefined(); // reward_address exists
```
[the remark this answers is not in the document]
wa0x6e: Maybe testing the call on `contract_parameters_v1` and `get_epoch_info` would be enough to ignore the block number.

We could simplify the test, by calling only on 2 different contracts, on the 2 functions above (or any functions on the contract returning static data), and test the response using `toMatchSnapshot`

We should not test the staking contract here, but just that calling at least 2 different contracts works, and that it formats the response correctly.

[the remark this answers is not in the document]
ChaituVR: Still keeping get_staker_info_v1 to test objects that have object inside 
```
{
    "Some": {
        "reward_address": "0xd3b910d8c528bf0216866053c3821ac6c97983dc096bff642e9a3549210ee7",
        "operational_address": "0x6012516a3ae0e8e8367abdb1db76ba56f7cb221aa3c1e4c02f52ab9d4d1ebc6",
        "unstake_time": {},
        "amount_own": "0x175a7be10a64c27440000",
        "unclaimed_rewards_own": "0x24bfe412f5bcc587a656",
        "pool_info": {
            "Some": {
                "pool_contract": "0x2cb02c72e8a0975e69e88298443e984d965a49eab38f5bdde1f5072daa09cfe",
                "amount": "0x40939c74aa59398193cee8",
                "commission": "0x0"
            }
        }
    }
}
```

test/integration/multicall/starknet-contracts.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+          }
+        }
+        return str || hexValue;
+      } catch (e) {
+        return hexValue;
+      }
+    }
+    return hexValue;
+  }
+
```
wa0x6e, line 62:
```suggestion
```

test/integration/multicall/starknet-contracts.spec.ts
```diff
@@ -0,0 +1,126 @@
+// To run this test:
+// npm test -- --run test/integration/multicall/starknet-contracts.spec.ts
+
+import { describe, it, expect } from 'vitest';
+import multicall from '../../../src/multicall/starknet';
+import getProvider from '../../../src/utils/provider';
+
+describe('Starknet Contract Reader Integration', () => {
```
wa0x6e, line 8:
```suggestion
describe('Starknet multicall', () => {
```

test/integration/multicall/starknet-contracts.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+        { type: 'core::integer::u256' }, // unclaimed_rewards_own
+        { type: 'core::integer::u8' }, // pool_info variant (0x0 = None, 0x1 = Some)
+        { type: 'core::starknet::contract_address::ContractAddress' }, // pool_contract
+        { type: 'core::integer::u256' }, // pool amount
+        { type: 'core::integer::u64' } // pool commission
+      ]
+    }
+  ];
+
+  it('should call identify, get_pool_member_info_v1, and get_staker_info_v1 using multicall', async () => {
```
wa0x6e, line 99:
```suggestion
  it('should return responses from different contract calls', async () => {
```

src/multicall/starknet.ts
```diff
@@ -17,67 +17,78 @@ function parseStarknetResult(rawResult: string[], functionAbi: any): any {
...
-        return rawValue;
-
-      // Byte array
-      case 'core::bytes_31::bytes31':
-        return rawValue;
-
-      default:
-        // Return raw value for unknown types
-        return rawValue;
+    for (let outputIndex = 0; outputIndex < outputs.length; outputIndex++) {
```
wa0x6e, line 26: Are we fine with having this outside the loop ?

This will return the rawResult if any of the parsing fail (will never return partially parsed results)

test/integration/multicall/__snapshots__/starknet-contracts.spec.ts.snap
```diff
@@ -0,0 +1,30 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Starknet Contract Reader Integration > should call identify, get_pool_member_info_v1, and get_staker_info_v1 using multicall 1`] = `
```
ChaituVR, line 3:
```suggestion
exports[`Starknet multicall > should return responses from different contract calls 1`] = `
```

PR #1183, opened 2025-09-16
feat: support starknet multisig signature

wa0x6e:
Continuing on https://github.com/snapshot-labs/snapshot.js/pull/1182

Continuing on the previous PR, which was adding support for signature with a Guardian,
this PR will add support for signature from argent X multisig wallets.

This PR is refactoring the way we verify a message, by using the provider native function, instead of calling the contract method, which will simplify the process of formatting the signature array before usage.

The code should now be able to verify all starknet signature format

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.14.12",
+  "version": "0.14.13",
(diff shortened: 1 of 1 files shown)


ChaituVR (changes requested):
utAck

PR #1196, opened 2026-04-09
fix: update Ethereum Classic network config

diega:
- Update multicall to Multicall3 canonical address (0xcA11bde05977b3631167028862bE2a173976CA11)
- Update RPC endpoint (ethercluster.com is no longer operational)
- Update Blockscout explorer URL

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.14.21",
+  "version": "0.14.22",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
--- a/src/networks.json
+++ b/src/networks.json
@@ -176,12 +176,12 @@
     "shortName": "Ethereum Classic",
     "chainId": 61,
     "network": "mainnet",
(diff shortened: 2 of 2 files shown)


diega:
Hi @ChaituVR, could this PR be reopened and reviewed? It was closed without any comment, but the ETC config it fixes is **currently broken on `main`**:

- The RPC `https://www.ethercluster.com/etc` is **dead** — it no longer responds (connection fails). Any Snapshot strategy on ETC (chainId 61) can't read voting power because there's no working RPC.
- The `multicall` address is updated to the canonical **Multicall3** (`0xcA11bde05977b3631167028862bE2a173976CA11`), which is verified deployed on ETC.
- The explorer URL is updated to `https://etc.blockscout.com` (the old `blockscout.com/etc/mainnet` now just redirects there).

Since `networks.json` here is the canonical registry consumed by snapshot.org, this also breaks downstream integrations that rely on it — e.g. the Snapshot Safe App embedded in Safe Wallet, where selecting ETC currently fails to compute voting power.

The diff is 3 lines and the values are verifiable on-chain. Happy to add a working public RPC alternative if `etc.rivet.link` isn't preferred. Thanks!

ChaituVR (approved):

PR #1200, opened 2026-05-13
fix(multicall): decode legacy bytes32 string returns

bonustrack:
## Summary
- Legacy tokens (MKR, SAI, AVT-era) declare `string` for `name()`/`symbol()` in their ABI but actually return `bytes32`. With the standard ERC20 ABI, `decodeFunctionResult` throws and fails the **entire** multicall batch — even if only one of N tokens is affected.
- Adds a narrow decode-side fallback: when strict decode fails **and** the function declares a `string` output, retry by decoding those outputs as `bytes32` and converting via `parseBytes32String`.
- Encode side is untouched (intentionally — silent `string → bytes32` padding would be a footgun for any strategy using `bytes32` keys/IDs).

## Why this matters for Snapshot
Strategies that touch a long list of tokens (e.g. multi-token balance aggregators, indexers) currently break if any single token in the list is a legacy `bytes32`-symbol token. With this change, one bad token in a 200-token batch no longer poisons the whole batch.

## Behavior matrix
| Case | Before | After |
|---|---|---|
| Strict decode succeeds | works | identical |
| Decode fails, function has no `string` output | throws original error | throws original error |
| Decode fails, function has `string` output, bytes32-fallback works (MKR/AVT) | **throws, batch fails** | returns decoded string |

## Test plan
- [x] Manually verified against UNI, ENS, AAVE (standard) + AVT, MKR (legacy `bytes32` symbol) — all five fetch correctly in one batch
- [ ] CI typecheck + lint
- [ ] Reviewer sanity-check on strategies that read token metadata in batches

cc @ChaituVR

### DIFF
--- a/src/multicall/evm.ts
+++ b/src/multicall/evm.ts
@@ -1,4 +1,5 @@
 import {
+  defaultAbiCoder,
   Fragment,
   FunctionFragment,
   Interface,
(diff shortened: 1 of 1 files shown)


ChaituVR:
@copilot+claude-opus-4.7 Check if this effect any existing strategy on https://github.com/snapshot-labs/score-api

PR #1203, opened 2026-06-05
fix(ens): resolve text records resolver-agnostically

ChaituVR:
## Summary

`getEnsTextRecord` matched the name's resolver against a hardcoded `ensResolvers` allowlist and returned `null` for anything else. So names on newer / **ENS v2** resolvers resolved to empty — breaking `snapshot` record reads (and any controller derived from them, e.g. via `getSpaceController` / `getSpaceUri`) for those names.

## Fix

Resolve the name's **actual** resolver from the registry and read the text record from it directly, instead of matching an allowlist — so any resolver (including ENS v2 and future ones) works.

Reverts on `text()` (CCIP-read, ENS v2, or broken resolvers) are caught and treated as "no record", so `getSpaceUri` / `getSpaceController` keep falling back to the name owner instead of surfacing an error — the same graceful behavior `getSpaceUri`'s existing `try/catch` already provided.

## Context

This is the server-side companion to [snapshot-labs/sx-monorepo#2143](https://github.com/snapshot-labs/sx-monorepo/pull/2143), which makes the equivalent change in the UI (`apps/ui/src/helpers/ens.ts`). The sequencer's "not allowed" controller check resolves via **this** function, so the UI fix alone doesn't change the server's decision for ENS-v2 names that delegate control via a `snapshot` text record — this PR closes that gap so UI and sequencer agree.

## Test plan

Verified locally against live RPC (Sepolia + mainnet):

- `bob.eth` (Sepolia, reverting non-allowlisted resolver) → record `null`, `getSpaceController` falls back to owner `0x179A86…` (no throw).
- `ens.eth` (Sepolia, allowlisted resolver) → `""` (unchanged).
- `fabien.eth` (mainnet) → reads its `snapshot` record via the actual resolver.

`yarn typecheck`, `yarn lint`, and the `getEns*` specs pass.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -615,12 +615,7 @@ export async function getEnsTextRecord(
   network = '1',
   options: any = {}
 ) {
-  const {
-    ensResolvers = networks[network]?.ensResolvers ||
-      networks['1'].ensResolvers,
-    broviderUrl,
-    ...multicallOptions
-  } = options;
+  const { broviderUrl } = options;
 
   let ensHash: string;
 
@@ -632,25 +627,23 @@ export async function getEnsTextRecord(
 
   const provider = getProvider(network, { broviderUrl });
 
-  const calls = [
-    [ENS_REGISTRY, 'resolver', [ensHash]], // Query for resolver from registry
-    ...ensResolvers.map((address: string) => [
-      address,
+  const resolverAddress: string = await call(provider, ENS_ABI, [
+    ENS_REGISTRY,
+    'resolver',
+    [ensHash]
+  ]);
+
+  if (!resolverAddress || resolverAddress === EMPTY_ADDRESS) return null;
+
+  try {
+    return await call(provider, ENS_ABI, [
+      resolverAddress,
       'text',
       [ensHash, record]
-    ]) // Query for text record from each resolver
-  ];
-
-  const [[resolverAddress], ...textRecords] = (await multicall(
-    network,
-    provider,
-    ENS_ABI,
-    calls,
-    multicallOptions
-  )) as string[][];
-
-  const resolverIndex = ensResolvers.indexOf(resolverAddress);
-  return resolverIndex !== -1 ? textRecords[resolverIndex]?.[0] : null;
+    ]);
+  } catch (e: any) {
+    return null;
+  }
 }
 
 export async function getSpaceUri(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,7 +2,8 @@ import crossFetch from 'cross-fetch';
 import { Contract } from '@ethersproject/contracts';
 import { getAddress, isAddress } from '@ethersproject/address';
 import { parseUnits } from '@ethersproject/units';
-import { namehash, ensNormalize } from '@ethersproject/hash';
+import { namehash, ensNormalize, dnsEncode } from '@ethersproject/hash';
+import { Interface } from '@ethersproject/abi';
 import { jsonToGraphQLQuery } from 'json-to-graphql-query';
 import Ajv from 'ajv';
 import addFormats from 'ajv-formats';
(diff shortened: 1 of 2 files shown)


tony8713:
## Review: resolver-agnostic ENS text records

Did a careful pass against the ENS docs + the surrounding code. **Net: this is a correct, net-positive fix** — the allowlist removal is the right call and I do *not* think it broadly breaks existing ENS cases. A couple of real (non-blocking) concerns below, mostly around the sequencer's controller decision.

### What it does
`getEnsTextRecord` (`src/utils.ts:612-654`) goes from one batched multicall (registry `resolver()` + a fixed `text()` against every allowlisted resolver, then index-match) to two sequential single calls: registry `resolver(node)` → the name's *actual* resolver, then `text(node, key)` on it, with `try/catch → null`. The `ensResolvers` allowlist param is dropped.

### Why it's correct / not a regression
- **Allowlist removal is the point.** Old code returned `null` for any name whose resolver wasn't 1 of the 2 hardcoded mainnet (1 Sepolia) addresses — including ENS v2 / custom resolvers. Real bug for `snapshot` / `getSpaceUri` / `getSpaceController`.
- **Return shape is now *more* correct.** Old returned `textRecords[resolverIndex]?.[0]` (multicall-decoded tuple, array-indexed). New returns `call(...)` directly, which for a single-return fn is the scalar string — which is exactly what consumers want (`isAddress(spaceUri)`, `spaceUri.split('/')` in `getEnsSpaceController`).
- **CCIP-read / offchain names (`*.cb.id`, `*.uni.eth`) — no regression.** ENS docs confirm an offchain resolver's `text()` reverts with `OffchainLookup`. `call()` (`src/utils.ts:257`) is a plain `Contract` method on `StaticJsonRpcProvider`; ethers v5 does **not** auto-follow CCIP-read on raw contract calls. So these revert → caught → null. The *old* path behaved identically: `multicall` uses `aggregate` (`src/multicall/evm.ts:31`), which is all-or-nothing, so an `OffchainLookup` revert in the batch reverted the whole multicall → threw → `getSpaceUri`'s try/catch → null. Same end state, now with per-name isolation. Offchain `snapshot` records never worked through this path before and still don't — that's a separate feature, not a break.
- **Forward-confirmation isn't relevant here.** That guard protects *reverse* (addr→name) resolution against spoofing. This is purely forward text-record reads, and snapshot.js has no reverse/`getNames` path at all (that's in sx-monorepo). No spoofing surface introduced.
- **Wildcard/ENSIP-10 with no registry record:** registry `resolver(node)` → `0x0` → early null. Old allowlist code also returned null (`0x0` ∉ allowlist). No change.

### Concerns worth addressing (not blockers)
1. **Over-broad error swallowing (`src/utils.ts:643-649`).** The bare `catch → null` treats *transient* infra errors (RPC timeout, rate-limit, network blip) identically to "no record." Since `getEnsSpaceController` falls back to the **name owner** when the record is null, a flaky RPC can transiently flip the sequencer's "who controls this space" decision. The old all-or-nothing multicall threw on transient errors too, but those propagated; here they're silently coerced. Suggest distinguishing an actual execution revert / `OffchainLookup` from infra errors (e.g. inspect `e.code`/`e.message`) and only returning null for the former — rethrow/log infra errors so the existing `getSpaceUri` try/catch can log them.
2. **Latency / RPC load.** 1 batched call → 2 **sequential** RPCs (text depends on resolver). At sequencer volume this is a measurable round-trip + load increase. Acceptable, but worth being aware of; can't easily batch since the second target is dynamic.

### Recommended tests before merge
- A name on a **non-allowlisted / ENS v2 resolver** with a real `snapshot` record → reads correctly (the case this fixes).
- An **offchain/CCIP name** (`*.cb.id`) → returns null, `getSpaceController` falls back to owner, no throw.
- A name whose registry resolver is `0x0` → null.
- Simulate an **RPC timeout** mid-`text()` and confirm the resulting controller decision is what you intend (this is the risky one given #1).

Verdict: looks correct and safe to ship; please tighten the catch in #1 (or at least log) so a transient RPC failure can't silently change a controller decision.

src/utils.ts
```diff
@@ -632,25 +642,47 @@ export async function getEnsTextRecord(
...
-    calls,
-    multicallOptions
-  )) as string[][];
-
-  const resolverIndex = ensResolvers.indexOf(resolverAddress);
-  return resolverIndex !== -1 ? textRecords[resolverIndex]?.[0] : null;
+    ]);
+  } catch (e: any) {
+    return null;
+  }
```
wa0x6e, line 685: Narrow this catch so only deterministic resolver reverts map to `null`. A bare `catch (e) => null` also swallows transient RPC/network/timeout errors, making infra failure indistinguishable from "no record" — downstream `getSpaceController` then falls back to the name owner (fail-open) during an outage. Re-throwing non-`CALL_EXCEPTION` errors keeps the ENS-v2/CCIP revert handling while failing closed on infra blips. (Same shape as the UI fix in snapshot-labs/sx-monorepo#2143.)

```suggestion
  } catch (e: any) {
    // The resolver reverts on text() (CCIP-read, ENS v2, or a broken
    // resolver). Treat as "no record" so callers degrade gracefully instead
    // of throwing. Network/timeout errors are re-thrown so a transient RPC
    // failure is not silently read as "no record".
    if (e?.code === 'CALL_EXCEPTION') return null;
    throw e;
  }
```

The Universal Resolver catch above (lines 662–664) has the same fail-open shape — same narrowing applies, though watch that an unfollowed `OffchainLookup` may not surface as `CALL_EXCEPTION` there, so confirm before tightening that one.

wa0x6e: Aside from that, it's working as intended

ChaituVR:
Superseded by #1227 (text records via the Universal Resolver, resolver-agnostic + CCIP/ENSv2 aware) and the strict-mode follow-ups #1231/#1232. Closing.

PR #1210, opened 2026-07-01
fix(copeland): break ties by average support

alextnetto:
Choices tied on pairwise victories get identical scores today, so their order is decided by choice-list position.

This adds the Average Support tiebreak from the [ENS RFP](https://discuss.ens.domains/t/rfp-for-improving-ranked-choice-voting-support-in-snapshot/19625): voting power a choice received across its matchups, averaged over the matchups. Tied choices are ordered by it.

**Why 0.5**

Any multiplier in (0, 1) yields the same ranking. Order is decided by victories first, then Average Support; the multiplier only sets how far apart tied choices sit in the normalized score, never who wins.

Victory scores are whole numbers, so distinct totals differ by at least 1. Average Support is in [0, 1], so a 0.5 multiplier caps the bonus at 0.5, half of that minimum gap. This guarantees the bonus only reorders genuine ties and can never overturn a victory. It also mirrors the 0.5 half-point Copeland already assigns to a pairwise tie.

**Scope**

`getScores()` only. Total-voting-power sum preserved. `getScoresByStrategy()` unchanged.


ChaituVR (approved):
tAck

tony8713 (approved):
Reviewed the rebase plus the by-strategy follow-up. netto's getScores is byte-identical to his original #1210 commit (only rebase context shifted the lines), isValidChoice matches merged #1214, and the new average-support logic is confined to getScoresByStrategy, mirroring his bonus per strategy column. What I verified:

- getScores diffs clean against 1ea2f17 — the tiebreak semantics are exactly netto's, untouched.
- Corpus dry-run over all 18 historical Copeland proposals: the bonus never overturns a pairwise-victory ordering, it only reorders choices already tied on victories (actually engaged in 4, including the Term 7 Meta-Gov and SPP2 Season II elections).
- Live Term 7 Meta-Gov reproduction matches the certified per-candidate results to 4 decimals (total 1,444,708.16).
- By-strategy columns sum back to the headline only when strategies are sign-collinear; that scope is now stated in the test name and comment, with a counterexample spec pinning the inherent divergence (totals still preserved either way).
- Single-choice (n=1) path traced and run: the `matchupsPerChoice > 0` guard is reachable (no early return shields it) and prevents a 0/0 NaN.
- Full src/voting suite green (113 tests), tsc and lint clean.

LGTM. Good to fast-track ahead of the ENS election.

PR #1214, opened 2026-07-07
fix(copeland): reject partial ballots (revives #1133)

tony8713:
Copeland currently treats any choice a voter leaves out as "no opinion", so a partial or bullet ballot ends up being meaningless and quietly gameable. Those partial ballots can only be cast through the API since the UI always submits a full ranking, which makes it an API-only asymmetry. This makes Copeland require a full ranking of every choice, the same way ranked choice already validates. ENS asked for this and Less gave the go-ahead in the Copeland thread.

This revives #1133, which made the same one-line fix but shipped no tests and would have failed CI on the old snapshots.

Note the behavior change flows through `examples/copeland.json`: vote #4 is a partial `[2, 1]` ballot that now gets filtered, so the scores/totals reflect 3 valid ballots instead of 4.

## Test plan

- `yarn test:once src/voting/copeland.spec.js` and the wider `src/voting` suite pass.
- The old "Partial ranking" test is rewritten to assert a partial ballot is filtered out and does not change scores or totals.
- Regenerated the copeland snapshots to match the filtered result.


ChaituVR (approved):
utAck

PR #1217, opened 2026-07-27
feat: ENSv2-ready name resolution via Universal Resolver and viem

yashgo0018:
## Summary

Prepares snapshot.js for ENSv2 by routing all ENS resolution through the Universal Resolver and migrating the ENS helpers from ethers v5 to viem. Added support for Universal Resolver (`0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe`),  so snapshot.js keep working unchanged when v2 goes live. Readiness is verified by the official integration check (`ur.integration-tests.eth` → `0x2222...2222`), which now runs in the e2e suite.

### ENSv2 groundwork 

- Added `ensUniversalResolver` to `networks.json` for mainnet and Sepolia.
- `getEnsOwner` now prefers `UniversalResolver.findOwner(dnsEncodedName)` (the v2-native ownership lookup that traverses the hierarchical registry) and only falls back to the v1 registry + NameWrapper unwrap when it returns nothing. This is needed for unmigrated names during the transition.
- `getEnsTextRecord` resolves through the Universal Resolver instead of querying resolvers directly.
- Added strict DNS packet encoding (`dnsEncodeName`) with empty/oversized label validation.

### viem migration of the ENS helpers 

- New memoized `getViemClient()` in `src/utils/provider.ts`, mirroring the existing brovider URL and timeout behavior.
- `getEnsTextRecord` uses viem `getEnsText`; `getEnsOwner` uses viem `readContract` for `findOwner`, the v1 registry fallback, and the NameWrapper unwrap, and viem `getEnsAddress` for the DNS-TLD and subdomain fallbacks.
- This replaces ethers v5 `provider.resolveName`, which bypassed the Universal Resolver and would fail for names upgraded to v2-native resolvers. viem handles CCIP-Read including the ENSIP-21 batch gateway.
- Name preprocessing (`normalize`, `namehash`) now comes from `viem/ens` (ENSIP-15); the `MUTED_ERRORS` ethers error-matching workaround is gone since viem returns `null` for unresolvable names.
- Removed the unused, pre-ENSIP-15 `@ensdomains/eth-ens-namehash` dependency. Non-ENS code (multicall, sign, verify) stays on ethers v5.

### Node 18 + readiness test

- `engines.node` bumped from `>=16` to `>=18` (viem requires native `fetch`; Node 16 is EOL). CI workflows pinned to Node 16 (typecheck, npm-publish, update-network, update-minor-version) now use Node 18.
[...]

PR #1220, opened 2026-08-05
chore: require Node.js >= 18

ChaituVR:
## What

Bumps the minimum supported Node.js version from 16 to 18:

- `package.json` engines: `>=16` → `>=18`
- CI workflows still installing Node 16 (`typecheck`, `npm-publish`, `update-minor-version`, `update-network`) → Node 18

## Why

- Node 16 reached end-of-life in September 2023 and no longer receives security updates.
- The test workflow already runs exclusively on Node 20 and 22, so Node 16 hasn't actually been exercised by CI for a while — engines just didn't reflect reality.
- Upcoming ENS work (#1217 and follow-ups) depends on viem, which requires Node >= 18. Landing the engines bump separately keeps that breaking change isolated and easy to revert or reason about.

⚠️ Note for consumers: this is a breaking change for anyone still installing snapshot.js on Node 16 — worth a callout in the release notes.

## How to test

**Before (master):**
1. `grep -rn "node-version" .github/workflows/` → `typecheck`/`npm-publish`/`update-minor-version`/`update-network` install Node 16.x
2. `package.json` engines says `>=16` while the test matrix runs 20/22

**After (this branch):**
1. Same grep → all workflows on Node 18+, consistent with the 20/22 test matrix
2. `yarn && yarn typecheck && yarn lint && yarn test:once` → all green (verified locally: 23 files, 340/340 tests)
3. No dependency or source changes — `git diff master --stat` shows only the 5 config lines

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "prepublishOnly": "yarn build"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "files": [
     "dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.14.27",
+  "version": "0.15.0",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: 18
           registry-url: https://registry.npmjs.org/
       - run: yarn install --frozen-lockfile
       - run: npm publish
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['20', '22']
+        node-version: ['18', '20', '22']
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "prepublishOnly": "yarn build"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
```
tony8713, line 64: Worth knowing where this lands: yarn classic hard-errors on an engines mismatch where npm only warns (reproduced by installing a `>=18` pack of this branch on 16.20.2), and snapshot-v1 still builds on Node 16 with yarn. Nothing breaks today since its lockfile pins 0.14.18 and dependabot ignores `@snapshot-labs/*`, but its `Update Snapshot packages` workflow runs `yarn upgrade @snapshot-labs/snapshot.js --latest` on Node 16 and will fail from the next release on. Since releases go out via `npm version patch` that arrives as 0.14.28, so a minor bump would at least make the drop visible.

tony8713 (approved):
Approving. I checked this against real Node runtimes rather than reading the diff, and it holds up.

Fresh install of the branch on Node 18.20.8: typecheck, lint, build and 340/340 tests all green, so `>=18` is an honest floor. I also built `dist/` from the same commit under Node 16 and under Node 18, and the three bundles come out byte-identical (same sha256 for cjs, esm and min), so the `npm-publish.yml` bump doesn't change what consumers actually receive.

One correction to the framing: master still passes 340/340 on Node 16.20.2, so `>=16` wasn't a stale claim, it was true. This is a deliberate drop of a working platform rather than engines catching up with reality, which is probably how it should read in the release notes.

The one thing I'd add: the test matrix is still `['20', '22']`, so nothing exercises the new floor and the same drift restarts from 18. Adding `'18'` to `test.yml` closes that and it's free, I ran the suite on 18 and it's green in ~36s.

PR #1218, opened 2026-07-30
chore(deps): drop unused @ensdomains/eth-ens-namehash

tony8713:
## Summary

Removes `@ensdomains/eth-ens-namehash` from `dependencies`. It is declared but **not imported anywhere** in the repo — ENS name preprocessing already goes through `namehash` / `ensNormalize` from `@ethersproject/hash` (ENSIP-15 compliant; `eth-ens-namehash` predates ENSIP-15).

No source changes, no API changes: `package.json` −1 line, `yarn.lock` −5 lines, 0 additions.

## Why a separate PR

This is one of the changes in #1217 (ENSv2 name resolution via Universal Resolver) that stands on its own against `master`. The dependency is already dead **today**, independently of anything ENSv2 needs, so it does not need to ride along with a large feature PR. Splitting it out reduces the diff and review surface of #1217.

## Verification

```
$ grep -rn "ensdomains\|eth-ens-namehash" . --exclude-dir=node_modules --exclude-dir=dist --exclude-dir=.git --exclude=yarn.lock
package.json:11:    "@ensdomains/eth-ens-namehash": "^2.0.15",
```

The only hit in the whole repo is the `package.json` declaration itself — no `src/`, `test/`, `scripts/` or build-config reference. The package declares no dependencies of its own, so nothing else is pruned from the lockfile.

Gates, run on Node 18:

```
$ yarn install --frozen-lockfile     # lockfile/package.json consistent
Done in 7.14s.
$ yarn lint                          # eslint . --ext .ts
Done in 4.19s.
$ yarn typecheck                     # tsc --noEmit
Done in 3.16s.
$ yarn build
created dist/snapshot.min.js in 45.6s
created dist/snapshot.cjs.js, dist/snapshot.esm.js in 5.2s
Done in 52.32s.
```

Bundle sanity — zero references to the removed package, ENS helpers still exported:

```
$ grep -c "eth-ens-namehash" dist/snapshot.cjs.js dist/snapshot.esm.js dist/snapshot.min.js
dist/snapshot.cjs.js:0
dist/snapshot.esm.js:0
dist/snapshot.min.js:0

$ node -e "const s = require('./dist/snapshot.cjs.js'); ..."
getEnsOwner: function | getEnsTextRecord: function | getSpaceUri: function | getSpaceController: function
```

[...]

PR #1221, opened 2026-08-06
chore: add viem with client factory and build support

ChaituVR:
## What

Infrastructure groundwork for ENS v2 / Universal Resolver support (planned successor to #1217), with no production code paths touched:

- **viem `^2.55.11`** — the library ENS's v2-readiness guidance is built around: automatic CCIP-Read (EIP-3668) handling and first-class Universal Resolver support. Requires the Node ≥ 18 floor that landed in #1220.
- **Rollup: dependency subpath imports treated as external.** `external` was an exact-name array, so an import like `viem/ens` would get bundled into the published cjs/esm dist instead of staying a runtime `require`. Now any subpath of a declared dependency is external (general fix, not viem-specific).
- **Rollup: `inlineDynamicImports` for the UMD build.** viem's CCIP-Read module loads via dynamic `import()`; rollup's default is code-splitting, which the single-file UMD format cannot do.
- **`getViemClient()`** in `src/utils/provider.ts` — memoized viem `PublicClient` per network, same brovider URL + timeout conventions as the existing `getProvider()`. Deliberately chain-less so the Universal Resolver address always comes from our config, never from viem's chain defaults (calls without an explicit address fail loudly).
- **Tests**: integration tests for `getViemClient` mirroring the `getProvider()` suite, plus the official [ENSv2 readiness canary](https://docs.ens.domains/web/ensv2-readiness/) as an e2e test — `ur.integration-tests.eth` must resolve to `0x2222…2222` through the Universal Resolver (`0x1111…1111` would mean the resolution stack bypasses it).
- **Test infra**: 30s timeouts on two slow live-RPC blockfinder e2e tests (from #1217).

## Why a separate PR

Keeps the upcoming ENS PR focused purely on resolution behavior (config + `utils.ts`), while the dependency/build/plumbing surface gets reviewed on its own. Nothing here changes runtime behavior for any consumer.

## How to test

**Before (master):**
1. `grep -n "external" rollup.config.js` → exact-name array; a `viem/ens` import would be bundled into dist
2. No viem anywhere; no ENSv2 readiness check in the suite

**After (this branch):**
1. `yarn && yarn typecheck && yarn lint && yarn build` → all green; `dist/snapshot.cjs.js` contains no bundled viem code (subpath externals working)
2. `yarn test:once` → full suite green including the new `getViemClient` tests and the live ENSv2 readiness canary against mainnet
3. `git diff master --stat` → build/plumbing/test files only; `src/utils.ts` untouched

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "cross-fetch": "^3.1.6",
     "json-to-graphql-query": "^2.2.4",
     "lodash.set": "^4.3.2",
-    "starknet": "^6.11.0"
+    "starknet": "^6.11.0",
+    "viem": "^2.55.11"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^18.1.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -11,7 +11,10 @@ import pkg from './package.json';
 
 const name = 'snapshot';
 const input = 'src/index.ts';
-const external = [...Object.keys(pkg.dependencies || {})];
+const dependencies = [...Object.keys(pkg.dependencies || {})];
+// match subpath imports too (e.g. viem/ens)
+const external = (id) =>
+  dependencies.some((dep) => id === dep || id.startsWith(`${dep}/`));
 
 export default [
   {
@@ -22,6 +25,9 @@ export default [
         name,
         file: pkg.browser,
         format: 'umd',
+        // viem uses dynamic import() internally (CCIP-Read module), which
+        // rollup would otherwise code-split — unsupported in UMD bundles
+        inlineDynamicImports: true,
         intro:
           'var global = typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {}'
       }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -1,6 +1,5 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import { RpcProvider } from 'starknet';
-import { createPublicClient, http, PublicClient } from 'viem';
 import networks from '../networks.json';
 
 export interface ProviderOptions {
@@ -10,7 +9,7 @@ export interface ProviderOptions {
 
 type ProviderInstance = StaticJsonRpcProvider | RpcProvider;
 
-type ProviderType = 'evm' | 'starknet';
+export type ProviderType = 'evm' | 'starknet';
 
 const DEFAULT_BROVIDER_URL = 'https://rpc.snapshot.org' as const;
 const DEFAULT_TIMEOUT = 25000 as const;
@@ -25,7 +24,7 @@ const providerFnMap: Record<
   starknet: getStarknetProvider
 };
 
-function normalizeOptions(
+export function normalizeOptions(
   options: ProviderOptions = {}
 ): Required<ProviderOptions> {
   return {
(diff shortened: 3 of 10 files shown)


test/integration/utils/providers.spec.ts
```diff
@@ -75,6 +75,36 @@ describe('test providers', () => {
...
+      const client2 = getViemClient('1', { timeout: 30000 });
+      const client3 = getViemClient('1', { broviderUrl: 'https://custom.rpc' });
+
+      expect(client1).not.toBe(client2);
+      expect(client1).not.toBe(client3);
+      expect(client2).not.toBe(client3);
+    });
+
+    test('should create separate instances for different networks', () => {
+      expect(getViemClient('1')).not.toBe(getViemClient('56'));
```
tony8713, line 120: The memoization itself isn't covered: delete the `viemClientMemo.get()` early return in `getViemClient` and all 22 tests in this file still pass. The `getProvider` suite has the positive case (`should memoize providers with same network and options`), so one `expect(getViemClient('1')).toBe(getViemClient('1'))` here would close the gap.

src/utils/provider.ts
```diff
@@ -73,6 +74,31 @@ export default function getProvider(
   return provider;
 }
 
+const viemClientMemo = new Map<string, PublicClient>();
+
+export function getViemClient(
```
tony8713: This doesn't dispatch on `getProviderType` the way `getProvider` does, so `getViemClient('0x534e5f4d41494e')` hands back an EVM client pointed at `https://rpc.snapshot.org/sn` and only fails at call time. Given the chain-less design is deliberately fail-loud, throwing for non-EVM networks here would be more consistent.

rollup.config.js
```diff
@@ -11,7 +11,10 @@ import pkg from './package.json';
 
 const name = 'snapshot';
 const input = 'src/index.ts';
-const external = [...Object.keys(pkg.dependencies || {})];
+const dependencies = [...Object.keys(pkg.dependencies || {})];
+// match subpath imports too (e.g. viem/ens)
+const external = (id) =>
```
tony8713, line 16: Worth keeping, but it doesn't do quite what the description says. The cjs/esm config has no node-resolve plugin, so bare subpaths were already left external: I added a real `viem/ens` import and built it both ways, and the output is byte-identical with `require('viem/ens')` in both. What the change actually buys is silencing the `Unresolved dependencies` warning, which is still fine by me.

package.json
```diff
@@ -23,7 +23,8 @@
     "cross-fetch": "^3.1.6",
     "json-to-graphql-query": "^2.2.4",
     "lodash.set": "^4.3.2",
-    "starknet": "^6.11.0"
+    "starknet": "^6.11.0",
+    "viem": "^2.55.11"
```
tony8713, line 27: One consumer-visible effect even though nothing calls viem yet: the published cjs keeps a bare `require('viem')`, since rollup can't drop a side-effect import of an external, so `require('@snapshot-labs/snapshot.js')` goes from 512 to 954 loaded modules. viem sets `sideEffects: false` so the UI's bundler drops it, but hub and sequencer pay it at boot. Also no version bump in this PR, unlike #1220 and #1218.

tony8713 (approved):
Approving. Ran it: 346/346 green on Node 22, build and lint clean, and I checked that the two build changes actually do something rather than taking the description's word for it.

`inlineDynamicImports` is load-bearing. I made viem genuinely reachable from `src/index.ts` (a `getEnsAddress` path) and then removed the option, and the UMD build dies with `UMD and IIFE output formats are not supported for code-splitting builds`. Worth keeping the comment above it so nobody deletes the line as unused. That same experiment puts the UMD bundle at 540 KB gzipped vs 432 KB today, which is a number for the ENS PR to answer for rather than this one.

Four small things below, none blocking.

PR #1222, opened 2026-08-06
refactor: move ENS resolution into src/utils/ens.ts

ChaituVR:
## What

Moves the ENS resolution code out of `utils.ts` into a dedicated `src/utils/ens.ts` module — **a pure move, zero behavior change**:

- `getEnsTextRecord` and `getEnsOwner` relocate **verbatim** (still ethers v5, still the same resolver-allowlist multicall), together with their private helpers (`getDomainType`, `getDNSOwner`) and constants (`MUTED_ERRORS`, `ENS_REGISTRY`, `ENS_ABI`).
- `utils.ts` imports and re-exports both, so the public API (`snapshot.utils.getEnsTextRecord` / `getEnsOwner`, signatures and defaults) is unchanged.
- `utils.ts` diff is +2/−158 — one import line, one re-export line, the rest deletions.

## Why

Prep for the ENS v2 / Universal Resolver support (#1217 follow-up, on top of #1220/#1221): with ENS resolution in its own module, the upcoming functional change lands as an old-function → new-function diff **inside one file**, instead of interleaving a rewrite through a 900-line utils.ts. Reviewers get: this PR = boring move, next PR = pure behavior.

## How to test

**Move verification (mechanical, not by eye):** every relocated symbol was extracted from both `master:src/utils.ts` and the new module and compared after whitespace normalization — all 9 identical; every symbol remaining in `utils.ts` compared against master — all identical.

**Before/after:**
1. `git diff master --stat` → 2 files only, +168/−158 (the +10 net = module imports + blank lines)
2. `yarn typecheck && yarn lint && yarn build && yarn test:once` → all green, **350/350, zero test modifications needed** — no observable behavior moved
3. Entry-point guard from #1221 still passes: no new imports reach the published bundles

### DIFF
--- a/src/utils/ens.ts
+++ b/src/utils/ens.ts
@@ -1,0 +1,166 @@
+import { Contract } from '@ethersproject/contracts';
+import { getAddress } from '@ethersproject/address';
+import { namehash, ensNormalize } from '@ethersproject/hash';
+import getProvider from './provider';
+import { multicall } from '../multicall';
+import networks from '../networks.json';
+
+type DomainType = 'ens' | 'tld' | 'other-tld' | 'subdomain';
+
+const MUTED_ERRORS = [
+  // mute error from coinbase, when the subdomain is not found
+  // most other resolvers just return an empty address
+  'response not found during CCIP fetch',
+  // mute error from missing offchain resolver (mostly for sepolia)
+  'UNSUPPORTED_OPERATION'
+];
+const ENS_REGISTRY = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
+const ENS_ABI = [
+  'function text(bytes32 node, string calldata key) external view returns (string memory)',
+  'function resolver(bytes32 node) view returns (address)' // ENS registry ABI
+];
+const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+function getDomainType(domain: string): DomainType {
+  const isEns = domain.endsWith('.eth');
+
+  const tokens = domain.split('.');
+
+  if (tokens.length === 1) return 'tld';
+  else if (tokens.length === 2 && !isEns) return 'other-tld';
+  else if (tokens.length > 2) return 'subdomain';
+  else if (isEns) return 'ens';
+  else throw new Error('Invalid domain');
+}
+
+// see https://docs.ens.domains/registry/dns#gasless-import
+async function getDNSOwner(domain: string): Promise<string> {
+  const response = await fetch(
+    `https://cloudflare-dns.com/dns-query?name=${domain}&type=TXT`,
+    {
+      headers: {
+        accept: 'application/dns-json'
+      }
+    }
+  );
+
+  const data = await response.json();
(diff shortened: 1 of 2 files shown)


src/utils/ens.ts
```diff
@@ -0,0 +1,166 @@
...
+  if (tokens.length === 1) return 'tld';
+  else if (tokens.length === 2 && !isEns) return 'other-tld';
+  else if (tokens.length > 2) return 'subdomain';
+  else if (isEns) return 'ens';
+  else throw new Error('Invalid domain');
+}
+
+// see https://docs.ens.domains/registry/dns#gasless-import
+async function getDNSOwner(domain: string): Promise<string> {
+  const response = await fetch(
```
tony8713, line 39: This `fetch` is now the global one; in master it resolved to the wrapper defined further up `utils.ts` (30s AbortController over cross-fetch), so the DNS lookup has silently lost its timeout — against a server that accepts and never responds, master aborts at 30s with `Request timeout after 30000ms` where this branch is still pending at 35s. `import { fetch } from '../utils';` restores it: build and tests stay green, and the only cost is a `utils -> ens -> utils` cycle of the kind the repo already has with blockfinder.

tony8713 (changes requested):
Not refactor-only, but it's close: one thing changed. `getDNSOwner` used to call the `fetch` wrapper defined in `utils.ts` and now calls the global `fetch`, so the Cloudflare DNS lookup has lost its 30s timeout. One-line fix, comment inline.

Everything else holds up. I built both sides and diffed the bundles — the only differences are module ordering, a duplicate `EMPTY_ADDRESS` (same literal), and rollup renaming the utils `fetch` to `fetch$1`, which is what gave the binding change away. All four moved function bodies are character-identical to master, the public API is unchanged down to export key order and the `.d.ts` signatures, and both sides run the same 350 tests green with nothing added, dropped or renamed.

Worth saying that CI can't see this one. I mutated the moved code six ways and the tests caught five, so the coverage is real, but nothing in the suite distinguishes the two `fetch`es. No interaction with #1219 either — different files, and `getEnsTextRecord` goes through the EVM multicall path.

src/utils/ens.ts
tony8713: Fixed, and I checked it in the built bundle rather than the diff, since a text diff can't see a binding change. Before, rollup renamed the utils wrapper to `fetch$1` and left `getDNSOwner` on the global; now there is a single `function fetch` and that is the wrapper. Behaviour matches master too: against a server that accepts and never responds the lookup aborts at 30s with `Request timeout after 30000ms`, where the pre-fix commit was still pending at 45s.

tony8713 (approved):
LGTM. The `fetch` binding is back on the timeout wrapper, confirmed in the bundle and by behaviour rather than from the diff, and everything else still holds: the moved bodies are character-identical to master, the public API and `.d.ts` signatures are unchanged, and 387 tests pass identically on both sides.

The import does add a `utils -> ens -> utils` cycle, which rollup now warns about next to the existing blockfinder one. UMD is unaffected, still a single file with no chunks and no surviving dynamic `import()`, and it loads and runs, so `inlineDynamicImports` is still doing its job.

PR #1225, opened 2026-08-07
fix(provider): honour the timeout option on Starknet providers

### LINKED ISSUE #1224: getStarknetProvider drops the timeout option
`getProvider(network, { timeout })` honours the timeout on EVM networks but not on Starknet ones. `getStarknetProvider` forwards only `nodeUrl`:

```ts
// src/utils/provider.ts
function getStarknetProvider(networkKey, options) {
  return new RpcProvider({
    nodeUrl: `${options.broviderUrl}/${networkKey}`
  });
}
```

`getEvmProvider` right above it passes `timeout: options.timeout`, and `getViemClient` does the same, so the option is silently dropped only on the Starknet path. starknet.js v6 has no timeout of its own, so nothing downstream bounds the request either, and the 25s default is not applied either.

The reason this is worse than a slow request: a hang is not contained by a caller's `try/catch`. In stamp, address resolution awaits every resolver under one `Promise.all` (`src/addressResolvers/index.ts`), so one unresponsive Starknet RPC stalls the whole `lookup_addresses` request, inputs with no Starknet component included. A throw degrades one resolver, a hang takes the request. stamp's Starknet resolver used to go through axios with a real 5s socket timeout and lost it when it moved to this provider (snapshot-labs/stamp#491, which carries a local `Promise.race` workaround meant to be dropped once this ships).

### Repro

Point it at a socket that accepts the connection and never answers, which is the hung-proxy failure mode rather than connection-refused:

```js
const server = net.createServer(() => {}); // accepts, never writes back
const provider = new RpcProvider({ nodeUrl }); // what getStarknetProvider builds
await provider.getSpecVersion();
```

Still pending at 8s. Left alone it runs to undici's ~300s ceiling.

### Suggested fix

`RpcProviderOptions` has no `timeout` field, but it does accept `baseFetch`, and starknet.js calls it with an init object carrying no signal of its own, so a wrapper is enough:

```ts
function getStarknetProvider(networkKey, options) {
  return new RpcProvider({
    nodeUrl: `${options.broviderUrl}/${networkKey}`,
    baseFetch: (url, init) =>
      fetch(url, { ...init, signal: AbortSignal.timeout(options.timeout) })
  });
}
```

Same repro against that provider rejects at 2005ms for a 2000ms timeout, verified on starknet 6.24.1 / node 22.

Two small notes. The abort surfaces as a plain `Error: The operation was aborted due to timeout`, so it may be worth normalising it to the `Request timeout after Xms` that the `fetch` helper in `src/utils.ts` throws. That helper cannot simply be reused here though, `src/utils.ts` imports `./utils/provider`, so importing it back would be a cycle. And `createMemoKey` already includes the timeout, so nothing else needs changing.

tony8713:
Fixes #1224.

### What was broken

`getStarknetProvider` forwarded only `nodeUrl`, so a caller-supplied `timeout` and the default were both dropped. The EVM provider and the viem client pass it through, so this was Starknet only.

Nothing downstream picked up the slack: starknet.js v6 has no timeout option and calls its transport with no signal of its own, so a node that accepts the connection and never answers hangs rather than throws. And a hang is not contained by a caller's `try/catch` the way a throw is: stamp's avatar endpoint fans its resolvers out under one `Promise.all` and the Starknet one uses this provider, so a single unresponsive RPC stalls the whole request.

### The fix

`RpcProviderOptions` has no `timeout`, but it does accept `baseFetch`, so the deadline goes on the fetch: a wrapper around native `fetch` that aborts the request once the timeout elapses. What it has to get right:

- The deadline is held until the response body is read, not until the fetch settles. `fetch` resolves as soon as the headers are in, and nothing bounds the read that follows.
- A caller `signal` is composed with the timeout rather than replaced, so a genuine caller abort still surfaces as itself.
- Bad input can neither disable the bound nor collapse it to the next tick: `normalizeOptions` sends a non-finite or negative `timeout` back to the default and caps it at the largest delay `setTimeout` accepts.

The wrapper is generic and lives in its own leaf module, `src/utils/fetch.ts`, with its tests alongside it. Leaf means it imports nothing, so it adds no cycle. Starknet's constraint stays at the call site instead of in the primitive: `RpcChannel` re-throws anything that is not a `LibraryError` as a bare `Error` and drops the `code`, so `getStarknetProvider` hands in an error factory producing a `LibraryError` with `code: 'TIMEOUT'` — the code ethers puts on the EVM provider's timeout, so both providers now fail the same way and stamp's `isSilencedError` recognises both.

The `starknet` floor moves up to the 6.x where `RpcProviderOptions.baseFetch` first exists; below that the wrapper is handed to a constructor that never calls it, and the emitted `.d.ts` fails `tsc` for consumers. The resolved version is unchanged.

`utils.fetch` keeps its own implementation rather than adopting the primitive: that would change two behaviours of a published function, which is semver-visible and wants its own PR.

### DIFF
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -155,6 +155,10 @@ function getStarknetProvider(
 ): RpcProvider {
   return new RpcProvider({
     nodeUrl: `${options.broviderUrl}/${networkKey}`,
-    baseFetch: withTimeout(crossFetch, options.timeout)
+    // `undefined` leaves starknet its own transport (`baseFetch ?? ponyfill`):
+    // with no timeout to add, swapping in crossFetch would only cost the
+    // caller a transport and the ceiling that comes with the stock one.
+    baseFetch:
+      options.timeout > 0 ? withTimeout(crossFetch, options.timeout) : undefined
   });
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -101,7 +101,9 @@ type BaseFetch = NonNullable<RpcProviderOptions['baseFetch']>;
 // Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
 // Safari 16.
 export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
-  // Without this, `setTimeout(..., 0)` aborts every request on the next tick.
+  // getStarknetProvider never calls in at 0, so this is the export's own
+  // precondition rather than a live path: without it, `setTimeout(..., 0)`
+  // would abort every request on the next tick.
   if (timeout <= 0) return baseFetch;
 
   return async (url, init) => {
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -1,5 +1,6 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { LibraryError, RpcProvider, RpcProviderOptions } from 'starknet';
+import { LibraryError, RpcProvider } from 'starknet';
+import { withTimeout } from './fetch';
 import networks from '../networks.json';
 
 export interface ProviderOptions {
@@ -96,71 +97,13 @@ function getEvmProvider(
   );
 }
 
-type BaseFetch = NonNullable<RpcProviderOptions['baseFetch']>;
-
-// Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
-// Safari 16.
-export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
-  // getStarknetProvider never calls in at 0, so this is the export's own
-  // precondition rather than a live path: without it, `setTimeout(..., 0)`
-  // would abort every request on the next tick.
-  if (timeout <= 0) return baseFetch;
-
-  return async (url, init) => {
-    const controller = new AbortController();
-    const callerSignal = init?.signal;
-    const abortFromCaller = () => controller.abort(callerSignal?.reason);
-    let timedOut = false;
-
-    const cleanup = () => {
-      clearTimeout(timer);
-      callerSignal?.removeEventListener('abort', abortFromCaller);
-    };
-    // Cleans up after itself: nothing else will if the body is never read.
-    const timer = setTimeout(() => {
-      timedOut = true;
-      controller.abort();
-      cleanup();
-    }, timeout);
-    // Node-only: browser timers are numbers with no `unref`. Without this,
-    // an unread response body holds the event loop open for the full
-    // timeout even though the request already finished.
-    if (typeof timer === 'object' && 'unref' in timer) timer.unref();
-
-    // LibraryError specifically: RpcChannel#errorHandler re-throws anything
-    // else as a bare `Error(message)`, which would drop `code`.
-    const timeoutError = () =>
-      Object.assign(new LibraryError(`Request timeout after ${timeout}ms`), {
-        code: 'TIMEOUT'
-      });
-
-    if (callerSignal?.aborted) abortFromCaller();
-    else callerSignal?.addEventListener('abort', abortFromCaller);
-
-    let response: Response;
-    try {
-      response = await baseFetch(url, { ...init, signal: controller.signal });
-    } catch (e) {
-      cleanup();
-      throw timedOut ? timeoutError() : e;
-    }
-
-    // Resolving only means the headers arrived; nothing bounds the body stream
-    // or the parse that follow. Hold the deadline over RpcChannel's read.
-    const json = response.json.bind(response);
-    response.json = async () => {
-      try {
-        return await json();
-      } catch (e) {
-        throw timedOut ? timeoutError() : e;
-      } finally {
-        cleanup();
-      }
-    };
-
-    return response;
-  };
-}
+// LibraryError specifically: RpcChannel#errorHandler re-throws anything else
+// as a bare `Error(message)`, which would drop `code`. Nothing outside this
+// call site needs it, so it stays out of the primitive.
+const starknetTimeoutError = (timeout: number) =>
+  Object.assign(new LibraryError(`Request timeout after ${timeout}ms`), {
+    code: 'TIMEOUT'
+  });
 
 function getStarknetProvider(
   networkKey: string,
@@ -176,7 +119,9 @@ function getStarknetProvider(
     // connection reuse), an XHR ponyfill in the bundled browser build.
     baseFetch:
       options.timeout > 0
-        ? withTimeout(fetch.bind(globalThis), options.timeout)
+        ? withTimeout(fetch.bind(globalThis), options.timeout, {
+            timeoutError: starknetTimeoutError
+          })
         : undefined
   });
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -123,6 +123,10 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
       controller.abort();
       cleanup();
     }, timeout);
+    // Node-only: browser timers are numbers with no `unref`. Without this,
+    // an unread response body holds the event loop open for the full
+    // timeout even though the request already finished.
+    if (typeof timer === 'object' && 'unref' in timer) timer.unref();
 
     // LibraryError specifically: RpcChannel#errorHandler re-throws anything
     // else as a bare `Error(message)`, which would drop `code`.
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -13,7 +13,7 @@ type ProviderInstance = StaticJsonRpcProvider | RpcProvider;
 export type ProviderType = 'evm' | 'starknet';
 
 const DEFAULT_BROVIDER_URL = 'https://rpc.snapshot.org' as const;
-const DEFAULT_TIMEOUT = 25000 as const;
+export const DEFAULT_TIMEOUT = 25000 as const;
 
 const providerMemo = new Map<string, ProviderInstance>();
 
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -1,5 +1,6 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { RpcProvider } from 'starknet';
+import crossFetch from 'cross-fetch';
+import { LibraryError, RpcProvider, RpcProviderOptions } from 'starknet';
 import networks from '../networks.json';
 
 export interface ProviderOptions {
@@ -87,11 +88,56 @@ function getEvmProvider(
   );
 }
 
+type BaseFetch = NonNullable<RpcProviderOptions['baseFetch']>;
+
+// starknet.js v6 has no timeout option, and its RpcChannel calls `baseFetch`
+// with no signal of its own, so bounding the fetch is the only way to bound a
+// request. Without this a node that accepts the connection but never answers
+// hangs until the runtime's own ceiling (~300s on undici), and a hang is not
+// contained by a caller's try/catch the way a throw is.
+//
+// AbortController + setTimeout rather than `AbortSignal.timeout`, which needs
+// Safari 16 / Firefox 100: this is what the `fetch` helper in ../utils already
+// does, so the browser bundle keeps the floor it has today. That helper cannot
+// be reused here, ../utils imports this module.
+export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
+  if (timeout <= 0) return baseFetch;
+
+  return async (url, init) => {
+    const controller = new AbortController();
+    let timedOut = false;
+    const timer = setTimeout(() => {
+      timedOut = true;
+      controller.abort();
+    }, timeout);
+
+    try {
+      return await baseFetch(url, { ...init, signal: controller.signal });
+    } catch (e) {
+      if (!timedOut) throw e;
+
+      // RpcChannel#errorHandler re-throws anything that is not a LibraryError
+      // as a bare `Error(message)`, which drops own properties, so `code` only
+      // survives on a LibraryError. `TIMEOUT` is the code ethers puts on the
+      // EVM provider's timeout, so both providers now fail the same way.
+      throw Object.assign(
+        new LibraryError(`Request timeout after ${timeout}ms`),
+        {
+          code: 'TIMEOUT'
+        }
+      );
+    } finally {
+      clearTimeout(timer);
+    }
+  };
+}
+
 function getStarknetProvider(
   networkKey: string,
   options: Required<ProviderOptions>
 ): RpcProvider {
   return new RpcProvider({
-    nodeUrl: `${options.broviderUrl}/${networkKey}`
+    nodeUrl: `${options.broviderUrl}/${networkKey}`,
+    baseFetch: withTimeout(crossFetch, options.timeout)
   });
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -36,7 +36,9 @@ export function normalizeOptions(
     // `setTimeout(..., NaN)` fires on the next tick, aborting every request.
     timeout:
       typeof timeout === 'number' && Number.isFinite(timeout) && timeout >= 0
-        ? timeout
+        ? // setTimeout's delay is a 32-bit signed int; above this Node
+          // silently clamps it to ~1ms instead of the requested duration.
+          Math.min(timeout, 2_147_483_647)
         : DEFAULT_TIMEOUT
   };
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -1,5 +1,4 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import crossFetch from 'cross-fetch';
 import { LibraryError, RpcProvider, RpcProviderOptions } from 'starknet';
 import networks from '../networks.json';
 
@@ -146,8 +145,8 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
       throw timedOut ? timeoutError() : e;
     }
 
-    // Resolving only means the headers arrived, and node-fetch bounds neither
-    // the body stream nor the parse. Hold the deadline over RpcChannel's read.
+    // Resolving only means the headers arrived; nothing bounds the body stream
+    // or the parse that follow. Hold the deadline over RpcChannel's read.
     const json = response.json.bind(response);
     response.json = async () => {
       try {
@@ -170,8 +169,14 @@ function getStarknetProvider(
   return new RpcProvider({
     nodeUrl: `${options.broviderUrl}/${networkKey}`,
     // At 0 there is no deadline to add, so leave starknet its own transport
-    // (`baseFetch ?? ponyfill`) rather than swap in crossFetch for nothing.
+    // (`baseFetch ?? ponyfill`) rather than swap the transport for nothing.
+    // Native `fetch`, bound like starknet's own browser default: the package
+    // floor has it everywhere (Node >= 18, evergreen browsers), and
+    // cross-fetch would downgrade it — node-fetch 2 on Node (losing
+    // connection reuse), an XHR ponyfill in the bundled browser build.
     baseFetch:
-      options.timeout > 0 ? withTimeout(crossFetch, options.timeout) : undefined
+      options.timeout > 0
+        ? withTimeout(fetch.bind(globalThis), options.timeout)
+        : undefined
   });
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -111,6 +111,15 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
       controller.abort();
     }, timeout);
 
+    // Compose with a caller's signal rather than replacing it, so an abort from
+    // above still cancels the request at once and arrives with its own reason
+    // instead of being held until the timeout. An engine that predates
+    // `abort(reason)` ignores the argument and still aborts.
+    const callerSignal = init?.signal;
+    const abortFromCaller = () => controller.abort(callerSignal?.reason);
+    if (callerSignal?.aborted) abortFromCaller();
+    else callerSignal?.addEventListener('abort', abortFromCaller);
+
     try {
       return await baseFetch(url, { ...init, signal: controller.signal });
     } catch (e) {
@@ -128,6 +137,7 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
       );
     } finally {
       clearTimeout(timer);
+      callerSignal?.removeEventListener('abort', abortFromCaller);
     }
   };
 }
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -28,9 +28,17 @@ const providerFnMap: Record<
 export function normalizeOptions(
   options: ProviderOptions = {}
 ): Required<ProviderOptions> {
+  const { timeout } = options;
+
   return {
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
-    timeout: options.timeout ?? DEFAULT_TIMEOUT
+    // Not `??`: it passes `NaN` through (`Number()` of an unset env var), and
+    // `setTimeout(..., NaN)` fires on the next tick, aborting every request.
+    // Here rather than in `withTimeout`: ethers needs the same normalization.
+    timeout:
+      typeof timeout === 'number' && Number.isFinite(timeout) && timeout >= 0
+        ? timeout
+        : DEFAULT_TIMEOUT
   };
 }
 
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -34,7 +34,6 @@ export function normalizeOptions(
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
     // Not `??`: it passes `NaN` through (`Number()` of an unset env var), and
     // `setTimeout(..., NaN)` fires on the next tick, aborting every request.
-    // Here rather than in `withTimeout`: ethers needs the same normalization.
     timeout:
       typeof timeout === 'number' && Number.isFinite(timeout) && timeout >= 0
         ? timeout
@@ -141,9 +140,8 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
       throw timedOut ? timeoutError() : e;
     }
 
-    // Resolving only means the headers arrived: RpcChannel reads the body
-    // afterwards with its own `.json()`, and node-fetch bounds neither the
-    // body stream nor the parse. Hold the deadline until that settles.
+    // Resolving only means the headers arrived, and node-fetch bounds neither
+    // the body stream nor the parse. Hold the deadline over RpcChannel's read.
     const json = response.json.bind(response);
     response.json = async () => {
       try {
@@ -165,9 +163,8 @@ function getStarknetProvider(
 ): RpcProvider {
   return new RpcProvider({
     nodeUrl: `${options.broviderUrl}/${networkKey}`,
-    // `undefined` leaves starknet its own transport (`baseFetch ?? ponyfill`):
-    // with no timeout to add, swapping in crossFetch would only cost the
-    // caller a transport and the ceiling that comes with the stock one.
+    // At 0 there is no deadline to add, so leave starknet its own transport
+    // (`baseFetch ?? ponyfill`) rather than swap in crossFetch for nothing.
     baseFetch:
       options.timeout > 0 ? withTimeout(crossFetch, options.timeout) : undefined
   });
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -98,34 +98,54 @@ export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
 
   return async (url, init) => {
     const controller = new AbortController();
+    const callerSignal = init?.signal;
+    const abortFromCaller = () => controller.abort(callerSignal?.reason);
     let timedOut = false;
+
+    const cleanup = () => {
+      clearTimeout(timer);
+      callerSignal?.removeEventListener('abort', abortFromCaller);
+    };
+    // Cleans up after itself: nothing else will if the body is never read.
     const timer = setTimeout(() => {
       timedOut = true;
       controller.abort();
+      cleanup();
     }, timeout);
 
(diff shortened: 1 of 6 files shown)


src/utils/provider.ts
```diff
@@ -87,11 +88,56 @@ function getEvmProvider(
...
+      // EVM provider's timeout, so both providers now fail the same way.
+      throw Object.assign(
+        new LibraryError(`Request timeout after ${timeout}ms`),
+        {
+          code: 'TIMEOUT'
+        }
+      );
+    } finally {
+      clearTimeout(timer);
+    }
```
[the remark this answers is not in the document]
tony8713: Fixed in a6f750b: a caller abort now aborts the timeout controller with the caller's own reason, a signal that is already aborted is caught before the fetch, and the listener is dropped when the request settles. One correction, at `timeout: 0` the helper returns the wrapped fetch untouched, so `init.signal` was already honoured on that path. Four new tests, all four failing with the old `signal: controller.signal` line back in place.

tony8713:
Fixed in a6f750b. A caller abort now aborts the timeout controller with the caller's own reason, a signal that is already aborted is caught before the fetch, and the listener is removed when the request settles. Four tests cover it, and all four fail if I put the old `signal: controller.signal` line back: a caller abort not rewritten to `TIMEOUT`, an already-aborted signal, listener cleanup, and one against the hung TCP server through the real `cross-fetch` to prove the abort reaches the socket rather than waiting out the timeout.

On severity I'd push back slightly: nothing in this repo can hand `withTimeout` a signal today. `RpcChannel` calls `baseFetch(nodeUrl, { method, body, headers })` with no signal of its own (starknet 6.x, `dist/index.js:4250`), and `baseFetch` is deliberately kept out of `ProviderOptions` because `createMemoKey` interpolates options and two distinct closures with the same body would collide. So the helper's contract was wrong rather than any production path, which is still worth fixing but is not a caller abort being ignored anywhere today.

tony8713:
Cut 28 lines of comment for 7 in 90655b9, no code touched: what went was restating the line under it or repeating the description. Kept four whys, that the thrown error has to be a `LibraryError` (`RpcChannel#errorHandler` re-throws anything else as a bare `Error` and drops the `code`, which silently breaks stamp's `isSilencedError`), `AbortController` rather than `AbortSignal.timeout` because of the browser bundle's floor, the `timeout <= 0` guard, and why the test server accepts a connection and never answers. The guard one is a new line rather than a survivor, it was only in the description before. Trimmed the description too.

src/utils/provider.ts
```diff
@@ -87,11 +88,53 @@ function getEvmProvider(
...
+      throw Object.assign(
+        new LibraryError(`Request timeout after ${timeout}ms`),
+        {
+          code: 'TIMEOUT'
+        }
+      );
+    } finally {
+      clearTimeout(timer);
+      callerSignal?.removeEventListener('abort', abortFromCaller);
+    }
```
wa0x6e: **The timeout only bounds the headers phase — a node that sends headers and then stalls the body still hangs, now without even the old 300s ceiling.**

On Node, `crossFetch` is `node-fetch@2.6.11`, whose `fetch()` promise resolves at the HTTP `response` event (headers received) with an unconsumed body stream. The `finally` here then clears the timer and abort wiring — but `RpcChannel#fetchEndpoint` reads the body afterwards with its own `await rawResult.json()`, which node-fetch does not bound (its body timeout only arms via its own `timeout` option, unset here). Reproduced against a server that sends `HTTP/1.1 200 OK` + `Content-Length: 2` and never writes the body: headers resolved at 57ms, `res.json()` still pending 3s+ after a 300ms timeout.

Aggravating: starknet's default Node fetch (`fetchCookie(global.fetch)` = undici) had a ~300s `bodyTimeout` ceiling; node-fetch v2 has none. So for the headers-then-stall failure mode this PR replaces a 300s ceiling with an unbounded hang — a hung proxy (the #1224 incident class) can wedge either before or after relaying upstream headers, so this path is as reachable as the one the PR fixes.

Smallest sound fix: keep the deadline armed until `.json()` settles (`RpcChannel`/`BatchClient` are the only `baseFetch` callers in starknet 6.24.1 and both only ever call `.json()` — verified by grep over the bundled dist). Verified against the same repro: rejects at 324ms with `code: 'TIMEOUT'`.

```suggestion
    const cleanup = () => {
      clearTimeout(timer);
      callerSignal?.removeEventListener('abort', abortFromCaller);
    };

    // LibraryError specifically: RpcChannel#errorHandler re-throws anything
    // else as a bare `Error(message)`, which would drop `code`.
    const timeoutError = () =>
      Object.assign(new LibraryError(`Request timeout after ${timeout}ms`), {
        code: 'TIMEOUT'
      });

    let response: Response;
    try {
      response = await baseFetch(url, { ...init, signal: controller.signal });
    } catch (e) {
      cleanup();
      throw timedOut ? timeoutError() : e;
    }

    // `baseFetch` resolving only means headers arrived: RpcChannel reads the
    // body afterwards with its own `r.json()`, so keep the deadline armed
    // until `.json()` settles.
    const originalJson = response.json.bind(response);
    response.json = async () => {
      try {
        return await originalJson();
      } catch (e) {
        throw timedOut ? timeoutError() : e;
      } finally {
        cleanup();
      }
    };

    return response;
```

Note: this changes *when* cleanup fires, so two existing `withTimeout()` tests that never call `.json()` (`'clears its timer so a resolved request holds nothing open'`, `'stops listening on the caller signal once the request settles'`) need an `await response.json()` before their assertions. A regression test whose server sends headers + partial body then stalls would pin this down.

src/utils/provider.ts
```diff
@@ -87,11 +88,53 @@ function getEvmProvider(
   );
 }
 
+type BaseFetch = NonNullable<RpcProviderOptions['baseFetch']>;
+
+// Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
+// Safari 16.
+export function withTimeout(baseFetch: BaseFetch, timeout: number): BaseFetch {
+  // Without this, `setTimeout(..., 0)` aborts every request on the next tick.
+  if (timeout <= 0) return baseFetch;
```
wa0x6e: **A `NaN` (or `Infinity`) timeout slips past this guard and aborts every request almost instantly.**

`NaN <= 0` is `false`, so `timeout: NaN` (the classic `Number(unsetEnvVar)` footgun) builds the wrapper and arms `setTimeout(cb, NaN)`, which Node fires in single-digit milliseconds — every Starknet request then rejects with `Request timeout after NaNms` before a real response can arrive (`Infinity` behaves the same: Node clamps delays > 2³¹−1 to 1ms). `normalizeOptions` doesn't catch it either: `??` only substitutes `null`/`undefined`. Meanwhile the EVM path silently falls back to ethers' own hardcoded 120s default for the same input (`fetchJson` only overrides its internal default when `typeof timeout === "number" && timeout > 0`), so the two providers diverge on identical options.

Suggest normalizing once in `normalizeOptions` — the single dispatch point for both paths — rather than widening this guard (which would make bad input *disable* the Starknet timeout, regressing to the unbounded hang this PR fixes). `timeout: 0` (explicit disable) stays intact:

```diff
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -28,9 +28,18 @@ const providerFnMap: Record<
 export function normalizeOptions(
   options: ProviderOptions = {}
 ): Required<ProviderOptions> {
+  const { timeout } = options;
+  // A non-finite or negative timeout (e.g. `Number()` of an unset env var,
+  // yielding NaN) must not reach either provider path: it passes Starknet's
+  // `timeout <= 0` guard in withTimeout and fires `setTimeout(cb, NaN)`
+  // almost immediately, while ethers silently falls back to its own 2-minute
+  // default on the EVM path. Normalising once here keeps both paths in sync.
   return {
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
-    timeout: options.timeout ?? DEFAULT_TIMEOUT
+    timeout:
+      typeof timeout === 'number' && Number.isFinite(timeout) && timeout >= 0
+        ? timeout
+        : DEFAULT_TIMEOUT
   };
 }
```

Typechecks clean under this repo's `tsc --strict`; verified `undefined→25000`, `NaN→25000`, `-100→25000`, `0→0`, `Infinity→25000`, `30000→30000`.

test/integration/utils/providers.spec.ts
```diff
@@ -1,8 +1,13 @@
...
-import getProvider from '../../../src/utils/provider';
+import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import { AddressInfo, createServer, Server, Socket } from 'net';
+import crossFetch from 'cross-fetch';
+import getProvider, { withTimeout } from '../../../src/utils/provider';
 import { getViemClient } from '../../../src/utils/viem';
 import { RpcProvider } from 'starknet';
 
+const STARKNET_NETWORK = '0x534e5f4d41494e';
+const DEFAULT_TIMEOUT = 25000;
```
wa0x6e: **`DEFAULT_TIMEOUT = 25000` here is a hand-copied literal of the un-exported constant in `src/utils/provider.ts:16` — the two can drift silently.**

If the source default later changes (say 25000 → 30000) without this copy, the `applies the default timeout when none is passed` test advances fake timers by the stale 25000ms, the abort never fires, and the test dies at Vitest's default 5s timeout with "Test timed out in 5000ms" — pointing nowhere near the real cause.

Suggest importing it instead (and adding `export` to `const DEFAULT_TIMEOUT` in `src/utils/provider.ts:16` — outside this diff, so not suggestible here; safe since that module already exports `withTimeout` purely for this test file, and none of its helper exports reach the published API):

```suggestion
import getProvider, {
  withTimeout,
  DEFAULT_TIMEOUT
} from '../../../src/utils/provider';
import { getViemClient } from '../../../src/utils/viem';
import { RpcProvider } from 'starknet';

const STARKNET_NETWORK = '0x534e5f4d41494e';
```

(Import reflowed to four lines: the single-line form is 88 chars, over the repo's Prettier 80.)

src/utils/provider.ts
```diff
@@ -87,11 +88,53 @@ function getEvmProvider(
...
+}
+
 function getStarknetProvider(
   networkKey: string,
   options: Required<ProviderOptions>
 ): RpcProvider {
   return new RpcProvider({
-    nodeUrl: `${options.broviderUrl}/${networkKey}`
+    nodeUrl: `${options.broviderUrl}/${networkKey}`,
+    baseFetch: withTimeout(crossFetch, options.timeout)
```
wa0x6e: **Question on the `timeout: 0` semantics: should it really mean "unbounded"?**

Two observations about the current state:

1. At `timeout: 0` this line still installs `crossFetch`, so the caller pays the transport swap (in the browser bundle that's cross-fetch's XHR shim — its ponyfill discards native `fetch` even when present, confirmed shipping in `dist/snapshot.min.js`) and gets no timeout in exchange. Starknet's own default (`fetchCookie(global.fetch)` on Node) also had undici's ~300s ceiling behind it; bare `crossFetch`/node-fetch has none, so "disabled" now means genuinely infinite rather than 300s.
2. The EVM path can't actually be disabled: ethers' `fetchJson` only honors `timeout > 0` and otherwise keeps its internal 120s default. So `timeout: 0` today means "hang forever" on Starknet but "120s" on EVM — the same option value, two very different meanings.

If 0-means-disable isn't a case anyone needs (no in-repo caller passes 0), one guard in `normalizeOptions` makes both paths uniformly bounded and also swallows the NaN/negative footgun from the earlier comment — amending that suggestion's `>= 0` to `> 0`:

```ts
timeout:
  typeof timeout === 'number' && Number.isFinite(timeout) && timeout > 0
    ? timeout
    : DEFAULT_TIMEOUT
```

Trade-off, stated plainly: this drops the documented 0-disable escape hatch, which `src/utils.ts`'s `fetch` and viem's transport do honor — so Starknet/EVM become consistent with each other at the cost of consistency with the rest of the package. Your call; if 0-disable stays, the smaller fix is gating the install so 0 at least restores starknet's stock transport:

```ts
baseFetch:
  options.timeout > 0 ? withTimeout(crossFetch, options.timeout) : undefined
```

(`RpcProvider` treats explicit `undefined` as omitted — `baseFetch ?? fetchPonyfill_default`.)

wa0x6e (commented):
Two items with no diff line to anchor to, so they live here:

**`starknet` floor `^6.11.0` silently no-ops this whole PR for downstream lockfiles on 6.11–6.20 (package.json:26).** `baseFetch` first appears in starknet.js **6.21.0** — verified against the published tarballs of every 6.x from 6.11.0 to 6.24.1: below 6.21.0 neither the types nor the runtime reference it, so the wrapper is passed and never called, and a hung Starknet RPC hangs exactly as before this fix. This repo's own lockfile resolves to 6.24.1, so CI here can't catch it; it bites consumers whose lockfiles resolve `^6.11.0` to something older (any lockfile frozen before 6.21.0's release, or a hoisted older `starknet` from a sibling dep). Separately, the emitted `dist/src/utils/provider.d.ts` indexes `RpcProviderOptions['baseFetch']`, which fails `tsc` under the default `skipLibCheck: false` against starknet <6.21.0 — reproduced end-to-end with a consumer importing only `getProvider`:

```
node_modules/@snapshot-labs/snapshot.js/dist/src/utils/provider.d.ts(12,49): error TS2339: Property 'baseFetch' does not exist on type 'RpcProviderOptions'.
```

Fix: bump the floor —

```diff
-    "starknet": "^6.11.0",
+    "starknet": "^6.21.0",
```

— and regenerate `yarn.lock` (`yarn install` rewrites the `starknet@^6.11.0:` key; resolved version should stay 6.24.1, worth confirming).

**Follow-up PR suggestion (not blocking this one): extract the timeout-fetch primitive.** This is now the second implementation of AbortController + `setTimeout` + finally-clear + `timeout > 0` gate in the package — `src/utils.ts:299-328` has the same mechanism and throws the identical `Request timeout after ${timeout}ms` string. The import-cycle justification in the PR body is accurate (`src/utils.ts` imports `./utils/provider`), but a leaf `src/utils/fetch.ts` both can import — with `src/utils.ts` re-exporting `fetch` to keep the public surface — removes the duplication without touching the published API. Worth doing soon rather than later: the new copy already behaves better than the old one (it doesn't misattribute a caller abort as a timeout, which the `src/utils.ts` helper does), and each divergence makes the eventual unification harder. Deliberately out of scope for this PR — one concern per PR. Same follow-up should settle `withTimeout`'s shape: the name reads transport-agnostic but the body throws starknet's `LibraryError` (forced by `RpcChannel#errorHandler` — anything else loses `code`), so either the extracted primitive stays generic with the `LibraryError` mapping lifted to the `getStarknetProvider` call site, or the helper gets a name that says it's Starknet-specific.

src/utils/provider.ts
tony8713: Confirmed and fixed in `6fcf374`. Reproduced your case as described (headers at 11ms, `.json()` still pending 3s after a 300ms timeout), and the deadline now clears when `.json()` settles instead of when the fetch does.

Two changes from your suggestion: the timer unhooks the caller listener itself, so a response whose body is never read cannot leave one attached, and the shared `Response` in the unit tests had to become a factory since a body only reads once. The regression test is a server that sends complete headers and stalls; it and the body-read unit test both fail with the old cleanup-at-headers line back in place.

src/utils/provider.ts
tony8713: Fixed in `f8a9298`, normalized in `normalizeOptions` exactly as you suggested rather than by widening the guard.

Verified both halves: `setTimeout(cb, NaN)` fires in 2ms here (`Infinity` clamps to 1ms with a `TimeoutOverflowWarning`), and ethers does keep its 120s default on the same input (`@ethersproject/web/lib/index.js:102`, `typeof (connection.timeout) === "number" && connection.timeout > 0`). Kept your `>= 0` so an explicit 0 still disables, per the other thread.

test/integration/utils/providers.spec.ts
tony8713: Done in `09303dc`: `DEFAULT_TIMEOUT` is exported and imported, in the four-line form you wrote.

src/utils/provider.ts
tony8713: Took the second option in `ba1fa64`, `baseFetch: undefined` when the timeout is 0, and kept 0-disable.

Dropping it would quietly turn a deliberate 0 into 25s for anyone passing it today, and 0 does mean disabled for `utils.fetch` and the viem transport, so I would rather the Starknet path stay consistent with the package than buy consistency with ethers' inability to be switched off at all. Both observations were right, including the ponyfill: `var ctx = __self__` in cross-fetch's `browser-ponyfill.js` with the `global.fetch ? global : __self__` line commented out, so native `fetch` is discarded. Leaving this open in case you read the semantics differently.

tony8713:
Floor bumped to `^6.21.0` in `5aacfc9`, `yarn.lock` regenerated and the resolved version stayed 6.24.1 as you expected.

Your boundary is exact. I pulled every 6.x tarball from 6.11.0 up: `baseFetch` is absent from both `dist/index.js` and `dist/index.d.ts` through 6.20.3 and present from 6.21.0 on. Also reproduced the consumer failure end to end, `dist/src/utils/provider.d.ts(13,49): error TS2339: Property 'baseFetch' does not exist on type 'RpcProviderOptions'` against starknet 6.20.3, clean against 6.21.0 and 6.24.1.

On extracting the primitive: agreed, and agreed it does not belong in this PR. I will do it as its own with the leaf `src/utils/fetch.ts` and `src/utils.ts` re-exporting `fetch`. My read on the shape question is that the primitive stays generic and the `LibraryError` mapping moves to the `getStarknetProvider` call site, since that constraint comes from `RpcChannel#errorHandler` and nothing else, but that is the follow-up PR’s argument to have rather than this one’s.

The other four are answered per thread: body-phase deadline in `6fcf374`, non-finite timeout in `f8a9298`, transport at 0 in `ba1fa64`, test constant in `09303dc`. Full gate green on `09303dc` (build, typecheck, lint, 407 tests across 27 files).

Also refreshed the PR body: it still said the timer cleared in a `finally`, which stopped being true with the body-phase fix. Same defect you caught on stamp#491, so worth not leaving there.

src/utils/provider.ts
tony8713: Not from inside the package, but it's the `options.timeout > 0` gate in `getStarknetProvider` doing that rather than the normalization: normalization deliberately passes an explicit `0` through, and the call site then hands starknet `baseFetch: undefined` instead of calling in.

I made the guard log and ran the suite: it fired once across 407 tests, from the unit test that passes `0` directly. So I've kept it as the exported helper's precondition and said as much in the comment (`2fd1cc5`) rather than delete it, since #1226 turns it into a leaf module anyone can call.

tony8713:
@chaituvr on both:

**1. `withTimeout` vs the other timeouts.** You are right that there is duplication, and it is worse than a second wrapper: `utils.fetch` has the same bug this PR fixes. Its timer is cleared when the fetch promise settles, which on node-fetch means when the headers arrive, so the body read is unbounded. Against a server that sends headers and then stalls, `utils.fetch(url, { timeout: 400 })` followed by `res.text()` was still pending at 2071ms, and `getScores`, `getVp`, `validate` and `subgraphRequest` all read the body after that point. It also drops a caller's `signal` outright, overridden in the bounded path and deleted in the disabled one: caller aborted at 200ms, request still pending at 1507ms.

The rest of the package, for the record. ethers (`getEvmProvider`) races its timer against the body read *and* the JSON parse and throws `code: 'TIMEOUT'`, though `timeout: 0` does not disable it, it falls back to ethers' own 2 minutes. viem (`getViemClient`) bounds only the headers, `readResponseBody` runs outside the deadline. So `withTimeout` matches ethers here, and `utils.fetch` and viem are the two with the gap.

Reusing `utils.fetch` for Starknet would therefore have shipped a fix that does not fix the reported hang: no body deadline, and a plain `Error` that `RpcChannel#errorHandler` re-throws as a bare `Error(message)`, dropping the `code` that stamp's `isSilencedError` matches on. (`src/utils.ts` imports `./utils/provider`, so the import direction is a cycle too.)

But "reuse it as-is" is not the only option, and I no longer think two wrappers is the right end state. I prototyped one primitive serving both, parameterised only by the error factory, with the timer `unref`'d and every body accessor wrapped rather than just `.json()`:

```
starknet, stalled body:       LibraryError code=TIMEOUT at 410ms
utils.fetch, stalled body:    Error "Request timeout after 400ms" at 432ms   (today: hangs)
utils.fetch, caller abort:    AbortError at 232ms                            (today: ignored)
utils.fetch, body never read: process exits at 0.41s with a live 30s deadline
```

That last line retracts what I argued on #1226, that sharing would leave a 30s timer holding the event loop behind every `getScores` and `subgraphRequest`. `unref()` removes it, and a late abort on a response that was already read is a no-op.

I would still rather not fold it into this PR. It changes a published `utils.fetch` in three ways: a caller `signal` starts being honoured, so `AbortError` newly escapes a helper documented to throw `Request timeout after Nms`; the body read becomes bounded, so `getScores`/`getVp`/`validate` can reject where they used to hang; and the returned Response's body methods are no longer the native ones. Each is an improvement and each wants its own review. #1226 already moves the primitive into `src/utils/fetch.ts`, which is exactly where it should absorb `utils.fetch`. Say the word and I will open that on top of it.

**2. Comments.** Done in `5180298`, a second pass on top of `90655b9`. Cut the `normalizeOptions` placement line outright, and trimmed four others down to the part the code does not already show, 13 lines out for 7. Kept five, each because a reader gets something wrong without it: the `AbortSignal.timeout` and `LibraryError` notes record external constraints invisible at the call site, the timer's self-cleanup reads as a mistake without it (`clearTimeout` inside the timer's own callback), the `NaN` note is the one thing stopping this going back to `?? DEFAULT_TIMEOUT`, and the `timeout <= 0` precondition answers @wa0x6e's reachability question from earlier today. I had this last one down to cut before that thread landed, so if you read it as noise now it goes.

Full gate green on `5180298`: build, typecheck, lint, 407 tests across 27 files. Circular dependency warnings unchanged at seven. #1226 is a commit behind its base now and will want a rebase.

src/utils/provider.ts
```diff
@@ -27,9 +28,16 @@ const providerFnMap: Record<
...
+  const { timeout } = options;
+
   return {
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
-    timeout: options.timeout ?? DEFAULT_TIMEOUT
+    // Not `??`: it passes `NaN` through (`Number()` of an unset env var), and
+    // `setTimeout(..., NaN)` fires on the next tick, aborting every request.
+    timeout:
+      typeof timeout === 'number' && Number.isFinite(timeout) && timeout >= 0
+        ? timeout
```
wa0x6e: `normalizeOptions` has no upper bound, so a finite `timeout` above 2^31−1 passes both guards and Node clamps the `setTimeout` delay to ~1 ms (`TimeoutOverflowWarning`) — `getProvider(net, { timeout: 3e9 })` aborts every request on the next tick with `Request timeout after 3000000000ms`. Reachable from outside: `verify()` forwards caller-supplied `options.timeout` straight into `getProvider`, so a timestamp-instead-of-duration mixup lands here. Fails loud rather than hanging, so it's minor, but the clamp is one line:

```suggestion
        ? // setTimeout's delay is a 32-bit signed int; above this Node
          // silently clamps it to ~1ms instead of the requested duration.
          Math.min(timeout, 2_147_483_647)
```

Verified locally: `yarn typecheck` and the provider spec pass with the clamp in place.

src/utils/provider.ts
```diff
@@ -87,11 +95,77 @@ function getEvmProvider(
...
+    const cleanup = () => {
+      clearTimeout(timer);
+      callerSignal?.removeEventListener('abort', abortFromCaller);
+    };
+    // Cleans up after itself: nothing else will if the body is never read.
+    const timer = setTimeout(() => {
+      timedOut = true;
+      controller.abort();
+      cleanup();
+    }, timeout);
```
wa0x6e: If `baseFetch` resolves and the body is never read, `cleanup()` never runs, so the ref'd timer holds the Node event loop open for the full timeout — reproduced: a process with a 10 s timeout and an unread body exits at 10.06 s vs 0.07 s with `unref()`. Not reachable through `RpcChannel`/`BatchClient` (both call `.json()` immediately), but `RpcProvider.fetch()` is a public passthrough on what `getProvider()` returns, and `withTimeout` is exported.

```suggestion
    const timer = setTimeout(() => {
      timedOut = true;
      controller.abort();
      cleanup();
    }, timeout);
    // Node-only: browser timers are numbers with no `unref`. Without this,
    // an unread response body holds the event loop open for the full
    // timeout even though the request already finished.
    if (typeof timer === 'object' && 'unref' in timer) timer.unref();
```

(The socket-destruction side is a non-issue — node-fetch 2 sends `Connection: close`, the connection is gone ~30–60 ms after the response, so the late abort lands on nothing.) `yarn typecheck`, `yarn lint`, and the provider spec pass with this applied.

src/utils/provider.ts
tony8713: Applied verbatim in `84e35d6`. Confirmed the reach before taking it: `verify()` hands its `options` straight to `getProvider` in `verify/starknet.ts:27`, and on node 22 a 3e9 delay fires at 6ms with the `TimeoutOverflowWarning`. Covered by two rows in the `normalizeOptions` table, 2^31 and 3e9, both of which fail without the clamp.

src/utils/provider.ts
tony8713: Applied verbatim in `0c7f2b7`. Reproduced your number first against the real `withTimeout` over a local server: a 10s timeout with an unread body exits at 10.01s, and 0.02s with the `unref`. New test asserts `hasRef()` is false after a response whose body is never read, and #1226 carries both into `fetch.ts` on its rebase.

src/utils/provider.ts
```diff
@@ -87,11 +97,81 @@ function getEvmProvider(
...
   networkKey: string,
   options: Required<ProviderOptions>
 ): RpcProvider {
   return new RpcProvider({
-    nodeUrl: `${options.broviderUrl}/${networkKey}`
+    nodeUrl: `${options.broviderUrl}/${networkKey}`,
+    // At 0 there is no deadline to add, so leave starknet its own transport
+    // (`baseFetch ?? ponyfill`) rather than swap in crossFetch for nothing.
+    baseFetch:
+      options.timeout > 0 ? withTimeout(crossFetch, options.timeout) : undefined
```
wa0x6e: **`withTimeout(crossFetch, …)` on the default path silently downgrades the transport for every Starknet request** — this fires whenever `timeout > 0`, i.e. essentially always (default 25000).

What the swap costs, verified against the installed deps:

- **Node**: starknet's own default is `fetch-cookie(global.fetch)` — native undici (`node_modules/starknet/dist/index.js:3728-3732`). `cross-fetch`'s Node entry unconditionally uses `node-fetch@2.6.11` instead, which sends `Connection: close` when no agent is supplied (`node_modules/node-fetch/lib/index.js:1363-1369`) and requests only `gzip,deflate` — so connection reuse to the brovider is lost entirely on every RPC call.
- **Browser**: `cross-fetch`'s ponyfill hardcodes `var ctx = __self__` (the native-fetch branch is commented out, `dist/browser-ponyfill.js:546-547`), so it's always the XHR polyfill even when `window.fetch` exists — and the UMD build (`rollup.config.js`, no `external`, `nodeResolve({ browser: true })`) bundles that into the shipped `dist/snapshot.min.js`.

The wrapper doesn't need cross-fetch at all: `baseFetch` is typed as `WindowOrWorkerGlobalScope['fetch']`, and the package floor (`engines.node >= 18`, evergreen browsers) has native `fetch` everywhere this ships. Bound, like starknet's own browser default (`window.fetch.bind(window)`) — a detached unbound `fetch` threw Illegal invocation in older browsers.

Checked on this head with the change applied: `yarn typecheck`, `yarn lint`, and the 47 tests in `providers.spec.ts` all pass (the `withTimeout(crossFetch, 5000)` test passes its own explicit transport, so it's unaffected).

```suggestion
    // At 0 there is no deadline to add, so leave starknet its own transport
    // (`baseFetch ?? ponyfill`) rather than swap the transport for nothing.
    // Native `fetch`, bound like starknet's own browser default: the package
    // floor has it everywhere (Node >= 18, evergreen browsers), and
    // cross-fetch would downgrade it — node-fetch 2 on Node (losing
    // connection reuse), an XHR ponyfill in the bundled browser build.
    baseFetch:
      options.timeout > 0
        ? withTimeout(fetch.bind(globalThis), options.timeout)
        : undefined
```

(Companion edit: the `cross-fetch` import on line 2 becomes unused — separate suggestion there.)

src/utils/provider.ts
```diff
@@ -1,5 +1,6 @@
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { RpcProvider } from 'starknet';
+import crossFetch from 'cross-fetch';
```
wa0x6e: Companion to the `baseFetch` suggestion below: with native `fetch` wrapped instead, this import is unused.

```suggestion
```

src/utils/provider.ts
tony8713: Applied verbatim in `88a5567`. Confirmed both halves against the installed deps and measured what the swap costs: cross-fetch opened three sockets for three sequential RPC calls where native `fetch` opened two, and the request it sends carries `Connection: close` against `connection: keep-alive`. The browser ponyfill does hardcode `var ctx = __self__` with the native-fetch line commented out, exactly as you quoted. New test asserts no `Connection: close` on the wire; it fails when I put cross-fetch back and passes on native.

One correction: the encoding half doesn't hold, at least on Node 22 over http both transports send `Accept-Encoding: gzip, deflate`, so connection reuse is the whole of the Node cost. And worth adding in your favour, this restores what master already does — master leaves `baseFetch` unset, so starknet falls back to `fetch-cookie(global.fetch)`. The only thing we still don't get back is that cookie jar, which passing `baseFetch` drops whatever we wrap it around.

src/utils/provider.ts
tony8713: Done in `88a5567`, import gone.

tony8713:
@wa0x6e Both bodies are cut down to the problem and the general shape of the fix: the test and file counts, the revert matrix, the measurements and the per-commit narration are all gone, since those were exactly the lines that kept going stale.

wa0x6e (approved):
utAck

src/utils/provider.ts
wa0x6e: Option 2 reads right to me — `0`-disable staying consistent with `utils.fetch` and the viem transport is the better consistency to keep, and the stock-transport test pins it. Resolving.

PR #1226, opened 2026-08-10
refactor(provider): extract the timeout-bounded fetch into src/utils/fetch.ts

tony8713:
Stacked on #1225, based on `fix/starknet-provider-timeout` rather than `master`, so that PR stays reviewable on its own and this one shows only the extraction. It needs to go in first.

### What this does

`withTimeout` moves out of `src/utils/provider.ts` into a new leaf module, `src/utils/fetch.ts`, with its tests alongside it. Leaf means it imports nothing at all, so it adds no cycle and either side of the duplication can reach it.

The primitive stays generic. The `LibraryError` carrying `code: 'TIMEOUT'` is a constraint of starknet's `RpcChannel`, which re-throws anything else as a bare `Error` and loses the `code`, so it moves to the `getStarknetProvider` call site and reaches the primitive as an optional error factory. The starknet-derived `BaseFetch` type becomes a local `FetchLike`, which is what lets the module import nothing.

### The one behaviour change

The deadline now covers every body reader, not just `.json()`. A caller reading the body any other way was still aborted at the deadline but got the raw `AbortError` rather than the supplied error, and a successful read on those paths left the timer armed instead of clearing it. That was defensible while the wrapper was private to `getStarknetProvider`, whose only callers read JSON; it is not once the same function is a leaf advertising a plain `FetchLike`. Reading `response.body` as a stream is still out of reach and still surfaces the raw abort, and the comment says so.

### What deliberately did not move

`utils.fetch` keeps its own implementation, even though it duplicates this mechanism. Adopting the primitive would change two behaviours of a published function: the deadline would start covering the body read, so a caller with a slow body would begin timing out where it does not today, and a caller `signal` would be composed rather than overridden, so an `AbortError` would newly escape a helper documented to throw `Request timeout after Nms`. That is semver-visible and wants its own PR rather than a ride on an extraction.

No public API change here: `withTimeout` was introduced in #1225 and has never been published, and it is not reachable from `src/index.ts` in either place.

### DIFF
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -1,0 +1,73 @@
+export type FetchLike = (
+  input: RequestInfo | URL,
+  init?: RequestInit
+) => Promise<Response>;
+
+export interface WithTimeoutOptions {
+  // For callers whose stack constrains the error type, e.g. starknet's
+  // RpcChannel. The default is a plain `Error`.
+  timeoutError?: (timeout: number) => Error;
+}
+
+const defaultTimeoutError = (timeout: number) =>
+  new Error(`Request timeout after ${timeout}ms`);
+
+// Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
+// Safari 16.
+export function withTimeout(
+  fetchFn: FetchLike,
+  timeout: number,
+  { timeoutError = defaultTimeoutError }: WithTimeoutOptions = {}
+): FetchLike {
+  // Without this, `setTimeout(..., 0)` aborts every request on the next tick.
+  if (timeout <= 0) return fetchFn;
+
+  return async (url, init) => {
+    const controller = new AbortController();
+    const callerSignal = init?.signal;
+    const abortFromCaller = () => controller.abort(callerSignal?.reason);
+    let timedOut = false;
+
+    const cleanup = () => {
+      clearTimeout(timer);
+      callerSignal?.removeEventListener('abort', abortFromCaller);
+    };
+    // Cleans up after itself: nothing else will if the body is never read.
+    const timer = setTimeout(() => {
+      timedOut = true;
+      controller.abort();
+      cleanup();
+    }, timeout);
+    // Node-only: browser timers are numbers with no `unref`. Without this,
+    // an unread response body holds the event loop open for the full
+    // timeout even though the request already finished.
+    if (typeof timer === 'object' && 'unref' in timer) timer.unref();
+
+    if (callerSignal?.aborted) abortFromCaller();
+    else callerSignal?.addEventListener('abort', abortFromCaller);
+
+    let response: Response;
+    try {
+      response = await fetchFn(url, { ...init, signal: controller.signal });
+    } catch (e) {
+      cleanup();
+      throw timedOut ? timeoutError(timeout) : e;
+    }
+
+    // Resolving only means the headers arrived: a caller reading the body
+    // afterwards with its own `.json()` gets no bound from node-fetch on the
+    // body stream or the parse. Hold the deadline until that settles.
+    const json = response.json.bind(response);
+    response.json = async () => {
+      try {
+        return await json();
+      } catch (e) {
+        throw timedOut ? timeoutError(timeout) : e;
+      } finally {
+        cleanup();
+      }
+    };
+
+    return response;
+  };
+}
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -14,13 +14,15 @@ const defaultTimeoutError = (timeout: number) =>
 
 // Every body reader a `Response` may carry, patched only where it exists:
 // node-fetch 2 ships no `formData`, and a hand-rolled test double has only
-// the reader it needs.
-export const BODY_READERS: readonly (keyof Response)[] = [
+// the reader it needs. `bytes` is spelled out because TypeScript's DOM lib
+// does not carry it yet, though Node 22 does.
+export const BODY_READERS: readonly (keyof Response | 'bytes')[] = [
   'json',
   'text',
   'arrayBuffer',
   'blob',
-  'formData'
+  'formData',
+  'bytes'
 ];
 
 // Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -4,8 +4,8 @@ export type FetchLike = (
 ) => Promise<Response>;
 
 export interface WithTimeoutOptions {
-  // For callers whose stack constrains the error type, e.g. starknet's
-  // RpcChannel. The default is a plain `Error`.
+  // For callers whose stack constrains the error type. The default is a
+  // plain `Error`.
   timeoutError?: (timeout: number) => Error;
 }
 
@@ -15,7 +15,13 @@ const defaultTimeoutError = (timeout: number) =>
 // Every body reader a `Response` may carry, patched only where it exists:
 // node-fetch 2 ships no `formData`, and a hand-rolled test double has only
 // the reader it needs.
-const BODY_READERS = ['json', 'text', 'arrayBuffer', 'blob', 'formData'];
+export const BODY_READERS: readonly (keyof Response)[] = [
+  'json',
+  'text',
+  'arrayBuffer',
+  'blob',
+  'formData'
+];
 
 // Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
 // Safari 16.
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -19,7 +19,9 @@ export function withTimeout(
   timeout: number,
   { timeoutError = defaultTimeoutError }: WithTimeoutOptions = {}
 ): FetchLike {
-  // Without this, `setTimeout(..., 0)` aborts every request on the next tick.
+  // A precondition of the primitive, not a live path: the only caller in this
+  // repo gates the install above 0. Without it, `setTimeout(..., 0)` would
+  // abort every request on the next tick.
   if (timeout <= 0) return fetchFn;
 
   return async (url, init) => {
--- a/src/utils/fetch.ts
+++ b/src/utils/fetch.ts
@@ -12,6 +12,11 @@ export interface WithTimeoutOptions {
 const defaultTimeoutError = (timeout: number) =>
   new Error(`Request timeout after ${timeout}ms`);
 
+// Every body reader a `Response` may carry, patched only where it exists:
+// node-fetch 2 ships no `formData`, and a hand-rolled test double has only
+// the reader it needs.
+const BODY_READERS = ['json', 'text', 'arrayBuffer', 'blob', 'formData'];
+
 // Not `AbortSignal.timeout`: it would raise the browser bundle's floor to
 // Safari 16.
 export function withTimeout(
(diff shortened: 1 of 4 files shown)


src/utils/fetch.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (callerSignal?.aborted) abortFromCaller();
+    else callerSignal?.addEventListener('abort', abortFromCaller);
+
+    let response: Response;
+    try {
+      response = await fetchFn(url, { ...init, signal: controller.signal });
+    } catch (e) {
+      cleanup();
+      throw timedOut ? timeoutError(timeout) : e;
+    }
```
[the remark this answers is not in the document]
tony8713: Right, and fixed in `a14a6e4`. Reproduced your case against a real socket rather than a double: with a stalling body, `.text()` and `.arrayBuffer()` came back as a raw `AbortError`/`DOMException` where `.json()` came back with `code: TIMEOUT`, and the deadline was never cleared on those paths either, so the timer ran to the end and aborted an already-finished request.

Took the first of your two options: every reader the response actually carries is patched now, guarded on existence because node-fetch 2 ships no `formData`. Reverting to `.json()`-only fails 5 of the new tests. One limit worth naming rather than implying it is covered — reading `response.body` as a stream is still out of reach and still surfaces the raw abort; that is in the comment.

tony8713:
One correction on my own reasoning while I was verifying this head, since it was an argument I made to you rather than a count.

The PR body said `utils.fetch` cannot share this implementation because a `.text()` or unread-body caller would leave a live timer holding the event loop, with numbers to match. That measurement predates the `unref` from #1225 (`0c7f2b7`) that this branch is rebased onto, and it no longer reproduces: through the primitive on this head, `.json()` exits at +51ms, `.text()` at +36ms, no read at +35ms, against +23/+24/+21ms for `utils.fetch`. Nothing is held either way.

The conclusion is unchanged, but the real reason is narrower and I have rewritten the body to say it: adopting the primitive would move `utils.fetch`'s deadline to cover the body read and let a caller `AbortError` escape a helper documented to throw `Request timeout after Nms`, which is semver-visible and belongs in its own PR.

src/utils/fetch.ts
```diff
@@ -0,0 +1,91 @@
...
+  timeoutError?: (timeout: number) => Error;
+}
+
+const defaultTimeoutError = (timeout: number) =>
+  new Error(`Request timeout after ${timeout}ms`);
+
+// Every body reader a `Response` may carry, patched only where it exists:
+// node-fetch 2 ships no `formData`, and a hand-rolled test double has only
+// the reader it needs.
+const BODY_READERS = ['json', 'text', 'arrayBuffer', 'blob', 'formData'];
```
wa0x6e: Two things on this line, one suggestion:

1. The list is re-typed as a literal in `test/integration/utils/fetch.spec.ts:124` (`test.each([...])`) — two sources of truth, so a future reader added here ships with silent zero coverage in the parameterized test. Exporting it lets the test import the single source, matching `providers.spec.ts` importing `DEFAULT_TIMEOUT`/`normalizeOptions`. Nothing re-exports this module from `src/index.ts`, so no public surface change.
2. As a bare `string[]` behind the `Record<string, ...>` cast below, a typo'd or stale entry compiles silently (`tsc --noEmit` exits 0 with `'formDataTYPO'` in the list — the reader is just never patched, so its calls quietly fall outside the deadline). Typing the array as `readonly (keyof Response)[]` turns that into `TS2820`. (`satisfies` isn't available: the pinned `@typescript-eslint/parser@4.33.0` can't parse it.)

```suggestion
export const BODY_READERS: readonly (keyof Response)[] = [
  'json',
  'text',
  'arrayBuffer',
  'blob',
  'formData'
];
```

(Two companion suggestions on the spec file complete the export half.)

test/integration/utils/fetch.spec.ts
```diff
@@ -0,0 +1,266 @@
+import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import { AddressInfo, createServer, Server, Socket } from 'net';
+import { withTimeout } from '../../../src/utils/fetch';
```
wa0x6e: Companion to the `BODY_READERS` export suggestion on `src/utils/fetch.ts:18`:

```suggestion
import { BODY_READERS, withTimeout } from '../../../src/utils/fetch';
```

test/integration/utils/fetch.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    await expect(response.json()).rejects.toMatchObject({
+      message: 'gone',
+      code: 'TIMEOUT'
+    });
+  });
+
+  // The primitive advertises a plain `FetchLike`, so a caller is free to read
+  // the body with any of these; the deadline and the error have to be the same
+  // whichever one it picks.
+  test.each(['json', 'text', 'arrayBuffer', 'blob', 'formData'])(
```
wa0x6e: Companion to the `BODY_READERS` export suggestion — the parameterized test then tracks the implementation's list automatically:

```suggestion
  test.each(BODY_READERS)(
```

src/utils/fetch.ts
```diff
@@ -0,0 +1,91 @@
+export type FetchLike = (
+  input: RequestInfo | URL,
+  init?: RequestInit
+) => Promise<Response>;
+
+export interface WithTimeoutOptions {
+  // For callers whose stack constrains the error type, e.g. starknet's
+  // RpcChannel. The default is a plain `Error`.
```
wa0x6e: Nit: the module's code is now starknet-agnostic, but this doc comment still names the starknet stack. Generic phrasing keeps the rationale without the coupling:

```suggestion
  // For callers whose stack constrains the error type. The default is a
  // plain `Error`.
```

test/integration/utils/fetch.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+describe('withTimeout() over a real socket', () => {
+  // Accepts the connection and never answers; a refused connection would fail
+  // fast on its own.
+  let server: Server;
+  let sockets: Socket[] = [];
+  let url: string;
+
+  beforeAll(async () => {
+    server = createServer((socket) => sockets.push(socket));
+    await new Promise<void>((resolve) =>
```
wa0x6e, line 252: The caller-abort test below tears down the client mid-request; if the server-side socket surfaces the RST as `'error'`, there's no listener and an unhandled `'error'` on a `net.Socket` crashes the vitest worker (a confusing flake rather than a failed test). Same gap pre-exists in `providers.spec.ts` — flagging it here since this file is new:

```suggestion
    server = createServer((socket) => {
      socket.on('error', () => {}); // client aborts mid-request; RST is expected
      sockets.push(socket);
    });
```

test/integration/utils/fetch.spec.ts
```diff
@@ -0,0 +1,266 @@
+import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
```
wa0x6e, line 1: Placement thought: 19 of the 21 tests here are pure unit tests over `vi.fn()` doubles — only the trailing `describe` touches a socket. The repo already colocates unit tests with their module (`src/verify/index.spec.ts`, `starknet.spec.ts`, `evm.spec.ts`), so the double-driven tests could live at `src/utils/fetch.spec.ts` next to the primitive, leaving just the real-socket suite in `test/integration/`. Non-blocking — `providers.spec.ts` mixes the same way today.

test/integration/utils/providers.spec.ts
```diff
@@ -356,18 +355,17 @@ describe('Starknet provider timeout', () => {
...
-  test('lets a caller abort cancel a hung request before the timeout', async () => {
-    const caller = new AbortController();
-    const start = Date.now();
-    const request = withTimeout(fetch.bind(globalThis), 5000)(broviderUrl, {
-      method: 'POST',
-      body: '{}',
-      signal: caller.signal
+  // A plain Error would come back out of RpcChannel#errorHandler stripped of
+  // `code`, so the class is what keeps the assertion above true.
+  test('rejects with a LibraryError, the only class that keeps the code', async () => {
```
wa0x6e: This test and the `code: 'TIMEOUT'` one above it each pay a real 400ms hang for what is one rejection — and the comment says as much. One await can carry both assertions:

```ts
test('rejects with the same error code as the EVM provider', async () => {
  const provider = getProvider(STARKNET_NETWORK, {
    broviderUrl,
    timeout: 400
  });
  const request = provider.getSpecVersion();

  // A plain Error would come back out of RpcChannel#errorHandler stripped of
  // `code`, so the class is what keeps the code assertion true.
  await expect(request).rejects.toBeInstanceOf(LibraryError);
  await expect(request).rejects.toMatchObject({ code: 'TIMEOUT' });
});
```

(The second await settles instantly — the promise is already rejected.) Saves 400ms per run. Not a suggestion block because the merge touches the earlier test outside this PR's hunk.

src/utils/fetch.ts
tony8713: Both applied in `c979510`. The typing half checks out: with `'formDataTYPO'` in the list `tsc --noEmit` exits 0 as it stands, and gives `TS2820: Type '"formDataTYPO"' is not assignable to type 'keyof Response'. Did you mean '"formData"'?` once it is `readonly (keyof Response)[]`.

One flip side worth naming since it is now a shared list: deleting a reader drops its `test.each` case silently rather than failing it. The standalone `.text()` timer test still catches that one.

test/integration/utils/fetch.spec.ts
tony8713: Applied in `bbf0848`.

test/integration/utils/fetch.spec.ts
tony8713: Applied in `bbf0848`. Still five cases, still 21 tests in the file.

src/utils/fetch.ts
tony8713: Applied verbatim in `c979510`.

test/integration/utils/fetch.spec.ts
tony8713: Applied in `68dfca9`, and to both servers in `providers.spec.ts` while I was in there, since that file is already in this diff.

Your block as written trips `@typescript-eslint/no-empty-function`, so the comment moved inside the handler body, which is also what silences the rule.

I could not actually provoke the RST: 40 client aborts on Node 22, including one mid-send of a 64MB body, produced no server-side `'error'`. Keeping the guard anyway, because an unhandled one exits the worker with code 1 instead of failing a test, which is exactly the flake you describe.

test/integration/utils/fetch.spec.ts
tony8713: Leaving it where it is, for two reasons. Every test for a `src/utils/*` module lives under `test/integration/utils/`, so splitting would put this module's tests in two places while its neighbour in the same directory stays whole. And there is no runner split to buy: `yarn test:once` is the only test job in CI and it collects `src/**` and `test/**` in one pass, so nothing runs more often for having moved.

If you would rather the `src/verify` precedent win, say so and I will move the double-driven half.

test/integration/utils/providers.spec.ts
tony8713: Done in `8b292d4`, your version as written. The file goes from 1672/1808ms to 1248/1232ms over two runs each.

The merged assertion still bites: swapping `LibraryError` for a plain `Error` in `starknetTimeoutError` fails it on the instanceof line.

test/integration/utils/fetch.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+    await expect(response.json()).rejects.toMatchObject({
+      message: 'gone',
+      code: 'TIMEOUT'
+    });
+  });
+
+  // The primitive advertises a plain `FetchLike`, so a caller is free to read
+  // the body with any of these; the deadline and the error have to be the same
+  // whichever one it picks.
+  test.each(BODY_READERS)(
```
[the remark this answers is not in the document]
tony8713: Right, and the tautology was hiding a live instance: `Response.prototype.bytes` exists on Node 22 and was missing from `BODY_READERS`, so against a real socket that sends headers and then stalls the body, `.bytes()` rejected with `AbortError: This operation was aborted` where `.json()` and `.text()` gave `Request timeout after 300ms`. Fixed in `6ba4196` and `2e52dc2` by adding `bytes` and asserting completeness against `Response.prototype` instead of a literal, which fails on the deletion you describe (dropping `formData` took the file from 21 tests to 20 with typecheck and lint green; it now fails) and also on the omission a literal cannot see. That needed the type widened to `readonly (keyof Response | 'bytes')[]` since TS 5.5's DOM lib has no `bytes`, and the typo check still holds.

tony8713:
Folded into #1225 at @wa0x6e's request: that branch now carries both the timeout fix and the extraction, so there is nothing left here that is only this PR's. Branch kept, nothing lost.

Flagging one thing rather than letting it go quiet with the PR: the test-placement thread above is still open and unanswered. `test/integration/utils/fetch.spec.ts` is part of #1225's diff now, so that is where it should carry on.

PR #1227, opened 2026-08-10
feat(ens): resolve text records via the ENS Universal Resolver

ChaituVR:
## What

`getEnsTextRecord` now resolves through the [ENS Universal Resolver](https://docs.ens.domains/web/ensv2-readiness/) (viem `getEnsText`) instead of the hardcoded resolver-allowlist multicall. The UR address is per-chain config: `ensUniversalResolver` in networks.json (mainnet + Sepolia only; other networks throw `Network not supported`).

`getEnsOwner` is untouched except computing namehash via viem (same ENSIP-15 algorithm). ENSv2 owner support (`findOwner`) comes in a follow-up PR.

## Why

- **Testnet path:** Sepolia runs ENSv2, where the UR is the only way to resolve names (prerequisite for fixing testnet space creation; completed by the follow-up PR).
- **Mainnet fix:** names with a `snapshot` record on any resolver outside the old 2-3 address allowlist resolved to null; now they resolve (supersedes #1203). Records on CCIP-Read/offchain resolvers become readable too.
- Groundwork from #1220/#1221/#1222; approach adapted from @yashgo0018's #1217.

## Behavior notes for reviewers

⚠️ **Mainnet reads switch to the UR** — the original plan (workflow#823) said testnet-only; doing both was a deliberate simplification (single code path, and the UR is what viem/ensjs/ens-app already default to on mainnet). Turning mainnet off = deleting one config line. Explicit sign-off requested.
- A (likely tiny) set of mainnet spaces whose record sat on a non-allowlisted resolver will see their controller change from name-owner to the record's address — records only the name owner could have set. A pre-release diff-scan of all spaces can enumerate them if wanted.
- `options.ensResolvers` is now silently ignored (the allowlist concept is gone).
- Fail-closed: RPC transport failures throw instead of reading as "no record"; resolver-level failures (incl. CCIP gateway errors) return null, matching the previous fallback-to-owner behavior.
- Entry points now import viem eagerly (asserted in dist-entry.spec.ts, flipped intentionally): 556→879ms and +19MB RSS at require() for hub/sequencer (measured); UMD grows 1.48→1.78MB.
- No consumer gets this without deliberately bumping to 0.15.x (`^0.14` never auto-pulls it).

## How to test

**Before (master):** `getEnsTextRecord('stakedao.eth','snapshot','1')` works only because its resolver is allowlisted; any newer resolver → null.
**After:** resolver-agnostic. New live e2e tests with real spaces: ens.eth (legacy `ipns://` record), stakedao.eth (address record), fabien.eth, vitalik.eth (unset → null), aave.eth controller, plus Sepolia: live testnet spaces (boorger.eth, demodao.eth), owner fallback, unowned/non-existent names.

`yarn typecheck && yarn lint && yarn build && yarn test:once` → 402/402 (27 files) locally, incl. the ENSv2 readiness canary.

### DIFF
--- a/src/networks.json
+++ b/src/networks.json
@@ -10,6 +10,7 @@
       "0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41",
       "0xF29100983E058B709F3D539b0c765937B804AC15"
     ],
+    "ensUniversalResolver": "0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe",
     "ensNameWrapper": "0xD4416b13d2b3a9aBae7AcD5D6C2BbDBE25686401",
     "ensSubgraph": "https://subgrapher.snapshot.org/subgraph/arbitrum/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH",
     "rpc": [
@@ -1893,6 +1894,7 @@
     "ensResolvers": [
       "0x8FADE66B79cC9f707aB26799354482EB93a5B7dD"
     ],
+    "ensUniversalResolver": "0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe",
     "ensNameWrapper": "0x0635513f179D50A207757E05759CbD106d7dFcE8",
     "ensSubgraph": "https://subgrapher.snapshot.org/subgraph/arbitrum/CA5t8ep58QKXBdLrjJHaREETHpCsy63kro9Eem6zBQAD",
     "explorer": {
--- a/src/utils/ens.ts
+++ b/src/utils/ens.ts
@@ -79,13 +79,14 @@ export async function getEnsTextRecord(
 
   const client = getViemClient(network, options);
 
-  // viem returns null for unresolvable names and resolver-level failures
-  // (including CCIP gateway errors, matching the previous behavior where
-  // offchain records were unreadable); RPC transport failures still throw
+  // unresolvable names and resolver-level failures resolve to null;
+  // transport failures throw instead of reading as "no record"
   return await client.getEnsText({
     name: normalized,
     key: record,
-    universalResolverAddress
+    universalResolverAddress,
+    blockNumber: options.blockNumber,
+    blockTag: options.blockTag
   });
 }
 
--- a/src/utils/ens.ts
+++ b/src/utils/ens.ts
@@ -1,8 +1,8 @@
 import { Contract } from '@ethersproject/contracts';
 import { getAddress } from '@ethersproject/address';
-import { namehash, ensNormalize } from '@ethersproject/hash';
+import { namehash, normalize } from 'viem/ens';
 import getProvider from './provider';
-import { multicall } from '../multicall';
+import { getViemClient } from './viem';
 import { fetch } from '../utils';
 import networks from '../networks.json';
 
@@ -16,10 +16,6 @@ const MUTED_ERRORS = [
   'UNSUPPORTED_OPERATION'
 ];
 const ENS_REGISTRY = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
-const ENS_ABI = [
-  'function text(bytes32 node, string calldata key) external view returns (string memory)',
-  'function resolver(bytes32 node) view returns (address)' // ENS registry ABI
-];
 const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000';
 
 function getDomainType(domain: string): DomainType {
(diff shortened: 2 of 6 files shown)


src/utils/ens.ts
```diff
@@ -67,42 +63,30 @@ export async function getEnsTextRecord(
...
+  const client = getViemClient(network, options);
+
+  // viem returns null for unresolvable names and resolver-level failures
+  // (including CCIP gateway errors, matching the previous behavior where
+  // offchain records were unreadable); RPC transport failures still throw
+  return await client.getEnsText({
+    name: normalized,
+    key: record,
+    universalResolverAddress
+  });
```
tony8713, line 93: The `timeout` option doesn't reach the CCIP-Read hop, so an offchain resolver can hang this call indefinitely. `getEnsText` never builds a `requestOptions`, and viem's `ccipRequest` then calls bare `fetch` with no signal, so the transport timeout only covers the JSON-RPC leg.

I stood up a gateway that accepts the socket and never replies: `getEnsTextRecord('test.eth', 'snapshot', '1', { timeout: 3000 })` was still pending at 390s on a single `eth_call`, with undici's 300s default as the only backstop. Since the gateway URL comes from the name's own resolver, any name owner can hold a sequencer request open that long, and on master this wasn't reachable at all because text records never did CCIP-Read. Worth wrapping this in a real deadline — #1226's `withTimeout` would drop straight in.

src/utils/ens.ts
```diff
@@ -67,42 +63,30 @@ export async function getEnsTextRecord(
...
+
+  if (!universalResolverAddress) {
+    throw new Error('Network not supported');
+  }
+
+  const client = getViemClient(network, options);
+
+  // viem returns null for unresolvable names and resolver-level failures
+  // (including CCIP gateway errors, matching the previous behavior where
+  // offchain records were unreadable); RPC transport failures still throw
```
tony8713: Small correction to this comment: gateway failures don't uniformly become null. `localBatchGatewayRequest`'s `encodeError` only emits `HttpError` when the error carries an HTTP status; a transport-level failure (refused, reset, DNS) is encoded as a plain `Error`, which `isNullUniversalResolverError` doesn't match, so it rethrows. Doesn't change behaviour since `getSpaceUri` catches either way, but the comment will mislead the next reader.

src/utils/ens.ts
```diff
@@ -125,7 +109,7 @@ export async function getEnsOwner(
   let ensHash: string;
 
   try {
-    ensHash = namehash(ensNormalize(ens));
+    ensHash = namehash(normalize(ens));
```
tony8713: Swapping ethers' `ensNormalize` for viem's `normalize` is a bigger change than "same ENSIP-15 algorithm" suggests — they're different revisions of the tables. Scanning single-codepoint labels, ethers accepts 10,708 that viem rejects, viem accepts 5,771 that ethers rejects, and 6 normalize to different strings (`ẞ.eth` is `ss.eth` under ethers and `ß.eth` under viem — a different namehash, so a different owner).

I checked it against reality and there's nothing to fix: zero divergence across all 16,000 space ids in the hub. Worth a release-note line rather than a code change.

src/networks.json
```diff
@@ -10,6 +10,7 @@
       "0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41",
       "0xF29100983E058B709F3D539b0c765937B804AC15"
     ],
+    "ensUniversalResolver": "0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe",
```
tony8713, line 13: Is there a matching sx-monorepo PR queued behind this? `apps/ui/src/helpers/ens.ts` keeps its own copy of the old allowlist multicall (`deepResolve`) and never calls snapshot.js, so once the sequencer bumps to 0.15.x the UI and the sequencer disagree about the controller for exactly the names this fixes — the UI shows the name owner while the sequencer accepts writes from the record address. Its Sepolia allowlist also carries a fourth resolver that networks.json doesn't have.

test/e2e/utils.spec.js
```diff
@@ -25,12 +26,112 @@ describe('utils', () => {
...
+      await expect(
+        getEnsTextRecord('vitalik.eth', 'snapshot', '1')
+      ).resolves.toBe(null);
+    });
+
+    test('return null for an unset record on testnet', async () => {
+      await expect(
+        getEnsTextRecord('ens.eth', 'snapshot', '11155111')
+      ).resolves.toBe(null);
+    });
```
tony8713, line 76: Nothing in the suite actually exercises the Sepolia Universal Resolver: all four Sepolia names used here return `null` from `getEnsTextRecord`, so every Sepolia `getSpaceController` case falls through to the owner path.

I pointed `ensUniversalResolver` for 11155111 at a junk address and only this one test failed, and only because it threw instead of returning null — the six Sepolia `getSpaceController` cases all still passed. Given ENSv2 on Sepolia is the main motivation, one testnet name with a real `snapshot` record would be worth adding. Mainnet is well covered by comparison: swapping in the legacy UR fails 8 tests.

tony8713 (changes requested):
Cloned both sides and ran them rather than reading the diff: typecheck, lint, build and the full suite pass on each (master 387/387, this branch 403/403), and the rollup cycle list is identical at 7 on both. No overlap with #1225/#1226 — those only touch `provider.ts`/`fetch.ts`, and this actually benefits from 1226's `normalizeOptions` hardening if that lands first.

One thing I think needs fixing before merge, plus the answer to your mainnet sign-off question.

**Mainnet sign-off: I ran the diff-scan.** `getSpaceUri` from master vs this branch, side by side, over the 15,932 mainnet `.eth` spaces I could page out of the hub. Zero controller changes. 15,783 of them differ only by `''` -> `null` (no record at all), which every consumer already treats as falsy — the sequencer's `poke.ts` does `if (uri)`. The other 149 have a real record and every one resolves identically. So the population you were worried about is empirically zero. The flip side is that the mainnet upside is currently zero too: not one existing space has a `snapshot` record on a non-allowlisted resolver. I'd sign off on doing mainnet.

**The effective timeout quadrupled.** viem retries 3 times, so `timeout` is now per-attempt rather than a budget. Against a black-hole RPC, master threw at 25.0s after one TCP connection; this branch throws at 101.4s after four. Probably acceptable, but the sequencer's worst case for a single space read went from 25s to ~100s — `retryCount: 0` on the transport restores the old shape.

**The require() cost is bigger than the description says.** Median of 7 runs on `dist/snapshot.cjs.js`: 556ms -> 879ms, RSS +19MB, 517 -> 1061 modules — not ~100ms/+3MB. The UMD browser bundle also goes 1.48 -> 1.78MB (435 -> 542KB gzipped), since viem is external for cjs/esm but inlined there.

Two small ones not worth their own thread: the old `...multicallOptions` spread meant a caller could pin `blockTag`, which is now silently dropped (viem's `getEnsText` takes `blockNumber`/`blockTag`, so it's a one-line forward), and `ensResolvers` isn't really gone — it's still the network-support gate in `getEnsOwner`, so a network configured with only `ensUniversalResolver` would read records but throw on the owner fallback.

src/networks.json
yashgo0018: There is a draft pr on the sx-monorepo, https://github.com/snapshot-labs/sx-monorepo/pull/2198

There also needs to be another pr on stamps api, which i think also depends on snapshot.js.

tony8713 (approved):
Approving. My blocking objection was the CCIP-Read timeout, and that one is cleared: the deadline now reaches the CCIP hop rather than only the JSON-RPC leg, so a hanging offchain gateway can no longer hold a request open past it.

The normalization blocker I raised after that is fixed by restoring `ensNormalize` and keeping viem only for `namehash`. I also owe you a correction there. My earlier "zero divergence" claim on that swap was wrong, and you resolved a thread on the strength of it. Viem's `normalize` did reject real `.eth` space ids, which would have made `getSpaceController` return the zero address for them and locked those spaces out of their own settings.

One non-blocking gap remains. Nothing asserts a record actually resolving through the Sepolia Universal Resolver, so a wrong or dead Sepolia UR address would still pass the suite. Worth a follow-up rather than holding this PR.

PR #1230, opened 2026-08-14
feat(ens): resolve name owners via the Universal Resolver findOwner

### LINKED ISSUE #1232: getEnsTextRecord reads gateway failures as a missing record
`getEnsTextRecord` calls viem's `getEnsText` non-strict, and viem maps every CCIP gateway `HttpError` to `null` regardless of status. A gateway that is rate limiting or down therefore reads as "this name has no snapshot record" rather than as a failure.

The hazard is not the null itself, it is where the null goes. For a space that uses the `snapshot` text record to name a controller, a gateway hiccup makes `getEnsSpaceController` fall through to the ENS owner and silently resolve a **different** controller. That is a wrong authorization answer, not an outage.

Verified live against a gateway currently returning 503: `getEnsTextRecord` returns null and `getSpaceUri` returns null. `getSpaceController` happens to be safe today only because it falls through to `getEnsOwner`, which throws after #1231.

The comment above that call, saying transport failures throw instead of reading as no record, is not true for gateway HTTP failures.

Fix direction: the same `getEnsAddressStrict` treatment #1231 introduced, so only reverts that genuinely mean "no record" return null. A 429/5xx retry inside `ccipReadWithDeadline` would help both call sites.

See #1231 for context, this is master's problem and not that PR's.

ChaituVR:
## What

Completes the Universal Resolver migration started in #1227:

1. **ENSv2 owner resolution, Sepolia only**: `getEnsOwner` asks the UR's `findOwner` on Sepolia (`11155111`), where ENSv2 is live. A name absent from ENSv2 is a successful zero-address return (verified against the live Sepolia UR), which falls back to the v1 registry + NameWrapper unwrap; **any revert throws** — there is no revert `findOwner` legitimately produces, so none may resolve a stale v1 owner. **Mainnet never calls `findOwner`** — it stays on the v1 registry path exactly as on master (ENSv2 is not deployed there). Name encoding uses viem's `packetToBytes` — the same wire encoding as every other UR call.
2. **Strict record reads** (closes #1232): `getEnsText`/`getEnsAddress` run `strict: true` with one shared revert classifier, parameterized by name class. Genuine no-record answers (no resolver, `NotImplemented()` reverts, gateway 404) resolve as before; failures throw. A resolver's **bare `revert()`** is scoped by name: for **other-tld names** it is a no-record answer — un-imported DNS domains answer every read that way (we3home.com, jolt.finance stay on master behavior) — while for **.eth names and subdomains** it is a resolver failure and throws, since those are the names where a swallowed failure could authorize the name owner in place of the controller. `getSpaceUri` no longer swallows errors, so failures reach `getSpaceController` instead of silently resolving the name owner as controller.
3. Continues @yashgo0018's #1217 (org-fork branches can't be updated by maintainers; his commits are in this branch's history).

## Behavior change to be aware of (deliberate) — ships as 0.17.0

`getSpaceUri` / `getSpaceController` / `getEnsOwner` now **reject** on RPC or CCIP-gateway failure where they used to resolve `null` / the name owner. Fail-closed is the point — a wrong authorization answer is worse than a loud error — and it is why this is a minor bump, not a patch.

**Spaces affected today** — record behind a dead CCIP gateway, so they error until the gateway heals instead of silently falling back to the name owner: **pooltogether.eth, rdr.eth, greenpill.eth, blossomlabs.eth** (503) and **edenonline.eth** (500). That is from a sweep of the newest 4,000 mainnet spaces; the full list is longer.

**Downstream call sites that must expect a rejection** (none handle it today; all pin with a caret, so they adopt this on their next bump):

| Repo | Call site | Handled by |
|---|---|---|
| sx-monorepo | `apps/sequencer/src/writer/settings.ts:44` `getSpaceController` | snapshot-labs/sx-monorepo#2273 |
| sx-monorepo | `apps/sequencer/src/writer/delete-space.ts:15` `getSpaceController` | snapshot-labs/sx-monorepo#2273 (via the shared helper) |
| sx-monorepo | `apps/sequencer/src/helpers/poke.ts:28` `getSpaceUri` | snapshot-labs/sx-monorepo#2273 |
| snapshot-v1 | `src/composables/useSpaceController.ts` `getEnsOwner` / `getSpaceController` | to do on its bump |
| snapshot-sidekick | `src/lib/nftClaimer/utils.ts` `getSpaceController` | caller `validateSpace()` throws by design |

sx-monorepo `apps/ui` is not affected — its `helpers/ens.ts` is an independent implementation.

## Proof

- Live e2e: `test123.eth` (real ENSv2 Sepolia registration) resolves via `findOwner` — returns 0x0 on master.
- Mainnet: `getEnsOwner` byte-identical old vs new over 3,491 mainnet + 944 testnet spaces (reviewer-verified); a unit test pins the mainnet path to a single registry read, no `findOwner`.
- Mocked unit tests (`src/utils/ens.spec.ts`, 21) pin every behaviour this PR exists to make, each verified to go red when its change is reverted: `strict: true` on both resolver calls, `getSpaceUri` propagating to `getSpaceController`, `findOwner` actually called on Sepolia (string or numeric id), every `findOwner` failure throwing, and the classifier both ways (no-record → `null`, 5xx / data-carrying `ResolverError` / transport → throw).
- Live e2e for the exotic classes: DNS-imported space (defi.app), a name the strict DNS wire format cannot carry (84-byte emoji label), broken-gateway behavior verified against live 503s.
- 463/463 (29 files).

## How to test

**Before:** `getSpaceController('test123.eth', '11155111')` → `0x0000…`
**After:** → `0x1208a26FAa0F4AC65B42098419EB4dAA5e580AC6`
**Mainnet unchanged:** `getSpaceController('ens.eth', '1')` → `0xb6E040C9ECAaE172a89bD561c5F73e1C48d28cd9` on both.
**Fail-closed:** `getSpaceController('pooltogether.eth', '1')` → rejects (gateway 503) instead of the registry owner.
`yarn typecheck && yarn lint && yarn build && yarn test:once` → 463/463

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "dev": "rollup -c -w",
     "lint": "eslint . --ext .ts",
     "lint:fix": "yarn lint --fix",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit -p tsconfig.typecheck.json",
     "prepublishOnly": "yarn build"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.14.26",
+  "version": "0.16.0",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
@@ -23,8 +23,8 @@
     "cross-fetch": "^3.1.6",
     "json-to-graphql-query": "^2.2.4",
     "lodash.set": "^4.3.2",
-    "starknet": "^6.11.0",
-    "viem": "^2.55.10"
+    "starknet": "^6.21.0",
+    "viem": "^2.55.11"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^18.1.0",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "cross-fetch": "^3.1.6",
     "json-to-graphql-query": "^2.2.4",
     "lodash.set": "^4.3.2",
-    "starknet": "^6.11.0"
+    "starknet": "^6.11.0",
+    "viem": "^2.55.10"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^18.1.0",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   "browser": "dist/snapshot.min.js",
   "types": "dist/src/index.d.ts",
   "dependencies": {
-    "@ensdomains/eth-ens-namehash": "^2.0.15",
     "@ethersproject/abi": "^5.6.4",
     "@ethersproject/address": "^5.6.1",
     "@ethersproject/bignumber": "^5.7.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.16.0",
+  "version": "0.17.0",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "prepublishOnly": "yarn build"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "files": [
     "dist",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -3,6 +3,8 @@ import { Contract } from '@ethersproject/contracts';
 import { getAddress, isAddress } from '@ethersproject/address';
 import { parseUnits } from '@ethersproject/units';
 import { namehash, ensNormalize } from '@ethersproject/hash';
+import { Interface } from '@ethersproject/abi';
+import { hexlify, concat } from '@ethersproject/bytes';
 import { jsonToGraphQLQuery } from 'json-to-graphql-query';
 import Ajv from 'ajv';
 import addFormats from 'ajv-formats';
@@ -39,10 +41,23 @@ const MUTED_ERRORS = [
   'UNSUPPORTED_OPERATION'
 ];
 const ENS_REGISTRY = '0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e';
-const ENS_ABI = [
-  'function text(bytes32 node, string calldata key) external view returns (string memory)',
-  'function resolver(bytes32 node) view returns (address)' // ENS registry ABI
+const UNIVERSAL_RESOLVER_ADDRESS =
+  '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe';
+const UNIVERSAL_RESOLVER_ABI = [
+  'error ResolverNotFound(bytes name)',
+  'error ResolverNotContract(bytes name, address resolver)',
+  'error UnsupportedResolverProfile(bytes4 selector)',
+  'error ResolverError(bytes errorData)',
+  'error ReverseAddressMismatch(string primary, bytes primaryAddress)',
+  'error HttpError(uint16 status, string message)',
+  'function resolve(bytes name, bytes data) view returns (bytes result, address resolver)',
+  'function findOwner(bytes name) view returns (address owner)'
 ];
+const RESOLVER_ABI = [
+  'function text(bytes32 node, string key) view returns (string)',
+  'function addr(bytes32 node) view returns (address)'
+];
+const resolverInterface = new Interface(RESOLVER_ABI);
 const UD_MAPPING = {
   '146': {
     tlds: ['.sonic'],
@@ -225,6 +240,43 @@ function getDomainType(domain: string): DomainType {
   else if (tokens.length > 2) return 'subdomain';
   else if (isEns) return 'ens';
   else throw new Error('Invalid domain');
+}
+
+function dnsEncodeName(name: string): string {
+  const normalized = ensNormalize(name);
+  const labels = normalized.split('.');
+  const encodedLabels = labels.map((label) => {
+    if (!label.length || label.length > 63) {
+      throw new Error(`Invalid ENS label: ${label}`);
+    }
+    const labelBytes = new TextEncoder().encode(label);
+    return concat([Uint8Array.from([labelBytes.length]), labelBytes]);
+  });
+  return hexlify(concat([...encodedLabels, new Uint8Array([0])]));
+}
+
+function getUniversalResolverAddress(network: string, options: any = {}) {
+  return (
+    options.ensUniversalResolver ||
+    networks[network]?.ensUniversalResolver ||
+    UNIVERSAL_RESOLVER_ADDRESS
+  );
+}
+
+async function unwrapNameWrapperOwner(
+  owner: string,
+  ensHash: string,
+  ensNameWrapper: string,
+  provider: any
+): Promise<string> {
+  if (owner !== ensNameWrapper) return owner;
+
+  const ensNameWrapperContract = new Contract(
+    ensNameWrapper,
+    ['function ownerOf(uint256) view returns (address)'],
+    provider
+  );
+  return ensNameWrapperContract.ownerOf(ensHash);
 }
 
 // see https://docs.ens.domains/registry/dns#gasless-import
@@ -615,42 +667,38 @@ export async function getEnsTextRecord(
   network = '1',
   options: any = {}
 ) {
-  const {
-    ensResolvers = networks[network]?.ensResolvers ||
-      networks['1'].ensResolvers,
-    broviderUrl,
-    ...multicallOptions
-  } = options;
+  const { broviderUrl } = options;
 
   let ensHash: string;
+  let dnsEncodedName: string;
 
   try {
-    ensHash = namehash(ensNormalize(ens));
+    const normalized = ensNormalize(ens);
+    ensHash = namehash(normalized);
+    dnsEncodedName = dnsEncodeName(normalized);
   } catch (e: any) {
     return null;
   }
 
   const provider = getProvider(network, { broviderUrl });
+  const universalResolver = new Contract(
+    getUniversalResolverAddress(network, options),
+    UNIVERSAL_RESOLVER_ABI,
+    provider
+  );
 
-  const calls = [
-    [ENS_REGISTRY, 'resolver', [ensHash]], // Query for resolver from registry
-    ...ensResolvers.map((address: string) => [
-      address,
-      'text',
-      [ensHash, record]
-    ]) // Query for text record from each resolver
-  ];
-
-  const [[resolverAddress], ...textRecords] = (await multicall(
-    network,
-    provider,
-    ENS_ABI,
-    calls,
-    multicallOptions
-  )) as string[][];
-
-  const resolverIndex = ensResolvers.indexOf(resolverAddress);
-  return resolverIndex !== -1 ? textRecords[resolverIndex]?.[0] : null;
+  try {
+    const data = resolverInterface.encodeFunctionData('text', [
+      ensHash,
+      record
+    ]);
+    const [result] = await universalResolver.resolve(dnsEncodedName, data, {
+      ccipReadEnabled: true
+    });
+    return resolverInterface.decodeFunctionResult('text', result)[0];
+  } catch (e: any) {
+    return null;
+  }
 }
 
 export async function getSpaceUri(
@@ -671,7 +719,7 @@ export async function getEnsOwner(
   network = '1',
   options: any = {}
 ): Promise<string> {
-  if (!networks[network]?.ensResolvers?.length) {
+  if (!networks[network]?.ensUniversalResolver) {
     throw new Error('Network not supported');
   }
 
@@ -684,24 +732,47 @@ export async function getEnsOwner(
   );
 
   let ensHash: string;
+  let dnsEncodedName: string;
 
   try {
-    ensHash = namehash(ensNormalize(ens));
+    const normalized = ensNormalize(ens);
+    ensHash = namehash(normalized);
+    dnsEncodedName = dnsEncodeName(normalized);
   } catch (e: any) {
     return EMPTY_ADDRESS;
   }
 
   const ensNameWrapper =
     options.ensNameWrapper || networks[network].ensNameWrapper;
-  let owner = await ensRegistry.owner(ensHash);
-  // If owner is the ENSNameWrapper contract, resolve the owner of the name
-  if (owner === ensNameWrapper) {
-    const ensNameWrapperContract = new Contract(
+  const universalResolver = new Contract(
+    getUniversalResolverAddress(network, options),
+    UNIVERSAL_RESOLVER_ABI,
+    provider
+  );
+
+  let owner = EMPTY_ADDRESS;
+
+  // Prefer Universal Resolver findOwner (ENSv2); fall back when unavailable
+  try {
+    owner = await universalResolver.findOwner(dnsEncodedName);
+    owner = await unwrapNameWrapperOwner(
+      owner,
+      ensHash,
       ensNameWrapper,
-      ['function ownerOf(uint256) view returns (address)'],
       provider
     );
-    owner = await ensNameWrapperContract.ownerOf(ensHash);
+  } catch (e: any) {
+    owner = EMPTY_ADDRESS;
+  }
+
+  if (!owner || owner === EMPTY_ADDRESS) {
+    owner = await ensRegistry.owner(ensHash);
+    owner = await unwrapNameWrapperOwner(
+      owner,
+      ensHash,
+      ensNameWrapper,
+      provider
+    );
   }
 
   if (owner === EMPTY_ADDRESS && domainType === 'other-tld') {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,15 +2,14 @@ import crossFetch from 'cross-fetch';
 import { Contract } from '@ethersproject/contracts';
 import { getAddress, isAddress } from '@ethersproject/address';
 import { parseUnits } from '@ethersproject/units';
-import { parseAbi, concat, stringToBytes, toHex } from 'viem';
-import { namehash, normalize } from 'viem/ens';
-import type { Address, Hex, PublicClient } from 'viem';
+import { namehash, ensNormalize } from '@ethersproject/hash';
 import { jsonToGraphQLQuery } from 'json-to-graphql-query';
 import Ajv from 'ajv';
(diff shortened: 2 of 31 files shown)


src/utils/ens.ts
```diff
@@ -113,29 +148,34 @@ export async function getEnsTextRecord(
...
-    blockTag: options.blockTag
-  });
+  try {
+    return await client.getEnsText({
+      name: normalized,
+      key: record,
+      universalResolverAddress,
+      blockNumber: options.blockNumber,
+      blockTag: options.blockTag,
+      strict: true
```
tony8713, line 151: This throw never reaches a caller, so #1232 is still live: `getSpaceUri` catches everything and returns null, and nothing in snapshot-sequencer or snapshot-hub calls `getEnsTextRecord` directly. `pooltogether.eth` sits on a 503 gateway right now, and `getSpaceController('pooltogether.eth', '1')` returns `0x029Aa20Dcc15c022b1b61D420aaCf7f179A9C73f`, the registry owner, on this branch exactly as on master. Either make `getSpaceUri` rethrow or drop the "closes #1232".

src/utils/ens.ts
```diff
@@ -30,8 +45,41 @@ function getDomainType(domain: string): DomainType {
...
 
-// strict mode so gateway/infra failures throw; only reverts that mean the
-// name does not resolve (or a gateway 404) read as "no address"
+// takes an already-normalized name; DNS wire format per RFC 1035
+function dnsEncodeName(name: string): Hex {
+  const labels = name.split('.');
+  const encodedLabels = labels.map((label) => {
+    // DNS limits are in bytes, not characters
+    const labelBytes = stringToBytes(label);
+    if (!labelBytes.length || labelBytes.length > 63) {
```
tony8713: The scan undercounts: 11 of the 90,529 hub spaces fail this, three of them `.eth` with live proposals (the vampire name in your test, plus 🇨🇦x12.eth with 2 proposals and 👩🏾‍❤‍👩🏿x4.eth). The 63 byte limit is also stricter than the contract you are calling: viem's `packetToBytes`, which `getEnsText` above already uses on these same names, encodes them with a raw length byte, and the Sepolia UR parses those 84 and 96 byte labels and returns `0x0` rather than reverting, while it does revert on genuinely malformed input. Swapping the hand-rolled encoder for `packetToBytes` would keep the two functions agreeing and stop those spaces skipping `findOwner` once mainnet has v2.

src/utils/ens.ts
```diff
@@ -147,14 +187,50 @@ export async function getEnsOwner(
...
+        abi: UNIVERSAL_RESOLVER_ABI,
+        functionName: 'findOwner',
+        args: [dnsEncodedName]
+      });
+    } catch (e: any) {
+      const revert =
+        typeof e?.walk === 'function'
+          ? e.walk((err: any) => err instanceof ContractFunctionRevertedError)
+          : undefined;
+      if (!revert) throw e;
```
tony8713: An `HttpError` revert lands here and reads as unowned, which is exactly the distinction `isNoRecordRevert` makes two functions up. Not reachable today, I measured zero gateway fetches from `findOwner` on Sepolia, but v2 resolution is meant to go offchain and then a 5xx gateway silently returns `EMPTY_ADDRESS` instead of throwing. Reusing `isNoRecordRevert` here would keep it consistent with the rest of the file.

tony8713 (changes requested):
The owner half holds up. I ran `getEnsOwner` old vs new over 3,491 real hub space ids on mainnet and all 944 testnet spaces: zero divergence, zero errors. `findOwner` reverts on 599/599 mainnet names and returns `0x0` for all 944 existing Sepolia spaces, so today it buys one extra `eth_call` per `getSpaceController` and only answers for fresh v2 registrations like `test123.eth`, which resolves exactly as you describe.

Two claims in the description don't survive testing though, comments below.

src/utils/ens.ts
```diff
@@ -30,8 +39,26 @@ function getDomainType(domain: string): DomainType {
...
+    typeof e?.walk === 'function'
+      ? e.walk((err: any) => err instanceof ContractFunctionRevertedError)
+      : undefined;
+  const errorName = (revert as any)?.data?.errorName;
+  const args = (revert as any)?.data?.args;
+  return (
+    errorName === 'ResolverNotFound' ||
+    errorName === 'ResolverNotContract' ||
+    errorName === 'UnsupportedResolverProfile' ||
+    (errorName === 'ResolverError' && args?.[0] === '0x') ||
```
tony8713: `ResolverError(NotImplemented())` is a no-record answer and this throws on it, taking four live spaces with it: defi.app (5 proposals, 213 followers), lightchain.ai (15 proposals, 241 followers), arc.market and artha.world. None of them has a v1 resolver, so they resolve through the TLD-level DNS resolver, which reverts `0xd6234725` = `NotImplemented()` for the `text` profile; master read that as null and fell through to the `getDNSOwner` path these spaces are designed to use. Accepting `NotImplemented()` alongside `0x` restores them without loosening the 5xx handling.

src/utils.ts
```diff
@@ -563,12 +563,9 @@ export async function getSpaceUri(
...
 ): Promise<string | null> {
-  try {
-    return await getEnsTextRecord(id, 'snapshot', network, options);
-  } catch (e: any) {
-    console.log(e);
-    return null;
-  }
+  // no catch: a failed read must not look like a missing record, or the
+  // controller silently falls back to the name owner
+  return await getEnsTextRecord(id, 'snapshot', network, options);
```
tony8713, line 568: Nine live spaces stop resolving a controller once this rethrows, seven on mainnet and two on Sepolia. Four are the `NotImplemented()` bug above; the rest are genuinely dead gateways, pooltogether.eth (38 proposals, 5,228 followers) and rdr.eth on 503, edenonline.eth on 500, all still failing 9.5 hours after I first hit them, so those spaces are indefinitely un-editable rather than briefly. That is the fail-closed trade working as designed and I am not asking you to revert it, but it belongs in the description instead of "zero regressions", and note that `getSpaceController` in the sequencer's settings.ts and `getSpaceUri` in poke.ts both sit outside their try blocks, so the raw viem error becomes the rejection value.

tony8713 (changes requested):
Both fixes land, and the `packetToBytes` swap is clean: `getEnsOwner` is byte-identical old vs new over 3,491 mainnet and all 944 testnet space ids, and `findOwner` never produces a decodable revert on either network (6,000/6,000 mainnet reverts undecodable, 944/944 Sepolia clean), so the new throw branch is unreachable today.

Two things left. The narrowed classifier takes four live spaces down with it, comment below. And the description no longer matches the code: item 1 still describes names that "skip `findOwner` and keep resolving via the hash-based registry", which this push deleted, and "Differentials ... zero regressions" is no longer true.

src/utils/ens.ts
```diff
@@ -147,14 +168,42 @@ export async function getEnsOwner(
...
+      });
+    } catch (e: any) {
+      const revert =
+        typeof e?.walk === 'function'
+          ? e.walk((err: any) => err instanceof ContractFunctionRevertedError)
+          : undefined;
+      if (!revert || (revert as any).data?.errorName) throw e;
+      owner = EMPTY_ADDRESS;
+    }
+  }
```
wa0x6e: **Fail-open on undecoded `findOwner` reverts.** This catch treats any revert without a decoded `errorName` as "name not migrated to ENSv2" and silently falls back to the v1 registry owner. Measured against the live Sepolia Universal Resolver (`0xeEeE…EeEe`): unmigrated and nonexistent names (`bob.eth`, `ens.eth`, `snapshotdoesnotexist123456789.eth`) never revert — `findOwner` returns a successful `0x000…0`, and the v1 fallback happens via the normal success path. The deployed `UniversalResolverV2.findOwner` source (verified via Sourcify) confirms it: "no owner" is a successful zero-address return, and the resolver-error family (`ResolverNotFound`, `HttpError`, …) is thrown only by `resolve()`/`reverse()` paths `findOwner` never reaches.

So this fallback has no legitimate trigger: every revert `findOwner` can actually produce (paused/upgraded proxy, wrong UR address, unknown custom error, dataless `execution reverted`) is a genuine failure — and this branch converts it into a silently-resolved stale v1 owner, the exact outcome the comment above says must not happen. Fail-closed here means: any revert throws.

```suggestion
  // findOwner is ENSv2-only, live on Sepolia and not yet on mainnet; a name
  // absent from ENSv2 resolves EMPTY_ADDRESS successfully (verified against
  // the live Sepolia Universal Resolver), so any revert here is a genuine
  // failure and must throw, never fall back to a stale v1 owner
  if (network === '11155111') {
    owner = await client.readContract({
      address: universalResolverAddress,
      abi: UNIVERSAL_RESOLVER_ABI,
      functionName: 'findOwner',
      args: [dnsEncodedName]
    });
  }
```

Companion test update in `src/utils/ens.spec.ts` (the current test locks in the fallback):

```diff
-  test('falls back to the registry when findOwner reverts undecoded', async () => {
+  test('throws when findOwner reverts undecoded', async () => {
     const client = mockClient();
-    client.readContract
-      .mockRejectedValueOnce(undecodedRevert())
-      .mockResolvedValueOnce(OWNER);
-    await expect(getEnsOwner('x.eth', '11155111', opts)).resolves.toBe(OWNER);
-    expect(client.readContract).toHaveBeenCalledTimes(2);
+    client.readContract.mockRejectedValueOnce(undecodedRevert());
+    await expect(getEnsOwner('x.eth', '11155111', opts)).rejects.toBeDefined();
+    expect(client.readContract).toHaveBeenCalledTimes(1);
   });
```

With both applied: `yarn vitest run src/utils/ens.spec.ts` 15/15. (This also removes the now-unused `ContractFunctionRevertedError` import if nothing else references it — it's still used by `isNoRecordRevert`, so no import change needed.)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.16.0",
+  "version": "0.16.1",
```
wa0x6e: **Throwing behavior ships as a patch bump, and the consumer audit misses call sites.** `getSpaceUri`/`getSpaceController` now reject on transient RPC/CCIP-gateway failures where they previously resolved `null` — deliberate and documented, but it ships as `0.16.0 → 0.16.1`, and every first-party consumer pins `@snapshot-labs/snapshot.js` with a caret range, so they adopt the new rejection contract on their next lockfile refresh with no signal. #1227 (`5f2dbfba`), which started this same UR migration, took a MINOR bump (`0.15.2 → 0.16.0`).

```suggestion
  "version": "0.17.0",
```

Audited the actual downstream call sites (fetched current files from GitHub):

| Repo | File:line | Calls | Locally handled? | Pin | In PR body? |
|---|---|---|---|---|---|
| sx-monorepo | `apps/sequencer/src/writer/settings.ts:44` | `getSpaceController` | No | `^0.15.1` | Yes |
| sx-monorepo | `apps/sequencer/src/writer/delete-space.ts:15` | `getSpaceController` | No | `^0.15.1` | **No** |
| sx-monorepo | `apps/sequencer/src/helpers/poke.ts:28` | `getSpaceUri` | No | `^0.15.1` | Yes |
| snapshot-v1 | `src/composables/useSpaceController.ts:80-92` | `getEnsOwner`, `getSpaceController` | No | `^0.14.18` | **No** |
| snapshot-sidekick | `src/lib/nftClaimer/utils.ts:81-88` | `getSpaceController` | No (caller `validateSpace()` throws by design) | `^0.14.24` | **No** |

`apps/sequencer/src/writer/delete-space.ts:15` is the notable one — same unhandled-rejection shape as the `settings.ts` site the PR body names, sitting right next to it, but absent from the audit. (sx-monorepo `apps/ui` is *not* affected — its `helpers/ens.ts` is an independent local ENS implementation, zero imports of this package's ENS helpers.)

Worth adding `delete-space.ts` / snapshot-v1 / sidekick to the PR body's migration note so the sequencer/UI bumps handle the rejection path knowingly.

src/utils/ens.ts
```diff
@@ -147,14 +168,42 @@ export async function getEnsOwner(
...
-    abi: ENS_REGISTRY_ABI,
-    functionName: 'owner',
-    args: [ensHash]
-  });
+
+  let owner: string = EMPTY_ADDRESS;
+
+  // findOwner is ENSv2-only, live on Sepolia and not yet on mainnet; a decoded
+  // revert or transport failure must throw, not resolve a stale v1 owner
+  if (network === '11155111') {
```
wa0x6e: **Numeric `network` silently skips the ENSv2 path.** `networks[network]` resolves for a numeric `11155111` (JS coerces object keys), but this strict `===` against the string doesn't — so an untyped JS caller passing the chain id as a number gets the v1-registry answer with no error, a silent stale-owner variant of the same fail-open theme. Elsewhere in the repo a numeric network for a supported chain fails loud (`ENS_CHAIN_IDS.includes(network)` → "Network not supported"); this is the one place it fails silently with a plausible-looking wrong answer.

Suggest coercing in the gate (also applies to the rewritten block proposed in my other comment on this range):

```js
if (String(network) === '11155111') {
```

src/utils.ts
wa0x6e, line 568: Heads-up on release sequencing for this PR's fail-closed change.

`getSpaceUri` now throws instead of resolving `null` on RPC/gateway
failure. In sx-monorepo, `apps/sequencer/src/helpers/poke.ts` calls it
outside any try block, so the error reaches
`apps/sequencer/src/api.ts`'s `GET /spaces/:key/poke` handler, which does
`res.json({ error: err })` with no Sentry capture. I reproduced this
against `pooltogether.eth` (one of the gateway-broken spaces named
above) — the serialized response is ~9.9 KB and includes the RPC
endpoint URL and the full eth_call request body.

sx-monorepo#2273 fixes this (catch + capture + clean rejection message)
but is still a draft and unmerged. sx-monorepo's dependabot.yml auto-bumps
`@snapshot-labs/*` daily, so once 0.17.0 ships to npm a routine-looking
bump PR for the sequencer can land before #2273 does, making this leak
live with zero observability in the meantime.

Suggest: don't publish 0.17.0 to npm until #2273 is merged, or otherwise
hold/pause the sequencer's dependabot bump for @snapshot-labs/snapshot.js
until it lands.

src/utils/ens.ts
```diff
@@ -147,14 +173,34 @@ export async function getEnsOwner(
...
-    args: [ensHash]
-  });
+
+  let owner: string = EMPTY_ADDRESS;
+
+  // findOwner is ENSv2-only, live on Sepolia and not yet on mainnet. A name
+  // absent from ENSv2 resolves EMPTY_ADDRESS successfully, so any revert is a
+  // genuine failure and must throw, never fall back to a stale v1 owner
+  if (String(network) === '11155111') {
+    owner = await client.readContract({
```
wa0x6e: `dnsEncodedName` is computed on every call but only consumed inside the Sepolia branch below — on mainnet (the hot path) it is computed and discarded. Moving it inside the branch keeps the mainnet path allocation-free:

```suggestion
  const ensNameWrapper =
    options.ensNameWrapper || networks[network].ensNameWrapper;

  let owner: string = EMPTY_ADDRESS;

  // findOwner is ENSv2-only, live on Sepolia and not yet on mainnet. A name
  // absent from ENSv2 resolves EMPTY_ADDRESS successfully, so any revert is a
  // genuine failure and must throw, never fall back to a stale v1 owner
  if (String(network) === '11155111') {
    // viem's own encoding, so labels the strict DNS format rejects still resolve
    const dnsEncodedName = toHex(packetToBytes(normalized));
    owner = await client.readContract({
```

wa0x6e (approved):
utAck

src/utils/ens.ts
```diff
@@ -30,8 +44,43 @@ function getDomainType(domain: string): DomainType {
...
+    typeof e?.walk === 'function'
+      ? e.walk((err: any) => err instanceof ContractFunctionRevertedError)
+      : undefined;
+  const errorName = (revert as any)?.data?.errorName;
+  const args = (revert as any)?.data?.args;
+  return (
+    errorName === 'ResolverNotFound' ||
+    errorName === 'ResolverNotContract' ||
+    errorName === 'UnsupportedResolverProfile' ||
+    (errorName === 'ResolverError' && args?.[0] === NOT_IMPLEMENTED_ERROR) ||
```
wa0x6e: **Dropping the `'0x'` clause here breaks controller resolution for live DNS-domain spaces.**

An un-imported DNS domain answers a `text()` read with a bare `revert()` — the exact `ResolverError('0x')` this line stopped classifying as "no record". Verified live against mainnet through this branch's own exports:

| name | `getEnsOwner` | `getSpaceController` | at d8b43d5 / master |
|---|---|---|---|
| `we3home.com` (live space) | `0x0` | **rejects** | `EMPTY_ADDRESS` |
| `jolt.finance` (live space) | `0x0` | **rejects** | `EMPTY_ADDRESS` |
| `google.com` | `0x0` | **rejects** | `EMPTY_ADDRESS` |

Two things make this worse than the gateway-503 class in the PR body:

1. **It never heals.** These are deterministic on-chain reverts, not transient gateway failures — the spaces stay broken until the code changes.
2. **The lenient path you built for DNS domains never runs.** `getEnsSpaceController` reads the text record *before* the owner, so the throw happens in `getEnsTextRecord` and `isBareResolverRevert`/`getEnsAddressStrict` is unreachable for exactly this call. The commit's own rationale ("how unclaimed DNS domains answer an addr() read") holds for `text()` reads too — that premise is what broke.

The same line also locks out any `.eth` space whose resolver predates EIP-634 and bare-reverts on `text()` (mechanism confirmed via state-override probe — the UR direct-calls `text()` even when `supportsInterface` says unsupported). A sweep of 1,168 `.eth` spaces found zero such resolvers today, so that class is theoretical for now, but worth a decision: restore-and-scope, or accept and document.

Impact direction is deny → error, never deny → wrong owner, so this is availability, not an authorization hole — but 2 known live spaces break.

**Suggested fix:** key the bare-revert classification on the *name class* instead of the read path — `ResolverError('0x')` is a legitimate no-record answer for `other-tld` names (keep it lenient there on both read paths), while `.eth`/subdomain names keep the strict fail-closed behavior this commit correctly introduced. `getDomainType` already exists to make that split. That closes the fail-open hole for the names where owner-vs-controller confusion is possible, without regressing the DNS class.

Whichever way this lands, the PR body needs a matching update — it still says "empty … resolver reverts … resolve as before" and "one shared revert classifier", and its "Spaces affected today" list omits this permanent class (it's also the list the 0.17.0 publish-hold decision is based on).

wa0x6e (changes requested):
2dd5867's record-path tightening regresses live DNS-domain spaces — details and a fix direction in the inline comment.

wa0x6e (approved):
tAck

PR #1231, opened 2026-08-14
refactor(ens): port getEnsOwner to viem

ChaituVR:
## What

`getEnsOwner` makes the same lookups through the viem client instead of ethers:

- registry `owner()` and NameWrapper `ownerOf()` via `readContract`
- the DNS-import and offchain-subdomain fallbacks via `getEnsAddress` (a CCIP gateway 404 maps to not-found, replacing the old error-message string matching)

ethers `Contract`/`getProvider` are now fully out of the ENS module. One file, +41/−29.

## Why

Groundwork for ENSv2 `findOwner` support (#1230 rebases onto this and shrinks to ~50 lines of pure feature). Owner lookups also gain the CCIP timeout guard from #1227, which ethers' internal CCIP fetch never had.

## No behavior change

All 441 live tests pass **unchanged** — registry owners, wrapped names, `cb.id`/`uni.eth` offchain subdomains, DNS-imported `defi.app`, invalid names, unsupported networks. Only edge shift: a gateway failing with a non-404 error now throws instead of depending on its error text — fail-closed.

## How to test

`yarn typecheck && yarn lint && yarn build && yarn test:once` → 441/441 (28 files), zero test modifications needed.

### DIFF
--- a/src/utils/ens.ts
+++ b/src/utils/ens.ts
@@ -1,22 +1,21 @@
-import { Contract } from '@ethersproject/contracts';
 import { getAddress } from '@ethersproject/address';
 import { ensNormalize } from '@ethersproject/hash';
+import { parseAbi, HttpRequestError } from 'viem';
 import { namehash } from 'viem/ens';
-import getProvider from './provider';
+import type { Address, Hex } from 'viem';
 import { getViemClient } from './viem';
 import { fetch } from '../utils';
 import networks from '../networks.json';
 
 type DomainType = 'ens' | 'tld' | 'other-tld' | 'subdomain';
 
-const MUTED_ERRORS = [
-  // mute error from coinbase, when the subdomain is not found
-  // most other resolvers just return an empty address
-  'response not found during CCIP fetch',
-  // mute error from missing offchain resolver (mostly for sepolia)
-  'UNSUPPORTED_OPERATION'
-];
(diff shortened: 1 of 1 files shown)


tony8713 (approved):
Ran the same corpus I used on the first pass, 437 real non-`.eth` space ids, five passes, alternating implementation order. The previous head returned a silently wrong zero address for 11 live spaces; this one returns zero silently for none of them and throws instead. `calderagov.calderans.eth`, whose gateway is permanently 503ing, now throws 5/5 and matches master.

I checked the classification directly rather than trusting the shape of it. viem hands the gateway status back as a number, so the `args?.[0] === 404` equality genuinely works, and under 150 lookups of induced rate limiting a 429 threw every time and was never read as not-found. The `other-tld` branch is now better than master, which was quietly returning a zero address when the resolver step failed.

One real difference from master remains. viem's `ccipRequest` does a single fetch where ethers retried a 429 up to twelve times, so 9 of my 2185 lookups surfaced as errors that master rode through. That is 0.41%, it fails closed, and a retry inside `ccipReadWithDeadline` would close it and cover `getEnsTextRecord` too. Not worth holding this.

Worth refreshing the description before merge, it still says "No behavior change" and describes the old 404-mapping approach. The behaviour did change, for the better.

PR #1234, opened 2026-08-18
feat(provider): type getProvider's return per chain family

### LINKED ISSUE #1233: Type getProvider's return per chain family instead of any
## Problem

`utils.getProvider` returns `any`, by deliberate choice:

```ts
// return loose `any` type to avoid typecheck issues on package consumers
function getProvider(network, options = {}) { ... }
```

This makes every call site unchecked. Consumers that care have to re-assert the type themselves, and the assertion is unverified — assigning `any` to a `StaticJsonRpcProvider` local succeeds even if the network is actually Starknet and the value is a starknet.js `RpcProvider`.

## Proposal

The runtime already discriminates exactly: `getProviderType` returns `'starknet'` when `networks[network].starknet` is true, which today is exactly the keys `'0x534e5f4d41494e'` and `'0x534e5f5345504f4c4941'`. That knowledge can be expressed as overloads, keeping the implementation unchanged:

```ts
type StarknetNetworkId = '0x534e5f4d41494e' | '0x534e5f5345504f4c4941';

export default function getProvider(network: StarknetNetworkId, options?: ProviderOptions): RpcProvider;
export default function getProvider(network: string | number, options?: ProviderOptions): StaticJsonRpcProvider;
export default function getProvider(network: string | number, options?: ProviderOptions): any { ... }
```

Call sites passing a literal network id (the common case) get the right concrete type; widened `string` still resolves to the EVM overload, which matches today's dominant usage.

In stamp we shipped this shape as a local wrapper (discriminating on the `0x` prefix rather than exact keys) and it type-checks all nine call sites correctly, including catching starknet.js method calls (`getStarkProfile`, `getAddressFromStarkName`) against the real `RpcProvider` type — see snapshot-labs/stamp#554. Upstream is the right home: the list of Starknet keys lives in networks.json here, and every consumer would benefit instead of each one re-annotating.

## Considerations

- **Consumer typecheck breaks:** moving off `any` can surface new tsc errors in consumers — that is the point, but it deserves at least a minor version bump and a changelog note.
- **`starknet` is a peer dependency:** the `.d.ts` would reference starknet's `RpcProvider`, so consumers need starknet types resolvable. An `import type` keeps it type-only; consumers without the peer installed may need `skipLibCheck` or the overload could live behind a type-only export.
- **Version skew:** consumers on a different starknet major could see structural mismatch noise on the Starknet overload.

If the hard typing is considered too breaking, a middle ground is exporting the overloaded signature as a separate named export (or a `.d.ts`-level `getTypedProvider`) and leaving the default `any` untouched.

tony8713:
Closes #1233.

`utils.getProvider` returned `any`, so every call site was unchecked, and a consumer that wanted a real type had to assert one itself with nothing verifying the assertion.

The runtime already discriminates exactly, on the `starknet` flag in `networks.json`. This puts that in the signature: a network the compiler can pin to a Starknet id resolves to starknet.js's `RpcProvider`, everything else resolves to `StaticJsonRpcProvider`. The implementation does not change.

A network the compiler cannot pin, whether a widened `string`/`number` or an `any`, resolves to `StaticJsonRpcProvider`. That case carries most of the weight here. Consumers set `noImplicitAny: false` nearly everywhere, so an untyped network argument is ordinary rather than exotic, and resolving those to the Starknet provider would trade an honest `any` for a confident wrong answer. It is the behaviour the type has to get right to be safe to adopt, so it is most of what the type test covers.

`src/verify/starknet.ts` is the first thing the narrowing caught: it took a widened `string` and called `getClassAt` and `verifyMessageInStarknet` on what the compiler now reads as an ethers provider. Its parameter is typed to the Starknet ids. The exported `verify` already casts at dispatch, so its signature is unchanged.

The Starknet id list is a literal union rather than something derived from `networks.json`, because deriving it puts a JSON import into the emitted `.d.ts` and consumers would then need `resolveJsonModule` to read our types. A test pins the union to `networks.json` instead, since that file is updated by a workflow rather than by hand.

A consumer that wraps `getProvider` behind its own `string | number` signature collapses both branches onto the EVM one. Making that wrapper generic is enough to restore it, and nothing needs to be imported from here. Inside a generic wrapper the return is still an unresolved conditional, so it propagates to callers but cannot be dereferenced in the wrapper itself without first widening it to `StaticJsonRpcProvider | RpcProvider`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "dev": "rollup -c -w",
     "lint": "eslint . --ext .ts",
     "lint:fix": "yarn lint --fix",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit -p tsconfig.typecheck.json",
     "prepublishOnly": "yarn build"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@snapshot-labs/snapshot.js",
-  "version": "0.16.0",
+  "version": "0.17.0",
   "repository": "snapshot-labs/snapshot.js",
   "license": "MIT",
   "main": "dist/snapshot.cjs.js",
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -12,6 +12,13 @@ type ProviderInstance = StaticJsonRpcProvider | RpcProvider;
 
 export type ProviderType = 'evm' | 'starknet';
 
+export const STARKNET_NETWORK_IDS = [
+  '0x534e5f4d41494e',
+  '0x534e5f5345504f4c4941'
+] as const;
+
+export type StarknetNetworkId = typeof STARKNET_NETWORK_IDS[number];
+
 const DEFAULT_BROVIDER_URL = 'https://rpc.snapshot.org' as const;
 export const DEFAULT_TIMEOUT = 25000 as const;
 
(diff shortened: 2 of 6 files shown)


wa0x6e (approved):
utAck

PR #1240, opened 2026-08-20
feat(provider): add RPC client attribution

### LINKED ISSUE #1238: feat(provider): clientName option for RPC client attribution
## Context

sx-monorepo#2106 added `?client=<app>` attribution to RPC traffic, but named a known gap: everything that builds its provider **inside** snapshot.js goes out unattributed, because `getProvider` hardcodes `${broviderUrl}/${networkId}` and rejects a query string smuggled through `broviderUrl`. Affected today:

- sequencer's starknet networks (`apps/sequencer/src/helpers/provider.ts:23` delegates back to `snapshot.utils.getProvider` for the `sn`/`sn-sep` broviderId mapping);
- `snapshot.utils.getSpaceController` / `getSpaceUri` (sequencer settings/poke path);
- every other snapshot.js consumer (score-api strategies, stamp, sidekick, blockfinder) — none can attribute without this.

sequencer had to fork a local EVM provider helper just to append the param; that fork can be deleted once this lands.

## Change

`clientName` option on `ProviderOptions`, appended as `?client=<value>` to the provider URL for both EVM and starknet factories:

- `src/utils/provider.ts` — add to `normalizeOptions` (default: none), include in `createMemoKey` (`${networkId}:${broviderUrl}:${timeout}:${clientName}` — otherwise two callers with different clients share one provider and the label lies), build the URL with the query in `getEvmProvider`/`getStarknetProvider`.
- `src/utils/viem.ts` — same option, same memo-key rule.
- `getSpaceController` / `getSpaceUri` / other helpers that create providers internally: accept and forward `clientName` through their existing options params.
- Validate the value (`/^[a-z0-9.-]{1,32}$/i`) — it lands in a URL and in brovider's metric label.

Non-breaking: absent option = today's URL exactly.

## Follow-ups unblocked (not this issue)

- sequencer: pass `clientName: 'sequencer'` everywhere, delete the local provider fork.
- score-api / stamp / sidekick / blockfinder: one-line attribution each.
- brovider#554 reads the param into `rpc_request_count{client}`; traffic still showing `client="other"` after rollout = remaining unattributed callers.

## Acceptance

- `getProvider(1, { clientName: 'sequencer' })` → provider URL ends `/1?client=sequencer`; starknet network → same on its nodeUrl.
- Two `getProvider(1, …)` calls with different `clientName` → two provider instances; same name → memoized single instance.
- No `clientName` → URL unchanged (byte-identical), existing memo keys unaffected.
- `getSpaceController(space, network, { clientName })` traffic carries the param.

tony8713:
Closes #1238.

Adds a validated `clientName` provider option and applies it consistently to ethers, Starknet, and viem RPC URLs. Provider memoization separates attributed clients while preserving existing URLs and memoization for callers that omit the option.

Existing space-resolution option forwarding carries the attribution through helpers that create viem clients internally.

### DIFF
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -6,7 +6,11 @@ import networks from '../networks.json';
 export interface ProviderOptions {
   readonly broviderUrl?: string;
   readonly timeout?: number;
+  readonly clientName?: string;
 }
+
+type NormalizedProviderOptions = Required<Omit<ProviderOptions, 'clientName'>> &
+  Pick<ProviderOptions, 'clientName'>;
 
 type ProviderInstance = StaticJsonRpcProvider | RpcProvider;
 
@@ -20,13 +24,14 @@ export const STARKNET_NETWORK_IDS = [
 export type StarknetNetworkId = typeof STARKNET_NETWORK_IDS[number];
 
 const DEFAULT_BROVIDER_URL = 'https://rpc.snapshot.org' as const;
+const CLIENT_NAME_PATTERN = /^[a-z0-9.-]{1,32}$/i;
 export const DEFAULT_TIMEOUT = 25000 as const;
 
 const providerMemo = new Map<string, ProviderInstance>();
 
 const providerFnMap: Record<
   ProviderType,
-  (networkId: string, options: Required<ProviderOptions>) => ProviderInstance
+  (networkId: string, options: NormalizedProviderOptions) => ProviderInstance
 > = {
   evm: getEvmProvider,
   starknet: getStarknetProvider
@@ -34,11 +39,19 @@ const providerFnMap: Record<
 
 export function normalizeOptions(
   options: ProviderOptions = {}
-): Required<ProviderOptions> {
-  const { timeout } = options;
+): NormalizedProviderOptions {
+  const { timeout, clientName } = options;
+
+  if (
+    clientName !== undefined &&
+    (typeof clientName !== 'string' || !CLIENT_NAME_PATTERN.test(clientName))
+  ) {
+    throw new Error('Invalid clientName');
+  }
 
   return {
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
+    clientName,
     // Not `??`: it passes `NaN` through (`Number()` of an unset env var), and
     // `setTimeout(..., NaN)` fires on the next tick, aborting every request.
     timeout:
@@ -64,9 +77,25 @@ export function getProviderType(network: string | number): ProviderType {
 
 export function createMemoKey(
   networkId: string,
-  options: Required<ProviderOptions>
+  options: NormalizedProviderOptions
 ): string {
-  return `${networkId}:${options.broviderUrl}:${options.timeout}`;
+  const key = `${networkId}:${options.broviderUrl}:${options.timeout}`;
+  return options.clientName
+    ? JSON.stringify([
+        networkId,
+        options.broviderUrl,
+        options.timeout,
+        options.clientName
+      ])
+    : key;
+}
+
+export function createProviderUrl(
+  networkId: string,
+  options: NormalizedProviderOptions
+): string {
+  const url = `${options.broviderUrl}/${networkId}`;
+  return options.clientName ? `${url}?client=${options.clientName}` : url;
 }
 
 type IsAny<T> = 0 extends 1 & T ? true : false;
@@ -103,11 +132,11 @@ export default function getProvider(
 
 function getEvmProvider(
   networkId: string,
-  options: Required<ProviderOptions>
+  options: NormalizedProviderOptions
 ): StaticJsonRpcProvider {
   return new StaticJsonRpcProvider(
     {
-      url: `${options.broviderUrl}/${networkId}`,
+      url: createProviderUrl(networkId, options),
       timeout: options.timeout,
       allowGzip: true
     },
@@ -125,10 +154,10 @@ const starknetTimeoutError = (timeout: number) =>
 
 function getStarknetProvider(
   networkKey: string,
-  options: Required<ProviderOptions>
+  options: NormalizedProviderOptions
 ): RpcProvider {
   return new RpcProvider({
-    nodeUrl: `${options.broviderUrl}/${networkKey}`,
+    nodeUrl: createProviderUrl(networkKey, options),
     // At 0 there is no deadline to add, so leave starknet its own transport
     // (`baseFetch ?? ponyfill`) rather than swap the transport for nothing.
     // Native `fetch`, bound like starknet's own browser default: the package
--- a/src/utils/provider.ts
+++ b/src/utils/provider.ts
@@ -40,18 +40,20 @@ const providerFnMap: Record<
 export function normalizeOptions(
   options: ProviderOptions = {}
 ): NormalizedProviderOptions {
-  const { timeout, clientName } = options;
+  const { timeout } = options;
+  let { clientName } = options;
 
   if (
     clientName !== undefined &&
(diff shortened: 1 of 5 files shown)


src/utils/provider.ts
```diff
@@ -64,9 +77,25 @@ export function getProviderType(network: string | number): ProviderType {
...
+): string {
+  const key = `${networkId}:${options.broviderUrl}:${options.timeout}`;
+  return options.clientName
+    ? JSON.stringify([
+        networkId,
+        options.broviderUrl,
+        options.timeout,
+        options.clientName
+      ])
+    : key;
```
wa0x6e: The two-encoding branch exists only to keep the no-`clientName` key byte-identical, but the key never leaves the module-private `providerMemo`/`viemClientMemo` Maps. A fixed 4-field key is unambiguous on its own: `clientName` is validated colon-free and `timeout` is a number, so the URL is always the middle segment. The collision the `should not collide with legacy memo keys` test guards against came from the *variable* field count, not the delimiter.

```suggestion
  return `${networkId}:${options.broviderUrl}:${options.timeout}:${options.clientName ?? ''}`;
```

Then `preserves the memo key when no client name is set` (providers.spec.ts:485) either asserts `'1:https://rpc.snapshot.org:25000:'` or goes away — it pins an internal encoding. The collision test can stay.

src/utils/provider.ts
```diff
@@ -20,25 +24,34 @@ export const STARKNET_NETWORK_IDS = [
...
-  const { timeout } = options;
+): NormalizedProviderOptions {
+  const { timeout, clientName } = options;
+
+  if (
+    clientName !== undefined &&
+    (typeof clientName !== 'string' || !CLIENT_NAME_PATTERN.test(clientName))
+  ) {
+    throw new Error('Invalid clientName');
+  }
```
wa0x6e: An invalid `clientName` shouldn't throw — just warn and drop it. As written the throw is inconsistent across `getSpaceController` branches: on the ENS path it propagates (client is built before the try, deliberately), but on the UD/Sonic path `getProvider(network, options)` sits inside `getUDNameOwner`'s catch-all and becomes `EMPTY_ADDRESS`, i.e. "this space has no controller" — and the new Sonic test in `ens.spec.ts` pins that. Warning and falling back to an unattributed URL matches the `timeout` policy a few lines below and removes the inconsistency at the source.

```suggestion
  const { timeout } = options;
  let { clientName } = options;

  if (
    clientName !== undefined &&
    (typeof clientName !== 'string' || !CLIENT_NAME_PATTERN.test(clientName))
  ) {
    console.warn(`Ignoring invalid clientName: ${String(clientName)}`);
    clientName = undefined;
  }
```

The `rejects client names with …` cases in `providers.spec.ts` then assert `normalizeOptions({ clientName }).clientName` is `undefined` (plus a `console.warn` spy) instead of `toThrow`.

src/utils/provider.ts
```diff
@@ -20,25 +24,34 @@ export const STARKNET_NETWORK_IDS = [
...
 
   return {
     broviderUrl: options.broviderUrl || DEFAULT_BROVIDER_URL,
+    clientName,
```
wa0x6e: The pattern is `/i` but the value is used as-is, so `Sequencer` and `sequencer` get separate URLs, separate memo entries, and eventually separate `client` labels in brovider — the opposite of what attribution is for. Fold once here.

```suggestion
    clientName: clientName?.toLowerCase(),
```

wa0x6e (changes requested):
Requesting changes per the three suggestions above on `src/utils/provider.ts` (single memo-key format, warn-and-drop invalid `clientName`, lowercase it).

src/utils/provider.ts
tony8713: Updated `createMemoKey` to the fixed four-field format and adjusted the no-client key expectation.

src/utils/provider.ts
tony8713: Invalid names now warn and normalize to `undefined`; the table asserts both the fallback and the warning.

src/utils/provider.ts
tony8713: `normalizeOptions` now lowercases once before URL and memo-key construction. Mixed-case provider and viem cases assert lowercase URLs and shared instances.

src/utils/provider.ts
tony8713: Dropped the value from the message: interpolating it is what threw for a non-coercible object (`Object.create(null)` and friends), which put the ENS branch back to rejecting while Sonic swallowed it into `EMPTY_ADDRESS`. The message still names the option, so it says which one was dropped, just not what it was.

src/utils/provider.ts
```diff
@@ -20,25 +24,36 @@ export const STARKNET_NETWORK_IDS = [
...
-): Required<ProviderOptions> {
+): NormalizedProviderOptions {
   const { timeout } = options;
+  let { clientName } = options;
+
+  if (
+    clientName !== undefined &&
+    (typeof clientName !== 'string' || !CLIENT_NAME_PATTERN.test(clientName))
+  ) {
+    console.warn('Ignoring invalid clientName');
```
wa0x6e: Follow-up on #discussion_r3823129289 — the fix is right, dropping the interpolation was the correct call: `String(Object.create(null))` throws `TypeError: Cannot convert object to primitive value`, so my suggestion would have put the ENS/Sonic split back for exactly the values it was meant to remove.

But the value can still be reported, because `console.warn`'s second argument is never coerced — it goes to `util.inspect` on Node and to the console formatter in the browser, neither of which calls `ToPrimitive`:

```
console.warn('x', Object.create(null))  ->  x [Object: null prototype] {}   (no throw)
console.warn('x', Symbol('y'))          ->  x Symbol(y)                     (no throw)
```

As written the warning is unconditionally value-free, so the case it exists for loses its diagnostic: a service ships `CLIENT_NAME=snapshot_sequencer` (underscore, rejected), and the operator gets `Ignoring invalid clientName` with no way to tell which value was dropped or which of several call sites dropped it.

```suggestion
    console.warn('Ignoring invalid clientName:', clientName);
```

Second argument, not interpolation — interpolation is what threw.

The table case can then assert `toHaveBeenCalledWith('Ignoring invalid clientName:', clientName)`, which also pins that the value reaches the log without coercion.

---

Separately, and this one is on my original ask rather than the fix: `normalizeOptions` runs *before* the memo lookup in both `getProvider` (provider.ts:114-117) and `getViemClient` (viem.ts:39-42), so a misconfigured `clientName` warns on every call, memo hits included — unbounded on a per-request path in score-api/sequencer, where the `timeout` fallback four lines below is silent. I lean towards leaving it: the value is set once at service startup, so it is a fixed-rate log line rather than a leak. Noting it so it is a decision and not an oversight; no change requested.

tony8713: Confirmed independently, including your correction of my own earlier fix:

```
console.warn("x", Object.create(null))  -> x [Object: null prototype] {}   (no throw)
console.warn("x", Symbol("y"))          -> x Symbol(y)                     (no throw)
"interp: " + Object.create(null)        -> TypeError: Cannot convert object to primitive value
```

Second argument, not interpolation. Updated the call site and the test assertion to `toHaveBeenCalledWith("Ignoring invalid clientName:", clientName)`. Mutation-proved: reverting to the bare-string warn reddens all 10 cases in the `ignores client names with ...` table (including the `null` and `Object.create(null)` cases), restoring it goes green.

On the second point (warn firing before the memo lookup, so a bad `clientName` re-warns on every memoized call): agreed this is a decision, not an oversight, for the same reason you gave, it is set once at service startup rather than per-request input. Leaving as is.

Pushed in dc715e7.

wa0x6e (approved):
tAck

PR #1245, opened 2026-09-17
fix(provider): pin Starknet requests to the latest block

### LINKED ISSUE #1244: fix(provider): Starknet contract calls fail with "unknown block tag 'pending'"
Every Starknet contract read made through a provider from `getProvider()` fails against brovider. Any consumer that calls `callContract` — directly, or via starknet.js helpers like `getStarkProfile` / `getAddressFromStarkName` — without naming a block gets `-32602 Invalid Params: "unknown block tag 'pending'"` instead of a result. In stamp this means no avatar resolves for any Starknet address or `.stark` domain ([STAMP-7Z](https://snapshot-labs.sentry.io/issues/STAMP-7Z), first seen 2026-09-17 16:32 UTC).

`multicall()` is unaffected, so name and address lookups still work; only the direct-call paths are broken.

## Reproduction

1. Read any Starknet mainnet contract through the provider without passing a block:

```ts
import getProvider from './src/utils/provider';
import { CallData, starknetId } from 'starknet';

const provider = getProvider('0x534e5f4d41494e');

await provider.callContract({
  contractAddress: starknetId.getStarknetIdContract('0x534e5f4d41494e'),
  entrypoint: 'address_to_domain',
  calldata: CallData.compile({
    address:
      '0x3c4dd9305d6b3a2c1ba7ed06c2285125751600c821ec45279e3760db341d67e',
    hint: []
  })
});
```

Expected: the call result (`['0x0']` for that address).
Actual: `RpcError: RPC: starknet_call ... -32602: Invalid Params: "unknown block tag 'pending'"`.

## Cause

`src/utils/provider.ts:155` builds the `RpcProvider` without a `blockIdentifier`, so starknet.js applies its channel default of `pending` to every call that does not pass one (`starknet` 6.21's `defaultOptions.blockIdentifier`).

brovider's Starknet mainnet endpoint now answers `starknet_specVersion` with `0.10.2`. RPC 0.9 removed the `pending` block tag in favour of `pre_confirmed`, so the node rejects the tag outright — same call, only the tag differs:

```
$ curl -s https://rpc.brovider.xyz/sn -d '{...,"method":"starknet_call","params":{"request":{...},"block_id":"pending"}}'
{"error":{"code":-32602,"message":"Invalid Params","data":"unknown block tag 'pending'"}}

$ curl -s https://rpc.brovider.xyz/sn -d '{...,"block_id":"latest"}'
{"result":["0x0"]}
```

Not a regression: `getStarknetProvider` has never named a block, since Starknet support landed in [2f95e9b](https://github.com/snapshot-labs/snapshot.js/commit/2f95e9b404c0971d63a413afab9018559548a34e). It worked only while the node still accepted `pending` — last observed succeeding 2026-08-24. `src/multicall/starknet.ts:206` already falls back to `latest`, which is why `multicall()` keeps working.

## Fix

```diff
   return new RpcProvider({
     nodeUrl: createProviderUrl(networkKey, options),
+    blockIdentifier: 'latest',
```

`latest` rather than `pre_confirmed`: it is valid on every spec version, and no consumer needs unconfirmed state.

wa0x6e:
Closes #1244

Every Starknet request made through `getProvider()` that does not name a block fails against brovider with `-32602 Invalid Params: "unknown block tag 'pending'"`. starknet.js defaults those requests to `pending`; RPC 0.9 dropped that tag and the brovider Starknet endpoints now report spec `0.10.2`, so the node rejects it outright.

Impact is wider than contract reads. Starknet signature verification is broken too: it calls `getClassAt()` first, which is block-defaulted the same way, and the RPC error does not match the `Contract not found` remap, so callers get a raw `RpcError`. `multicall()` was never affected — it already passes an explicit tag.

`latest` rather than `pre_confirmed`: it is valid on every spec version, starknet.js 6.24.1 rejects `pre_confirmed` client-side, and nothing here needs unconfirmed state.

## Not fixed here

The same node upgrade broke a second, unrelated path, which this PR deliberately leaves alone.

Accounts that *revert* on a bad signature (Braavos `INVALID_SIG`, OpenZeppelin, Argent on a malformed signature) used to be reported as `false`. starknet.js recognises them by substring-matching the revert reason against a hardcoded list, and the node now returns that reason as a raw felt — `0x617267656e742f696e76616c69642d7369676e61747572652d666f726d6174` rather than `argent/invalid-signature-format`. The match therefore never fires and `verify()` throws instead of resolving `false`. Accounts that return `0x0` instead of reverting are unaffected, which is why most invalid-signature cases still behave.

This needs its own fix and its own issue; it is visible here only because the block-tag fix lets those calls reach the contract at all.

## CI

Two unrelated failures. The Sepolia ENS e2e cases in `test/e2e/utils.spec.js` broke with ENS's 2026-09-15 Sepolia redeploy (#1246) and are fixed by #1250. The one Starknet invalid-signature case is the reverting-account behaviour #1248 fixes on top of this branch. The merged tree of all three passes both specs locally.


tony8713 (approved):
LGTM on the provider fix. The new regression test fails when the latest default is removed, and explicit block selections are preserved.

One small cleanup: drop the explanatory comment above `blockIdentifier` in `src/utils/provider.ts`; the test already protects this setting.

CI remains red on eight existing Sepolia ENS cases and the separately documented Argent invalid-signature mapping, so this is approval of the block-tag fix, not a clean-suite result.

ChaituVR (approved):
utACK

Issue #1246, opened 2026-09-17
fix(ens): getEnsOwner throws for every Sepolia name


wa0x6e:
`getEnsOwner` and `getSpaceController` throw for every Sepolia name instead of returning an owner, so any consumer checking who controls a testnet name fails outright rather than resolving it. Mainnet is unaffected.

No code change caused it — the contract the package calls stopped implementing the function.

## Reproduction

1. `yarn test:once test/e2e/utils.spec.js`

Expected: the Sepolia cases pass.
Actual: every Sepolia `getEnsOwner` / `getSpaceController` case rejects — the underlying `findOwner(bytes)` call reverts with empty data:

```
$ curl -s https://rpc.snapshot.box/11155111 -H 'content-type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[{"to":"0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe","data":"0x97ad3b3b…"},"latest"]}'
{"error":{"code":3,"message":"execution reverted","data":"0x"}}
```

(`0x97ad3b3b` is `findOwner(bytes)`, called with the DNS encoding of `test123.eth`.)

## Cause

`src/utils/ens.ts:191` calls `findOwner` on `networks[11155111].ensUniversalResolver` for that chain with no catch — deliberately, so a revert cannot silently fall back to a stale v1 owner.

The address is not stale. `0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe` is the `UpgradableUniversalResolverProxy`, still ENS's canonical Sepolia entrypoint; it forwards to `ManagedUniversalResolverProxy` `0x6d80F2172CFdEc5730fE683860C33d26fC42e6F1`, whose implementation slot currently holds `0x5d25c1d6acbb71b7a28aa7899618a3412a8303e3` — the `UniversalResolverV2` of ENS's official 2026-09-15 deployment (`ensdomains/contracts-v2@366de741`, per `ensdomains/bigname` `docs/sepolia-deployment.md`).

[...]

PR #1248, opened 2026-09-17
fix(verify): report a reverting Starknet account as invalid, not an error

### LINKED ISSUE #1247: fix(verify): invalid Starknet signatures throw instead of being reported invalid
An invalid Starknet signature is no longer reported as invalid. For accounts that reject a bad signature by reverting, verification now fails with an RPC error instead of answering "not valid", so a caller cannot tell a rejected signature apart from an outage and surfaces a server error where it should surface a plain rejection.

Accounts that answer a bad signature with `0x0` rather than reverting are unaffected, which is why most invalid signatures still behave correctly.

## Reproduction

Requires #1245 (or any build that stops sending the `pending` block tag) — without it the call dies earlier, on the block tag, and never reaches the contract.

1. Verify a Starknet message with a malformed signature:

```ts
import verify from './src/verify/starknet';
import message from './test/fixtures/starknet/message-alias.json';

await verify(message.address, ['1'], message.data, '0x534e5f4d41494e');
```

Expected: resolves `false`.
Actual: rejects with `Error: Signature verification Error: RpcError: RPC: starknet_call ... 40: Contract error: {"revert_error":"0x617267656e742f696e76616c69642d7369676e61747572652d666f726d6174"}`.

The existing case `should return false when the signature is not valid` in `src/verify/starknet.spec.ts:128` covers this and is red.

## Cause

`src/verify/starknet.ts:34` delegates to starknet.js `verifyMessageInStarknet`, which recognises a rejected signature by substring-matching the revert reason against a hardcoded list (`argent/invalid-signature`, `is invalid, with respect to the public key`, `INVALID_SIG`) and returns `false` on a match, rethrowing everything else.

The brovider Starknet endpoints now report spec `0.10.2`, and return the revert reason as a raw felt rather than text:

```
$ curl -s https://rpc.snapshot.org/sn -d '{...,"method":"starknet_call",...}'
{"error":{"code":40,"message":"Contract error","data":{"revert_error":"0x617267656e742f696e76616c69642d7369676e61747572652d666f726d6174"}}}
```

That felt decodes to `argent/invalid-signature-format`, which would have matched, but the comparison runs against the hex, so the match never fires and the error is rethrown. starknet.js also flattens the error into a plain `Error` before it reaches this repo, so only the message string survives — the structured `revert_error` is not available to the caller.

Not a code regression: nothing in this repo changed. The same code and the same starknet.js version passed all 26 cases of that spec in CI on 2026-09-14 ([run 34823763596](https://github.com/snapshot-labs/snapshot.js/actions/runs/34823763596)); the node changed underneath. Measured on an Argent account; the list above implies Braavos and OpenZeppelin accounts are affected the same way.

## Fix

Decode the felts in the rethrown message and re-run the known-marker check before giving up:

```ts
const reverts = (message.match(/0x[0-9a-fA-F]{2,}/g) ?? []).map((felt) =>
  shortString.decodeShortString(felt)
);

if (reverts.some((reason) => INVALID_SIGNATURE_REVERTS.some((m) => reason.includes(m)))) {
  return false;
}
```

wa0x6e:
Closes #1247

Stacked on #1245 — without its block-tag fix these calls never reach the contract. Base is `fix/1244`; retarget to `master` once that merges.

An invalid Starknet signature on an account that rejects by reverting (Argent, Braavos) surfaced an RPC error instead of `false`, so a caller could not tell a rejected signature apart from a node outage.

starknet.js spots those accounts by substring-matching the revert reason against its own list of account error strings; RPC 0.9 nodes return that reason as a raw felt rather than text, so the match never fires. Matching the encoded form of the markers restores the rejection. The verification call also needed awaiting — returning its promise from inside the `try` let the rejection settle outside the `catch`, which is why no error handling there applied to it.

Accounts answering `0x0` instead of reverting were never affected, which is why most invalid signatures already behaved.

## CI

The only failing file is `test/e2e/utils.spec.js`, the Sepolia ENS cases that broke with ENS's 2026-09-15 Sepolia redeploy (#1246); #1250 fixes those. Starknet passes on this branch. The merged tree with #1250 passes both specs locally, so CI is green once both land, in either order.


tony8713 (changes requested):
At [starknet.ts:53-55](https://github.com/snapshot-labs/snapshot.js/blob/a36b520a9cfc5554d4a5ec5aa07af091fb875620/src/verify/starknet.ts#L53-L55), matching the full error message can turn a genuine RPC failure into `false`. Starknet.js embeds request calldata in that message: public `verify()` with `sig = ["0x494e56414c49445f534947"]` resolves `false` on `{code: 429, message: "Too many requests"}`, while `["1"]` correctly rejects the same error. Please match contract-revert response data rather than the full formatted message, and add a regression covering a marker in calldata.

Non-blocking cleanup: drop the added explanatory comments at lines 6-8 and 42, keeping `return await`.

tony8713 (approved):
The calldata/RPC-error issue is fixed, and the new regression fails when request-dump exclusion is removed. LGTM.

Non-blocking: please drop the four added explanatory comment blocks.

tony8713 (approved):
Re-reviewed at 4ba55ce. The calldata-matching issue is fixed: the marker match now runs only on the segment after the request dump, and the regression fails when that exclusion is removed. LGTM.

ChaituVR (approved):
utAck

PR #1250, opened 2026-09-17
fix(ens): read ENSv2 owners from the UniversalHelper

### LINKED ISSUE #1246: fix(ens): getEnsOwner throws for every Sepolia name

wa0x6e:
Closes #1246

`getEnsOwner` and `getSpaceController` threw for every Sepolia name, so nothing could check who controls a testnet name.

ENS's 2026-09-15 Sepolia deployment moved the registry lookups off the Universal Resolver into a separate `UniversalHelper` contract and renamed `findOwner` to `findExactOwner`; the call the package still made hit a contract that no longer implements it. The resolver address itself was never stale — the canonical proxy still fronts the current implementation, and the text-record and address paths that share it kept working.

That helper sits behind no proxy, and ENS has redeployed the Sepolia set repeatedly. A pinned address therefore outlives its deployment and keeps answering — from the root registry ENS has abandoned, which is worse than reverting, since a wrong owner authorizes the wrong space controller. The helper's root registry is now checked against the one the canonical Universal Resolver resolves through, and a mismatch throws. The issue's proposed one-line address swap is that failure mode: the implementation it pins is bound to the superseded root. The check is a deployment invariant, so it runs once per client and is re-verified every five minutes rather than on every lookup.

Mainnet stays on the v1 registry path until the ENS DAO upgrades the Universal Resolver proxy to the ENSv2 build; configuring `ensUniversalHelper` for network 1 before that throws on the first lookup.

Restoring the lookup exposed a second way to hand back an address nobody controls. ENSv1 registry entries survive the migration, so a name whose authority has moved to ENSv2 keeps an entry naming whoever held it before — `tiny.fox.eth`, whose parent is registered in v2, resolved to its v1 owner while ENSv2 reported it unowned. The v1 registry is now read only for names ENSv2 still resolves through one of its two v1 bridges: the mirror for `.eth` names, the DNS TLD resolver for imported domains. The same rule covers expiry: ENSv2 drops the mirror once the name's ENSv2 reservation expires, which happens before v1 grace ends, so such a name reads as unowned, which is what ENSv2 itself reports for it, rather than as its former holder. Networks with no ENSv2 helper are unaffected, so mainnet behaviour is unchanged.

The mirror rule depends on two more pinned resolver addresses. A network configured with the helper but without them would have skipped the v1 read for every name with no error; that configuration now fails on the first lookup.

`test123.eth`'s v1 registration expired past grace and the live root holds no record of it — its remaining owner exists only in the retired deployment — so the e2e expectation for it changes to the empty address.

The e2e timeout bump is unrelated housekeeping: those specs resolve names over live RPC and CCIP gateways and sit close enough to the 5s default to fail on contention alone.

## CI

The only failing file is `src/verify/starknet.spec.ts`, every case with `unknown block tag 'pending'`; #1245 fixes that, and #1248 is stacked on it. Those two fail only on the Sepolia ENS e2e cases this PR fixes. The two sides touch disjoint files, and the merged tree passes both specs locally, so CI is green once both land, in either order.


tony8713 (changes requested):
The matching-root check does not make the ENSv1 fallback authoritative (`src/utils/ens.ts:231-237`). On head `8c06d05`, at Sepolia block 11725881, `getSpaceController("tiny.fox.eth", "11155111")` returns the stale v1 owner `0x7Bc153b2a4C8a2f3428bd0da77a901b81c6dD809` even though both roots match, `findExactOwner` returns zero, and the canonical native-v2 resolver returns no address. Gate the fallback on the current resolution path actually delegating to v1, including on-chain imported DNS, otherwise the stale address becomes the controller used by settings and delete-space authorization.

Separately, please remove the added explanatory comments at `src/utils/ens.spec.ts:156`, `src/utils/ens.ts:192-198`, `test/e2e/utils.spec.js:178-179`, and `vite.config.js:5-6`; comment-gate flags all four, and their information is already in the code, named tests, or PR body.

ENS unit tests (30) and live utility tests (39) pass, and the changed tests fail under relevant mutations. The 17 CI failures reproduce the existing Starknet RPC rejection of the pending block tag, not an ENS regression. No files were changed.

wa0x6e:
Both points verified independently at `8c06d05`, fixed in e5b14b6.

**Stale v1 fallback** — reproduced exactly at block 11725986: `getSpaceController("tiny.fox.eth", "11155111")` → `0x7Bc153b2a4C8a2f3428bd0da77a901b81c6dD809`, with both roots matching, `findExactOwner` zero and the Universal Resolver returning no address. The mechanism predates this PR (the registry read is unchanged since before efa5de1) but the fix makes it reachable, so it belongs here.

Gated on `findResolver` returning the ENSv1 mirror `0xb2BF4a9A86d29661EA93223582b9945943931e42`, measured against every name the suite asserts on Sepolia:

```
ens.eth, bob.eth, boorger.eth, demodao.eth  -> 0xb2BF4a9A...  v1 mirror     -> fallback allowed
tiny.fox.eth                                -> 0x264268534A... fox.eth's v2 -> fallback denied
test123.eth, notfounddomain.uni.eth         -> 0x0                          -> fallback denied
defi.app                                    -> 0xb0C7881...   DNSTLDResolver -> fallback denied
```

All four v1-mirrored names keep their current owners; `tiny.fox.eth` now resolves to the empty address, covered by a live e2e case and by two unit cases (mirror -> registry read, other resolver -> no read). Networks without an ENSv2 helper skip the check, so mainnet is untouched. On-chain imported DNS names on Sepolia: I could not find one imported in v1 to measure, so that branch is reasoned, not verified — if you have a name, I will check it.

**Comments** — dropped the one at `src/utils/ens.spec.ts:156` (the mock chain already shows the order) and cut the e2e one to a single line. I kept the two that carry information the code cannot: `src/utils/ens.ts:192-198` states the threat the root check defends against, without which the check reads as ceremony and gets simplified away, and the `vite.config.js` line records where `20000` came from. `ens.ts` already carried 21 comment lines of that kind at `c0a043d`, so this matches the file rather than departing from it. I could not find `comment-gate` in the repo or in my own tooling — if it is yours, its output would help.

One correction to my own evidence, since it was published in #1246: I claimed the new `ETHRegistry` had emitted no events and that nothing had migrated. That was wrong — my scan wrapped `getLogs` in a `catch` that swallowed the RPC rejecting an unfiltered 50k-block range. It returns 190 logs in the last 800 blocks, and `fox.eth` is registered in the live root. The issue body is corrected. What survives unchanged: `test123.eth` has no owner there (direct read), and the retired implementation's `ROOT_REGISTRY()` still points at the abandoned root.

Verification at e5b14b6: `src/utils/ens.spec.ts` 31/31, `test/e2e/utils.spec.js` 40/40, typecheck and lint clean. The remaining CI failures are the pre-existing Starknet `pending` block tag rejection.

tony8713 (changes requested):
The original stale-controller finding is fixed on `e5b14b6`: `tiny.fox.eth` now returns zero, while the ordinary ENSv1 and ENSv2 owners checked still resolve correctly.

**Imported DNS ownership regresses at `src/utils/ens.ts:94-96`.** `ethplay.org` is a concrete imported name: at Sepolia block 11726158 its canonical `DNSTLDResolver.getResolver()` returns `[0x8948458626811dd0c23EB25Cc74291247077cC51, false]`, and both its canonical address and ENSv1 owner are `0x8D852E6cC57A855D0D75E1e2af57C9679D555958`, but `getSpaceController("ethplay.org", "11155111")` now returns zero. The mirror-only predicate rejects that valid on-chain delegation, then the successful address lookup falls through to DNS TXT where there is no `ENS1` record, losing the legitimate controller. Recognize verified on-chain DNS-to-v1 delegation too, without allowing every DNS resolver indiscriminately; this also disproves the new comment's assumption that delegation uses only one resolver.

On comments: `comment-gate` is my local reviewer tooling, not a repository or CI check, and it detects added prose rather than proving that prose wrong. Its current output is `comment-gate: this change ADDS code comments`, at `src/utils/ens.spec.ts:249`, `src/utils/ens.ts:211,250`, `test/e2e/utils.spec.js:146,186`, and `vite.config.js:5`. I still recommend keeping the explanatory context in the PR discussion; deleting the root guard already fails both its unit and live tests.

Current-head verification: ENS unit 31/31 and live utilities 40/40 pass, and the three tests added or changed by the follow-up fail under relevant mutations. CI run 35270821167 fails only on the existing 17 Starknet pending-tag cases; decoded, undecoded, and transport errors in the new resolver read propagate through the controller caller probe. No files were edited.

tony8713 (changes requested):
`tiny.fox.eth` is fixed. One ownership path still needs handling.

**P2: Preserve DNS-to-v1 delegation (`src/utils/ens.ts:94-96`).** The mirror-only check also rejects on-chain-imported DNS names when the canonical `DNSTLDResolver` delegates to ENSv1. I reproduced this with `gregskril.com` in a read-only `eth_call` state override simulating its existing owner setting the v1 resolver to `0x8FADE66B79cC9f707aB26799354482EB93a5B7dD`: the deployed DNS resolver returns `[resolver, false]`, but `getEnsOwner` and `getSpaceController` return zero, while bypassing only this gate restores its v1 owner. Recognize the DNS resolver's actual v1-delegated path as well as the mirror, and cover that path with a regression test.

wa0x6e:
Fixed in 1968be0. Reproduced your `ethplay.org` case independently at block 11726260 before changing anything — `DNSTLDResolver.getResolver` → `[0x8948458626811dd0c23EB25Cc74291247077cC51, false]`, v1 registry owner and resolver `0x8D852E6cC57A855D0D75E1e2af57C9679D555958` / `0x8948...`, canonical address `0x8D852E6c...`, and `getSpaceController` returning zero on `e5b14b6`. Your read of the mechanism was exactly right, including the fall-through to a DNS TXT lookup that has no `ENS1` record.

The gate now accepts ENSv2's DNS TLD resolver `0xb0C788195697dB17543bF22CBC1b0E2b4A04F9b8` alongside the v1 mirror, both pinned in networks.json. I did not add the `getResolver` probe to distinguish on-chain from DNSSEC delegation, because the same measurement showed the split cuts through a case your `ethplay.org` repro does not cover: `gregskril.com` answers `[0x35b920d4..., true]` with its v1 resolver unset, and an on-chain-only predicate would have kept its v1 owner `0x179A862703a4adfb29896552DF9e307980D19285` unreachable — a second silent controller loss, of the same class, one round later. ENSv2 has not taken over either name (neither is in the v2 registry), so their v1 entry is still the authoritative one and both are now read. The `.eth` names the gate was built for are unaffected: `tiny.fox.eth` still returns zero.

On indiscriminate resolvers: the accept-list is two pinned ENS deployment addresses, not a duck-typed check. That matters because a v2 name's owner chooses its own resolver, so a predicate that trusted any resolver implementing `getResolver` would let a v2 owner unlock a v1 read naming a different address than themselves.

Regression coverage: live cases for `ethplay.org` and `gregskril.com`, plus a unit case for the DNS branch. Both live cases fail when the DNS resolver is removed from the accept-list — verified by mutation, restored afterwards.

At 1968be0: ENS unit 32/32, live utilities 42/42, typecheck and lint clean; the only CI failures remain the 17 pre-existing Starknet pending-tag cases.

One process note: this PR was opened as a draft and is now marked ready for review. I did not change that.

tony8713 (approved):
Both correctness findings are fixed on `1968be0`: `tiny.fox.eth` remains unowned, `ethplay.org` retains its imported owner, and `gregskril.com` retains its ENSv1 owner while using DNSSEC.

You are right about the DNSSEC case. The deployed DNS TLD resolver checks ENSv1 first, both pinned bridges use the same v1 registry this code reads, and a simulated `setResolver` succeeds for the recorded owner but rejects an unrelated address. My earlier on-chain-only recommendation was unnecessarily restrictive.

ENS unit 32/32 and live utilities 42/42 pass; removing DNS-bridge acceptance fails exactly the three new tests. Resolver-read failures still propagate through the controller caller. CI run 35285283165 remains red only for the existing 17 Starknet pending-tag failures.

Nonblocking cleanup remains: the PR body still describes a mirror-only gate rather than both bridges, and the new fixture-provenance comment repeats the test names and discussion. The local comment check is advisory, not a repository CI requirement. No files were edited.

src/utils/ens.ts
```diff
@@ -73,6 +78,25 @@ function isNoRecordRevert(domainType: DomainType, e: any): boolean {
...
+    address: universalResolverAddress,
+    abi: UNIVERSAL_RESOLVER_ABI,
+    functionName: 'findResolver',
+    args: [toHex(packetToBytes(name))]
+  });
+
+  return ensV1Delegates.some(
+    (delegate) =>
+      !!delegate && delegate.toLowerCase() === resolver.toLowerCase()
+  );
```
tony8713, line 101: These two pins are compared raw, unlike the helper, which is checked against ROOT_REGISTRY at 285-290. If ENS repoints the reserved slots or the root's TLD resolvers without changing the root, readsEnsV1 goes false and every unmigrated .eth and DNS name reads as unowned, silently. Every redeploy so far replaced the root too, so low likelihood; nonblocking.

src/utils/ens.ts
```diff
@@ -90,6 +114,60 @@ async function getEnsAddressStrict(
...
+    if (helperRoot !== resolverRoot) {
+      throw new Error(
+        `ENSv2 helper ${universalHelperAddress} reads root registry ${helperRoot}, universal resolver reads ${resolverRoot}`
+      );
+    }
+  })();
+
+  clientCache.set(key, { promise, expiresAt: Date.now() + ROOT_MATCH_TTL_MS });
+  // evict on rejection so the next call retries
+  promise.catch(() => clientCache.delete(key));
```
tony8713: The cache semantics are untested: deleting this eviction, setting the TTL to Infinity, or skipping the lookup at 139 all keep the 38 unit tests green, because nothing calls getEnsOwner twice on one client. Without the eviction one transient ROOT_REGISTRY failure latches a rejected promise for five minutes. A second-call test on the same client would cover it; nonblocking.

src/utils/ens.ts
```diff
@@ -185,26 +263,61 @@ export async function getEnsOwner(
...
+      address: universalHelperAddress,
+      abi: UNIVERSAL_HELPER_ABI,
+      functionName: 'findExactOwner',
       // viem's own encoding, so labels the strict DNS format rejects still resolve
       args: [toHex(packetToBytes(normalized))]
     });
   }
 
+  // a name ENSv2 has taken over, or whose .eth registration lapsed past grace,
+  // keeps a stale v1 entry: read v1 only where ENSv2 still delegates to it
```
tony8713: Still off on timing: PermissionedRegistry.getResolver returns zero at the v2 reservation expiry, about 28 days before v1 grace ends, while ETHRenewerV1 can still renew. So the v1 read stops before grace lapses, not after. "whose .eth reservation expired in v2" would be accurate.

tony8713 (approved):
Approving. The DNS-to-ENSv1 delegation case from my earlier review is fixed: findResolver returns the DNS TLD resolver at the TLD boundary, so ethplay.org and gregskril.com keep their v1 owners and tiny.fox.eth stays zero.

This is a review of 5de195e. f0db0bb landed while I was writing it and only trims comments, so I read it too: it already drops the em dashes and the "verified or rejected" wording and cuts the narrative comments I was going to ask you to remove. Line numbers below are against f0db0bb.

Two nonblocking risks and one remaining comment fix, inline.

src/utils/ens.ts
wa0x6e: Agreed on the mechanism: a mismatch at this compare reads as "not delegated", so the name comes back unowned with nothing thrown.

What I measured: neither pin is a proxy (EIP-1967 implementation slot is zero on both; the Universal Resolver proxy shows one, as a control), so any ENS fix to the mirror or the DNS TLD resolver is a new address. The three Sepolia deployments visible in `ensdomains/contracts-v2` each replaced the root together with `ENSV1Resolver`, and that case is already loud through the root check. Only an in-place repoint under an unchanged root is silent.

Not closing it in this PR. The guard I would want — a 2LD `.eth` name with no ENSv2 owner but a non-zero, unpinned resolver is a reserved slot pointing somewhere new, so throw — depends on "zero owner + resolver set" meaning reserved and nothing else. I read that from `PermissionedRegistry` on `main`, whose committed Sepolia root is not the live one, so it is not verified against the deployed build. Tracking as a follow-up: #1251.

src/utils/ens.ts
wa0x6e: Confirmed by mutation before changing anything: with the eviction deleted, the TTL set to `Infinity`, or the lookup forced to miss, the 38 existing cases all stayed green.

Added three cases on a shared client in 64e073e: retry after a failed root check, one root check per client inside the TTL, and a re-check once it lapses. Each of those three mutants now fails exactly one of them.

src/utils/ens.ts
wa0x6e: You are right, and it is measurable. `ETHRenewerV1.GRACE_PERIOD()` is 90 days, and for `bob`, `ens`, `demodao` and `boorger` the v2 reservation expires 62 days after the v1 expiry — 28 days before v1 grace ends. Simulating `findResolver(bob.eth)` one day past its reservation expiry returns the zero address while the name is still inside v1 grace.

Comment reworded as you suggested in 64e073e, along with the unit case title that said the same thing, and the PR body sentence.

That measurement also put a date on the `bob.eth` e2e case: its reservation lapses on 2026-10-17, after which it reads as unowned. It duplicated the `ens.eth` case above it, so I replaced it with `fox.eth`, which is registered in ENSv2 — until now no live case asserted a non-zero `findExactOwner`. It fails if the helper read is dropped, since `fox.eth` resolves through its own resolver and the v1 read is denied.

src/utils/ens.ts
wa0x6e: 802c7f3 then replaced the hand-rolled cache with viem's `withCache` (same semantics: shared in-flight check, rejections never stored, 5-minute TTL). The same three cases guard it: setting `cacheTime` to 0 or `Infinity`, or dropping the client from the key, each fails them.

src/utils/ens.spec.ts
```diff
@@ -162,54 +177,192 @@ describe('getEnsOwner findOwner fallback', () => {
...
+    await expect(getEnsOwner('x.eth', '11155111', opts)).rejects.toBeDefined();
+    client.readContract
+      .mockResolvedValueOnce(ROOT)
+      .mockResolvedValueOnce(ROOT)
+      .mockResolvedValueOnce(OWNER);
+    await expect(getEnsOwner('x.eth', '11155111', opts)).resolves.toBe(OWNER);
+    expect(rootReads(client)).toBe(4);
+  });
+
+  test('checks the root once per client within the TTL', async () => {
```
tony8713, line 321: The shared in-flight half of the cache has no test: a result-only cache with no promise sharing passes all 41. Promise.all of two getEnsOwner calls on one mockVerifiedClient with readContract.mockResolvedValue(OWNER), then expect(rootReads(client)).toBe(2), catches it and passes at head. Low risk since the sharing is viem's; nonblocking.

test/e2e/utils.spec.js
```diff
@@ -113,9 +113,9 @@ describe('utils', () => {
       );
     });
 
-    test('fall back to the name owner on testnet for bob.eth', async () => {
-      await expect(getSpaceController('bob.eth', '11155111')).resolves.toBe(
-        '0x179A862703a4adfb29896552DF9e307980D19285'
+    test('resolve a name registered in ENSv2 through its ENSv2 owner', async () => {
+      await expect(getSpaceController('fox.eth', '11155111')).resolves.toBe(
```
tony8713, line 117: fox.eth's stale v1 registry entry holds this same owner, so the case also passes with a plain v1 read. It does fail on master and when the helper read is dropped with the gate kept, so it still guards the change; a name registered only in v2 would pin the helper read itself, but I could not find another Sepolia name with a v2 owner. Nonblocking.

vite.config.js
```diff
@@ -2,6 +2,9 @@ import { defineConfig } from 'vite';
 
 export default defineConfig({
   test: {
+    // e2e specs resolve names over live RPC and CCIP gateways: measured 2-4s per
```
tony8713, line 5: The 2-4s per case did not reproduce: the slowest e2e case over two local runs was 597ms, and CI ran all 42 in 15s. The 20s timeout is fine; the number is the part to drop.

tony8713 (approved):
Approving 5a233b8. The three commits since f0db0bb answer all three of my earlier comments: the cache tests each kill exactly one mutant (cacheTime 0, cacheTime Infinity, a cache that keeps a rejection), the withCache swap keeps the same semantics and its key is stable because getViemClient memoizes the client, the pin compare is tracked in #1251, and the grace wording is fixed. The guard drop is right: delegatesToEnsV1 only runs with the helper set, and that path throws before any read when a pin is missing.

Tests CI on this head is red only in src/verify/starknet.spec.ts (starknet_getClassAt RPC), which this PR does not touch; ens.spec.ts and the e2e utils file pass in that run.

The four comments left in ens.ts (121, 254, 263, 286) each describe something a test already reddens on deletion, so they can go if you want the file bare. Two nonblocking nits inline, plus one wrong number.

